PR #18, opened 2017-11-03
Fix lime-company/powerauth-crypto#137

petrdvorak:
Evaluate the necessity to increment counter on server side in failed signatures.


romanstrobl:
I have tested the basic scenarios with both offline and online modes and found no problems. I'll do some more testing next week, for now the change looks good as long as the user does not exceed the lookahead limit. In this case there is still a workaround - remove and re-add the activation.

PR #37, opened 2018-03-07
Fix #32: Database deadlock when multiple PA signatures are validated at the same time

### LINKED ISSUE #32: Database deadlock when multiple PA signatures are validated at the same time
When our mobile client is trying to create (`/pa/token/create`) a multiple tokens at the same time (for the same activation), then some of the requests ends with an unknown error:
```
{"status":"ERROR","responseObject":{"code":"ERROR_GENERIC","message":"error.unknown"}}
```
After a small investigation in the logs, we found out a database deadlock on the PA server:
```
2017-12-07 14:09:48.802  WARN 1 --- [nio-8080-exec-7] o.h.engine.jdbc.spi.SqlExceptionHelper   : SQL Error: 1213, SQLState: 40001
2017-12-07 14:09:48.802 ERROR 1 --- [nio-8080-exec-7] o.h.engine.jdbc.spi.SqlExceptionHelper   : Deadlock found when trying to get lock; try restarting transaction
2017-12-07 14:09:48.802  INFO 1 --- [nio-8080-exec-7] o.h.e.j.b.internal.AbstractBatchImpl     : HHH000010: On release of batch it still contained JDBC statements
```

### UPDATE

The actual problem is not related to the tokens at all. Look for discussion below.

romanstrobl:
* Fix #32: Database deadlock when multiple PA signatures are validated at the same time - the problem was fixed by introducing a row level PESSIMISTIC_WRITE DB lock and increasing lock timeout  
* Fix ordering of signature audit records - concurrent requests occurring within same second were ordered incorrectly in MySQL database, ordering is now done by record ID which has no such problems
* Added test for MySQL deadlock. The test is disabled to avoid failing Travis builds due to missing MySQL database.

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
@@ -38,7 +38,7 @@ public interface SignatureAuditRepository extends CrudRepository<SignatureEntity
      * @param endingDate   Ending date (date "to").
      * @return List of {@link SignatureEntity} instances.
      */
-    List<SignatureEntity> findByActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDesc(String userId, Date startingDate, Date endingDate);
+    List<SignatureEntity> findByActivation_UserIdAndTimestampCreatedBetweenOrderByIdDesc(String userId, Date startingDate, Date endingDate);
 
     /**
      * Return signature audit records for given user, application and date range.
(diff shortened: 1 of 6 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
```diff
@@ -38,7 +38,7 @@
      * @param endingDate   Ending date (date "to").
      * @return List of {@link SignatureEntity} instances.
      */
-    List<SignatureEntity> findByActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDesc(String userId, Date startingDate, Date endingDate);
+    List<SignatureEntity> findByActivation_UserIdAndTimestampCreatedBetweenOrderByIdDesc(String userId, Date startingDate, Date endingDate);
```
petrdvorak: IMO this will not work in some Oracle configurations where [Oracle may decide to generate ID sequence out of order](https://stackoverflow.com/questions/13513936/oracle-sequence-value-are-not-ordered). With our current `hibernate_sequence`, this should not happen. However, [we would like to migrate to more lose entity ID generation](https://github.com/lime-company/powerauth-server/issues/7), and then this could cause problem.

I suggest ordering records by timestamp and then by ID (to distinguish order for records in the same second)....

romanstrobl: Ok, makes sense, I will update the method name.

PR #38, opened 2018-03-08
Add blocked reason and vault unlocked reason

### LINKED ISSUE #11: Add information about BLOCK reason
When user blocks the device, we should be able to assign a reason. Also, we should have a flag to indicate that the device was blocked by the bank and therefore cannot be unblocked by the user.

### LINKED ISSUE #19: Add vault unlock reason
Currently, we just enable vault unlock. However, we should be able to specify the vault unlock reason for the auditing purposes. There are several reasons why user may need to use vault unlock:

- Enabling Touch ID / Biometry
- Signing a document
- Retrieving an encryption key

romanstrobl:
* Fix #11: Add information about BLOCK reason
* Fix #19: Add vault unlock reason
* Additional information is written in signature audit records with both blocked reason and vault unlocked reason
* Fixed typo in method getActivationListForUser()
* Added DB columns for blockedReason and additionalInfo
* Updated XSD and WSDL


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
```diff
@@ -125,4 +129,33 @@ void logSignatureAuditRecord(ActivationRecordEntity activation, SignatureType si
         signatureAuditRepository.save(signatureAuditRecord);
     }
 
+    private String keyValueMapToString(KeyValueMap keyValueMap) {
```
petrdvorak: This probably should go in a separate `KeyValueMapConverter` class, with methods `fromString(String s)` and `toString()`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
```diff
@@ -140,10 +139,12 @@ public VerifyOfflineSignatureResponse verifyOfflineSignature(String activationId
                                                                  String dataString, CryptoProviderUtil keyConversionUtilities)
             throws UnsupportedEncodingException, InvalidKeySpecException, InvalidKeyException {
 
-        final VerifySignatureResponse verifySignatureResponse = verifySignature(activationId, signatureType, signature, dataString, null, keyConversionUtilities, true);
+        KeyValueMap additionalInfo = new KeyValueMap();
```
petrdvorak: We should probably allow `additionalInfo` to be `null`, so that we do not need to create dummy objects whenever calling the method. We could do null check inside `verifySignature`...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
romanstrobl: Sure. I was thinking whether using JSON may be a better format for storing the data in the DB, what do you think? We would not need a converter.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
romanstrobl: Ok, I will update it.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
petrdvorak: I think we should use JSON (seems better, since DB engines can use their JSON processing features), but I still think that separate converter is better than in place usage of generic `ObjectMapper`.

romanstrobl: Ok, I'll switch to JSON and move conversion methods it to another class.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/AdditionalInformation.java
```diff
@@ -0,0 +1,16 @@
+package io.getlime.security.powerauth.app.server.database.model;
+
+/**
+ * Constants for additional information related to activations and signature audit records.
+ *
+ * @author Roman Strobl, roman.strobl@lime-company.eu
+ */
+public class AdditionalInformation {
+
+    public static String BLOCKED_REASON = "BLOCKED_REASON";
```
petrdvorak: ... `final`

petrdvorak:
Looks OK

Issue #7, opened 2017-07-22
Add SequenceGenerator annotation on entity ID's


petrdvorak:
In order to have multiple sequences under Oracle DB, a SequenceGenerator annotation should be present with the GeneratedValue annotation.

PR #40, opened 2018-03-13
Fix #12: Add activation status change log

### LINKED ISSUE #12: Add activation status change log
Currently, we do not store activation status changes. We should have some log of activation status changes.

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Entity representing activation history used for storing activation status changes.
+ *
+ * @author Roman Strobl, roman.strobl@lime-company.eu
+ */
+@Entity(name = "pa_activation_history")
+public class ActivationHistoryEntity implements Serializable {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
```
petrdvorak: We should probably add a `@SequenceGenerator` annotation, so that we do not use the default `hibernate_sequence` on Oracle. We should also fix the SQL script for Oracle to use another sequence. Issue is related to #7.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ActivationHistoryRepository.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+
+/**
+ * Database repository for accessing activation history.
+ *
+ * @author Roman Strobl, roman.strobl@lime-company.eu
+ */
+public interface ActivationHistoryRepository extends CrudRepository<ActivationHistoryEntity, Long> {
+
+    List<ActivationHistoryEntity> findByActivation_ActivationIdAndTimestampCreatedBetweenOrderByTimestampCreatedDesc(String activationId, Date startingDate, Date endingDate);
```
petrdvorak: Please use HQL / `@Query` annotation, so that you do not write the code the way I did - the method name is way too long...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
petrdvorak: (we do not need to fix #7, we just should not make the problem more pressing...)

PR #47, opened 2018-03-27
Prepare release 0.18.0

petrdvorak:
This PR will be unnecessarily large, since we were not integrating in master for quite some time. We need to review the changes and make sure to never do this mistake again...


petrdvorak:
Note: While we approve the pull request, we should also make sure that our docs are up to date.

PR #67, opened 2018-04-11
Fix #48: Rename long repository methods and use annotation

### LINKED ISSUE #48: Rename long repository methods and use annotation
Currently, we use some extremely long names in repository interfaces, such as:

- `findByActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDescIdDesc `
- `findByActivation_ApplicationIdAndActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDescIdDesc `

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
```diff
@@ -38,7 +39,8 @@
      * @param endingDate   Ending date (date "to").
      * @return List of {@link SignatureEntity} instances.
      */
-    List<SignatureEntity> findByActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDescIdDesc(String userId, Date startingDate, Date endingDate);
+    @Query("SELECT s FROM SignatureEntity s WHERE s.activation.userId = ?1 AND s.timestampCreated BETWEEN ?2 AND ?3 ORDER BY s.timestampCreated DESC")
```
petrdvorak: The `ORDER BY` clause is missing `AND id DESC`, to keep the query equivalent and assure best attempt ordering for records with the same timestamp.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
```diff
@@ -49,6 +51,7 @@
      * @param endingDate    Ending date (date "to").
      * @return List of {@link SignatureEntity} instances.
      */
-    List<SignatureEntity> findByActivation_ApplicationIdAndActivation_UserIdAndTimestampCreatedBetweenOrderByTimestampCreatedDescIdDesc(Long applicationId, String userId, Date startingDate, Date endingDate);
+    @Query("SELECT s FROM SignatureEntity s WHERE s.activation.application.id = ?1 AND s.activation.userId = ?2 AND s.timestampCreated BETWEEN ?3 AND ?4 ORDER BY s.timestampCreated DESC")
```
petrdvorak: Same as above, missing `ORDER BY ... AND id DESC`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/SignatureAuditRepository.java
romanstrobl: Good catch!

PR #79, opened 2018-05-30
Fix #39: Application level record encryption

### LINKED ISSUE #39: Application level record encryption
At the moment, we store device related server private key as a plain text in the database. This is not a big issue in some scenarios, since back-ends are in a secure infrastructure with very limited access, however we could still improve security of the key by:

1. Introducing a new property (in case the property is not set, no encryption is used):

```
powerauth.server.db.master.encryption.key=MTIzNDU2Nzg5MDEyMzQ1Ng==
```

2. Using conversion mechanism whenever loading a `serverPrivateKey` value for the activation entity.

Note that since there is the good old rule "Same data should result in different encrypted values", we should introduce IV value for the encryption that should be stored with the value, such as (pseudo-code):

```java
public byte[] encrypt(byte[] orig, SecretKey key) {
    byte[] iv = Bytes.random(16);
    byte[] encrypted = aes.encrypt(orig, iv, key);
    byte[] record = iv.append(encrypted)
    return record;
}

public byte[] decrypt(byte[] record, SecretKey key) {
    byte[] iv = record.byteRange(0, 16); // offset, length
    byte[] encrypted = record.byteRange(16, -1); // offset, remaining
    byte[] orig = aes.decrypt(encrypted, iv, key);
    return orig;
}
```

Also, in order to achieve consistency between activation record and encrypted server private key (to avoid partial record swap), we should pay attention to how we construct the encryption key. It should be derived from the master DB encryption key using [KDF_INTERNAL](https://github.com/lime-company/powerauth-crypto/wiki/Basic-definitions) with user ID and activation ID derivate as a base, like so:

```java
public SecretKey deriveSecretKey(SecretKey masterDbEncryptionKey, String userId, String activationId) {
    String base = activationId.getBytes("UTF-8");
    String salt = userId.getBytes("UTF-8");
    byte[] index = PBKDF2.expand(base, salt, iterations, lengthInBits);
    return KDF_INTERNAL.derive(masterDbEncryptionKey, index);
}
```

Since we need entity context in order to achieve this, we very likely aren't able to use (aspect / annotation based) converter...

romanstrobl:
Added default values in Oracle DDL based on MySQL DDL


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -105,6 +110,11 @@ public void setActivationHistoryServiceBehavior(ActivationHistoryServiceBehavior
         this.activationHistoryServiceBehavior = activationHistoryServiceBehavior;
     }
 
+    @Autowired
```
petrdvorak: Why do we use setter injection here? Could we use constructor injection that we already use in this class?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/EncryptionServiceBehavior.java
```diff
@@ -67,6 +71,11 @@ public void setLocalizationProvider(LocalizationProvider localizationProvider) {
         this.localizationProvider = localizationProvider;
     }
 
+    @Autowired
```
petrdvorak: dtto - using constructor injection?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
```diff
@@ -106,6 +109,11 @@ public void setCallbackUrlBehavior(CallbackUrlBehavior callbackUrlBehavior) {
         this.callbackUrlBehavior = callbackUrlBehavior;
     }
 
+    @Autowired
```
petrdvorak: dtto - using constructor injection?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/VaultUnlockServiceBehavior.java
```diff
@@ -55,6 +58,12 @@ public VaultUnlockServiceBehavior(ActivationRepository powerAuthRepository) {
 
     // Prepare converters
     private ActivationStatusConverter activationStatusConverter = new ActivationStatusConverter();
+    private ServerPrivateKeyConverter serverPrivateKeyConverter;
+
+    @Autowired
```
petrdvorak: dtto - using constructor injection?

petrdvorak:
I also prepared a documentation on our wiki:

https://github.com/lime-company/powerauth-server/wiki/Encrypting-Records-in-Database

petrdvorak:
I also implemented following minor enhancements: https://github.com/lime-company/powerauth-server/pull/79/commits/25fba1983a02f941d0c736a830a4b08e3fd6a527

@romanstrobl Could you have a look I didn't break anything?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/converter/ServerPrivateKeyConverter.java
```diff
@@ -0,0 +1,199 @@
...
+     */
+    private SecretKey deriveSecretKey(SecretKey masterDbEncryptionKey, String userId, String activationId) {
+        try {
+            // See: https://github.com/lime-company/powerauth-server/issues/39
+            // activationId is used as base for PBKDF2 function
+            String base = activationId;
+            // userId is used as salt
+            byte[] salt = userId.getBytes("UTF-8");
+            // Derive index using PBKDF2 function
+            byte[] index = keyGenerator.deriveSecretKeyFromPassword(base, salt).getEncoded();
```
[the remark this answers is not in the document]
romanstrobl: Good point.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: There are more fields in this class that use setter injection than fields using constructor injection, so I followed this pattern.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/EncryptionServiceBehavior.java
romanstrobl: I chose setter injection in ActivationServiceBehavior, so I followed the same pattern in all classes.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
romanstrobl: See previous comment.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/VaultUnlockServiceBehavior.java
romanstrobl: See previous comment.

romanstrobl:
Yes, the changes look good.

romanstrobl:
All requested changes should be implemented now (except for @Autowired changes, feel free to update the injection code if you prefer some other way).

PR #153, opened 2018-08-21
Fix #152: Investigate Spring-WS security latest version

### LINKED ISSUE #152: Investigate Spring-WS security latest version
We are still using old versions of the Spring WS-Security implementation. We should re-evaluate if the latest version is still not useable.

See wultra/powerauth-crypto#139 for details.

petrdvorak:
It seems that the latest version of Spring WS-Security fixes the issue.

However, based on comments in [SWS-970](https://jira.spring.io/browse/SWS-970), we need to make sure that the code is not leaking memory. @romanstrobl, could you have a look and verify this is not the case? (I have more trust in your profiling skills...)

### DIFF
--- a/powerauth-java-client-spring/pom.xml
+++ b/powerauth-java-client-spring/pom.xml
@@ -17,65 +17,65 @@
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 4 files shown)


romanstrobl:
Sure, I will observe PA server in the profiler with this fix.

romanstrobl:
I looked at the issue and the problem is not a memory leak, it is just a large number of NIO threads whose number increases until no new native threads can be created. We will for sure be able to spot such situation using Visual VM.

romanstrobl:
The mentioned problem with thread leaks does not occur in branch `issues/152-spring-ws-security`. I do see lots of NIO theads, however all of them are either running or parked. There is no problem with waiting threads as reported in https://jira.spring.io/browse/SWS-970.

PR #167, opened 2018-09-11
Fix #166: Improve the logging code clarity and performance

### LINKED ISSUE #166: Improve the logging code clarity and performance

petrdvorak:
Use SLF4J logger obtained via logger factory, instead of JULI logger obtain in place
Use generic string instead of null when logging fallback exceptions
Fix placeholders for SLF4J, use placeholders instead of string concatenation where possible


petrdvorak:
@romanstrobl I fixed the issues.

PR #174, opened 2018-09-13
Versioning of activations and migration to new status blob

### LINKED ISSUE #171: Crypto 3.0: Migrate to updated activation status blob
Currently the activation status blob uses `counter` as parameter in PowerAuth server. We should migrate PowerAuth server to Crypto 3.0 where `currentVersionByte` and `upgradeVersionByte` bytes are used instead of counter.

### LINKED ISSUE #172: Crypto 3.0: Versioning of activations
We should start versioning activations and store activation version in database.

romanstrobl:
Fix #172: Versioning of activations
Fix #171: Crypto 3.0: Migrate to updated activation status blob


romanstrobl:
Since there is no additional feedback, I am merging the pull request.

PR #178, opened 2018-09-27
Fix #162: Crypto 3.0: Tokens

### LINKED ISSUE #162: Crypto 3.0: Tokens
Create V3 interface for token create and remove actions as defined in specification:

```xml
<xs:element name="CreateTokenRequest">
    <xs:annotation>
    	<xs:documentation>Request for obtaining a new token for simple authentication.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="encryptedData" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="mac" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>

<xs:element name="CreateTokenResponse">
    <xs:annotation>
    	<xs:documentation>Request for obtaining a new token for simple authentication.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="encryptedData" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="mac" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>
```

The interface has changed only slightly for V3. The V2 implementation already uses ECIES, so we can reuse this implementation.

romanstrobl:
This PR finalizes migration of v3 tokens to ECIES with new sharedInfo parameters and request data.


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/common/KeyDerivationService.java
```diff
@@ -0,0 +1,70 @@
...
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+
+/**
+ * Service used for derivation of keys, shared by behaviors.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Service
```
petrdvorak: Man, I don't like this service on so many levels...

- It accepts DB entity and returns raw key data.
- It autowires one property and creates all other without DI.
- It mixes concerns randomly (derive key, decrypt private key, ...), but at the same time, it only introduces only one new method that could easily be a private method somewhere or a utility method in crypto...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v2/TokenBehavior.java
```diff
@@ -172,8 +169,9 @@ private EciesPayload createToken(String activationId, String ephemeralPublicKeyB
             final ObjectMapper mapper = new ObjectMapper();
             final byte[] tokenBytes = mapper.writeValueAsBytes(tokenInfo);
 
-            final BasicEciesDecryptor decryptor = new BasicEciesDecryptor((ECPrivateKey) privateKey);
-            return decryptor.encrypt(tokenBytes, (ECPublicKey) ephemeralPublicKey, ephemeralPublicKeyBytes);
+            final EciesDecryptor decryptor = new EciesDecryptor((ECPrivateKey) privateKey);
+            // The encryptResponseDirect() method is only used in version 2.0, because there is no incoming data to decrypt.
+            return decryptor.encryptResponseDirect(tokenBytes, ephemeralPublicKeyBytes);
```
petrdvorak: OK - I was a bit suspicious about this code before, but we should rewrite this. It is confusing - we work with key and then we randomly use `byte[]` to pass the public key, that is internally converted to key as a side effect and used as `sharedInfo1`. We should not need the`encryptResponseDirect` method at all if we just follow the ECIES template with the private key, public key and correctly setup `sharedInfo1` attribute (there is no change in functionality, we only updated API)... The problem propagates all the way to the `EnvelopeKey` class in crypto...

romanstrobl: I agree the `encryptResponseDirect` method is unusual. Note that it is only used for V2 tokens where there is no incoming `encryptedData` and `mac`. The method is equivalent to the original method:
https://github.com/wultra/powerauth-crypto/blob/2b420826bcf03e300ec72748818bf01d3a17d9cd/powerauth-java-crypto/src/main/java/io/getlime/security/powerauth/crypto/lib/encryptor/ecies/BasicEciesDecryptor.java#L176

I will look into rewriting the method.

petrdvorak: Let's have a look tomorrow.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/common/KeyDerivationService.java
petrdvorak: Let's have a look tomorrow.

romanstrobl: We agreed to split the concerns and keep decryption of server private key in behavior classes.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v2/TokenBehavior.java
romanstrobl: We agreed to keep the method as is for now and deprecate it as soon as version 2.0 of crypto is no longer used. We will only use the method once for V2 tokens.

romanstrobl:
I separated concerns in the transport key derivation method and converted the key derivation service into a utility class.

As for the `encryptResponseDirect` method, I added following warning for this method in `EciesDecryptor`: 

```This method will be deprecated in a future release and is used only in legacy ECIES support.```

PR #181, opened 2018-10-01
Invalid state response for signature verification

meddox:
Invalid verification should return real activation status not only REMOVED, to recognize BLOCKED state

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/SignatureServiceBehavior.java
@@ -207,7 +207,7 @@ public class SignatureServiceBehavior {
                     }
 
                     // return the data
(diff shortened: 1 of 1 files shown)


petrdvorak:
Hello @meddox, thank you very much for your contribution! :)

I have two comments:

1) Could you please only pass the activation status value in the parameter, not the whole activation object? This way, you will avoid passing `null` to the method and you can also drop the extra ternary operator.

2) Before we are able to accept your pull request, we need you to sign our CLA - could you please review it and if it is OK, sign it (on-line)? https://goo.gl/forms/VSTWvakH29uBXR8x2

petrdvorak:
Maybe one quick note, with historic references:

We originally intended to return `REMOVED` status for signature verification every time, in order to hide the reason of why the signature failed. However, this was then drawn useless after we started returning the raw activation status in `getActivationStatus` method - you can simply call this method and the value is returned...

After evaluating the changes you propose, we believe it makes sense to do these modifications since it will simplify integration - people will simply get the status whenever the signature si verified...

PR #185, opened 2018-10-02
Fix #163: Crypto 3.0: Get ECIES decryptor parameters

### LINKED ISSUE #163: Crypto 3.0: Get ECIES decryptor parameters
Implement method for obtaining the ECIES decryptor parameters which is required during the activation process, so that the intermediate application server can decrypt the request.

- New method for getting parameters for ECIES decryptor (E2EE)
- Application server can use this to decrypt E2EE encrypted payloads
- This interface doesn't allow keys derivation required for protocol internal usages (like Vault Unlock, Create Token, etc..)
- If `activationId` is not present, then it creates ECIES decryptor for "application" scope.
- If `activationId` is present, then it creates ECIES decryptor for "activation" scope (e.g. associated with the activation)

```xml
<xs:element name="GetEciesDecryptorRequest">
    <xs:annotation>
    	<xs:documentation>Request for obtaining ECIES decryptor.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <!-- activationId is optional -->
            <xs:element name="activationId" type="xs:string" minOccurs="0" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>

<xs:element name="GetEciesDecryptorResponse">
    <xs:annotation>
    	<xs:documentation>Response containing setup for ECIES decryptor.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="encKey" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="macKey" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="sharedInfo2" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>
```

romanstrobl:
This PR implements the `GetEciesDecryptor` service on PowerAuth server that will be used while processing the `@PowerAuthEncryption` annotation.

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java
@@ -115,17 +115,22 @@ public class EciesEncryptionBehavior {
         }
 
         try {
-            // Lookup the application
+            // Lookup the application version and check that it is supported
             final ApplicationVersionEntity applicationVersion = repositoryCatalogue.getApplicationVersionRepository().findByApplicationKey(request.getApplicationKey());
-            if (applicationVersion == null) {
+            if (applicationVersion == null || !applicationVersion.getSupported()) {
                 throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_APPLICATION);
             }
-            final ApplicationEntity application = applicationVersion.getApplication();
 
             // Get master private key
+            final ApplicationEntity application = applicationVersion.getApplication();
             final MasterKeyPairEntity masterKeyPairEntity = repositoryCatalogue.getMasterKeyPairRepository().findFirstByApplicationIdOrderByTimestampCreatedDesc(application.getId());
             final String masterPrivateKeyBase64 = masterKeyPairEntity.getMasterKeyPrivateBase64();
             final PrivateKey privateKey = keyConvertor.convertBytesToPrivateKey(BaseEncoding.base64().decode(masterPrivateKeyBase64));
+
+            // Check that private key was succesfully converted
+            if (privateKey == null) {
+                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_KEY_FORMAT);
+            }
 
             // Get application secret
             final byte[] applicationSecret = BaseEncoding.base64().decode(applicationVersion.getApplicationSecret());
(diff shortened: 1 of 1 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java
```diff
@@ -0,0 +1,214 @@
...
+                throw localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_INCORRECT_STATE);
+            }
+
+            // Lookup the application
+            final ApplicationVersionEntity applicationVersion = repositoryCatalogue.getApplicationVersionRepository().findByApplicationKey(request.getApplicationKey());
+            if (applicationVersion == null) {
+                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_APPLICATION);
+            }
+            final ApplicationEntity application = applicationVersion.getApplication();
+
```
petrdvorak: Here, we should probably add a check that `application.getId()` is equal to `activation.getApplicationId()`, to see that the provided application key is actually associated with the same application as the activation...

Generally speaking, we are sending application key here mainly to check if the given version is still supported. In principle, we have everything we need just from the activation ID. We are able to look up the application and master key pair. Is the `application` variable used, at all?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            // Check if the activation is in correct state
+            if (!ActivationStatus.ACTIVE.equals(activation.getActivationStatus())) {
+                throw localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_INCORRECT_STATE);
+            }
+
+            // Lookup the application
+            final ApplicationVersionEntity applicationVersion = repositoryCatalogue.getApplicationVersionRepository().findByApplicationKey(request.getApplicationKey());
+            if (applicationVersion == null) {
```
petrdvorak: We need to check that the activation version is supported.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java
```diff
@@ -0,0 +1,214 @@
...
+     */
+    private GetEciesDecryptorResponse getEciesDecryptorParametersForApplication(GetEciesDecryptorRequest request) throws GenericServiceException {
+        if (request.getApplicationKey() == null || request.getEphemeralKey() == null) {
+            throw localizationProvider.buildExceptionForCode(ServiceError.DECRYPTION_FAILED);
+        }
+
+        try {
+            // Lookup the application
+            final ApplicationVersionEntity applicationVersion = repositoryCatalogue.getApplicationVersionRepository().findByApplicationKey(request.getApplicationKey());
+            if (applicationVersion == null) {
```
petrdvorak: We need to check that the activation version is supported.

```java
if (!applicationVersion.isSupported()) {
    // ... throw
}

File: powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java (hunk 1 above)
Comment (romanstrobl) [reply powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/EciesEncryptionBehavior.java]: The `application` variable is used when getting the master server key:

```
final MasterKeyPairEntity masterKeyPairEntity = repositoryCatalogue.getMasterKeyPairRepository().findFirstByApplicationIdOrderByTimestampCreatedDesc(application.getId());
```

I will add the application ID check.

romanstrobl:
I added the checks + extra check to verify that converted private key is not null.

PR #187, opened 2018-10-04
Use ephemeralPublicKey instead of ephemeralKey

romanstrobl:
As agreed with @hvge, `ephemeralPublicKey` will be used instead of `ephemeralKey` in SOAP interfaces.


romanstrobl:
Additional pull requests depend on this change, so I will merge it to crypto30 branch, this change is trivial.

PR #189, opened 2018-10-08
Fix #165: Verify signature

### LINKED ISSUE #165: Crypto 3.0: Verify signature
Implement minor change in verify signature endpoint:

- Added `signatureVersion` to distinguish between V2 & V3 signatures.
- I think that we don't need to create a "v3" version of the method. We have added just one parameter.

```xml
<xs:element name="VerifySignatureRequest">
    <xs:annotation>
    	<xs:documentation>Request for the signature verification process.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="data" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="signature" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="signatureType" type="tns:SignatureType" minOccurs="1" maxOccurs="1"/>
            <xs:element name="signatureVersion" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>
```

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/SignatureServiceBehavior.java
```diff
@@ -332,12 +334,38 @@ private ValidateSignatureResponse validateSignature(ActivationRecordEntity activ
...
         for (long iteratedCounter = ctr; iteratedCounter < ctr + powerAuthServiceConfiguration.getSignatureValidationLookahead(); iteratedCounter++) {
-            signatureValid = powerAuthServerSignature.verifySignatureForData(signatureRequest.getData(), signatureRequest.getSignature(), signatureKeys, iteratedCounter);
+            switch (version) {
+                case 2:
+                    // Use numeric counter for counter data
+                    ctrData = ByteBuffer.allocate(16).putLong(8, iteratedCounter).array();
+                    break;
+                case 3:
+                    // Increment counter in case iterated counter is bigger than current counter
+                    if (iteratedCounter > ctr) {
```
petrdvorak: This `if` seems a bit weird to me - I think we should separate the idea of "counter" (both numerical and hash-based) and "counter data" (data based on a counter value, used in the signature).

```java
ctrData = ctrHash; // use ctrHash data as ctrData
ctrHash = hashBasedCounter.next(ctrHash); // always increment the counter
```

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/SignatureServiceBehavior.java
```diff
@@ -390,6 +418,19 @@ private void handleValidSignature(ActivationRecordEntity activation, ValidateSig
         // Get ActivationRepository
         final ActivationRepository activationRepository = repositoryCatalogue.getActivationRepository();
 
+        if (activation.getVersion() == 3) {
+            HashBasedCounter hashBasedCounter = new HashBasedCounter();
+            // Number of ctr_data increments is the lowest counter + 1 - current counter
+            // (+1, since the client has incremented the counter)
+            long incrementCount = validationResponse.getLowestValidCounter() + 1 - activation.getCounter();
+            byte[] ctrData = BaseEncoding.base64().decode(activation.getCtrDataBase64());
```
petrdvorak: I think we should pass the `ctrData` here, just as we do with counter (we use it only for v3) - otherwise, we iterate the hash twice, which is not necessary.

PR #202, opened 2018-10-24
Fix #201: Crypto 3.0: Add version to signature audit

### LINKED ISSUE #201: Crypto 3.0: Add version to signature audit
We should add version to signature audit, so that the version can be displayed in PowerAuth admin.

romanstrobl:
This PR adds version into signature audit, so that we can display signature version in audit log. Additionally the element name for `version` is unified.


romanstrobl:
We agreed to keep this improvement and prepare documentation about partitioning of signature audit log instead which will take care of the database table size problem.

Issue #9, opened 2017-08-08
Improve input parameter validation


petrdvorak:
The service behavior sometimes assumes it receives correct attribute values from our integration libraries. We should implement more bulletproof parameter checking.

This issue does not have a large impact, since it assumes incorrect deployment or invalid server setup.

PR #216, opened 2018-11-14
Fix #186: Crypto 3.0: Custom activation

### LINKED ISSUE #186: Crypto 3.0: Custom activation
Implement support for custom ECIES based activation as defined in specification:

### Layer 1

- Encrypted by ECIES at app scope
- Layer can be decrypted at the REST application server.
    ```json
    {
        "requestObject": {
            "type": "custom",
            "identityAttributes": {
                "login" : "...",
                "password" : "...",
                "arbitrary-key" : "value-dictionary"
            },
            "activationData": {
                "ephemeralKey"  : "BASE64",
                "encryptedData" : "BASE64",
                "mac"           : "BASE64"
            }
        }
    }
    ```
- PowerAuth server will return encrypted response:
  ```json
  {
      "encryptedData" : "BASE64",
      "mac"           : "BASE64"
  }
  ```
  Note that the response must be once more time encrypted with key derived for Layer 1.

### `CreateActivation` (custom activation)

- `nonce` must be validated to be non-empty

```xml
<xs:element name="CreateActivationRequest">
    <xs:annotation>
        <xs:documentation>Request for creating a new activation directly.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="userId" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="timestampActivationExpire" type="xs:dateTime" minOccurs="0" maxOccurs="1"/>
            <xs:element name="maxFailureCount" type="xs:long" minOccurs="0" maxOccurs="1"/>
            <!-- the rest is same as for regular activation -->
            <xs:element name="encryptedData" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="mac" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>

<xs:element name="CreateActivationResponse">
    <xs:annotation>
        <xs:documentation>Response for creating a new activation directly.</xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="encryptedData" type="xs:string" minOccurs="1" maxOccurs="1"/>
            <xs:element name="mac" type="xs:string" minOccurs="1" maxOccurs="1"/>
        </xs:sequence>
    </xs:complexType>
</xs:element>
```

romanstrobl:
This PR introduces custom activation implementation on PowerAuth server.

I added the timestamp expiration check which was previously missing to v2 version, too.

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
@@ -126,6 +126,11 @@ public class PowerAuthServiceImpl implements PowerAuthService {
     @Override
     @Transactional
     public GetActivationListForUserResponse getActivationListForUser(GetActivationListForUserRequest request) throws GenericServiceException {
+        if (request.getUserId() == null) {
+            logger.warn("Invalid request");
+            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+        }
+        // The applicationId can be null, in this case all applications are used
         try {
             String userId = request.getUserId();
             Long applicationId = request.getApplicationId();
@@ -142,6 +147,10 @@ public class PowerAuthServiceImpl implements PowerAuthService {
     @Override
     @Transactional
     public GetActivationStatusResponse getActivationStatus(GetActivationStatusRequest request) throws GenericServiceException {
+        if (request.getActivationId() == null) {
+            logger.warn("Invalid request");
+            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+        }
         try {
             String activationId = request.getActivationId();
             logger.info("GetActivationStatusRequest received, activation ID: {}", activationId);
@@ -161,6 +170,11 @@ public class PowerAuthServiceImpl implements PowerAuthService {
     @Override
     @Transactional
     public InitActivationResponse initActivation(InitActivationRequest request) throws GenericServiceException {
+        if (request.getUserId() == null) {
+            logger.warn("Invalid request");
+            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+        }
+        // The maxFailedCount and activationExpireTimestamp values can be null, in this case default values are used
         try {
             String userId = request.getUserId();
             Long applicationId = request.getApplicationId();
@@ -182,6 +196,10 @@ public class PowerAuthServiceImpl implements PowerAuthService {
     @Override
     @Transactional
     public PrepareActivationResponse prepareActivation(PrepareActivationRequest request) throws GenericServiceException {
+        if (request.getActivationCode() == null || request.getApplicationKey() == null || request.getEphemeralPublicKey() == null || request.getMac() == null || request.getEncryptedData() == null) {
+            logger.warn("Invalid request");
+            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+        }
         try {
             String activationCode = request.getActivationCode();
             String applicationKey = request.getApplicationKey();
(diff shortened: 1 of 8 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -205,7 +205,34 @@ public PrepareActivationResponse prepareActivation(PrepareActivationRequest requ
     @Override
     @Transactional
     public CreateActivationResponse createActivation(CreateActivationRequest request) throws GenericServiceException {
-        throw new IllegalStateException("Not implemented yet.");
+        try {
+            // Get request parameters
+            String userId = request.getUserId();
+            Date activationExpireTimestamp = XMLGregorianCalendarConverter.convertTo(request.getTimestampActivationExpire());
```
petrdvorak: `request.getTimestampActivationExpire()` could be `null`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -205,7 +205,34 @@ public PrepareActivationResponse prepareActivation(PrepareActivationRequest requ
...
+            Long maxFailedCount = request.getMaxFailureCount();
```
petrdvorak: Could I double check where the default value (`5`) is set? In general, I believe we should do some basic request validation here, to make the future work on wultra/powerauth-server#9 a bit easier.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -205,7 +205,34 @@ public PrepareActivationResponse prepareActivation(PrepareActivationRequest requ
...
+            String applicationKey = request.getApplicationKey();
+            byte[] ephemeralPublicKey = BaseEncoding.base64().decode(request.getEphemeralPublicKey());
+            byte[] mac = BaseEncoding.base64().decode(request.getMac());
```
petrdvorak: What if the request values are `null`?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
romanstrobl: Yes, this is handled as previously in the `initActivation()` method. By default the value is `null` and it is non-null only when it is modified.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
romanstrobl: Yes, this is handled as previously in the `initActivation()` method. By default the value is `null` and it is non-null only when it is modified.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
romanstrobl: Good catch, I will fix it in all occurrences.

romanstrobl: I will add new error `INVALID_REQUEST` and add null checks to all methods in `PowerAuthServiceImpl`, such validation is generally missing.

romanstrobl: I added nullability checks into all service methods, added the error as well as fixed missing error messages for other errors.

PR #224, opened 2018-11-21
Fix #209: Prepare DDL for indexes for supported databases

### LINKED ISSUE #209: Prepare DDL for indexes for supported databases
We should prepare an official DDL for indexes for supported databases.

romanstrobl:
I added DB indexes into official DDL scripts. They were tested on Oracle, MySQL and PostgreSQL.


romanstrobl:
Please don't approve this PR yet, I just found out that some databases do not add foreign key indexes automatically. MySQL does that (we had duplicate indexes for foreign keys), however Oracle and PostreSQL do not add indexes automatically on foreign keys.

romanstrobl:
The DB indexes are ready for review now.

PR #232, opened 2018-12-04
Fix #227: InitActivation should use version 2 duplicity check when checking for duplicates

### LINKED ISSUE #227: InitActivation should use version 2 duplicity check when checking for duplicates
Currently the same `initActivation` method is shared between version 2 and version 3 of crypto.

This is problematic for duplicity checks:
- in version 2 the activation short ID needs to be unique
- in version 3 the whole activation code needs to be unique

Currently generated activation codes are guaranteed to be unique, however generated activation short IDs may not be unique, because the duplicity check needs to be more strict for version 2 than for version 3.

romanstrobl:
The uniqueness check is reverted to version 2 of crypto (the condition is unique `activationIdShort` per application instead of unique `activationCode` per application).

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ActivationRepository.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ActivationRepository.java
@@ -104,14 +104,18 @@ public interface ActivationRepository extends CrudRepository<ActivationRecordEnt
     ActivationRecordEntity findCreatedActivationWithoutLock(Long applicationId, String activationCode, Collection<ActivationStatus> states, Date currentTimestamp);
 
     /**
-     * Get count of activations identified by an activation code associated with given application.
+     * Get count of activations identified by an activation short ID associated with given application.
+     *
+     * The check for the first half of activation code is required for version 2.0 of PowerAuth crypto. In future the
+     * uniqueness check will be extended to whole activation code once version 2.0 of PowerAuth crypto is no longer
+     * supported.
      *
      * @param applicationId     Application ID
-     * @param activationCode    Activation code
+     * @param activationIdShort Activation ID short
      * @return Count of activations matching the search criteria
      */
-    @Query("SELECT COUNT(a) FROM ActivationRecordEntity a WHERE a.application.id = ?1 AND a.activationCode = ?2")
-    Long getActivationCountByActivationCode(Long applicationId, String activationCode);
+    @Query("SELECT COUNT(a) FROM ActivationRecordEntity a WHERE a.application.id = ?1 AND a.activationCode LIKE ?2%")
+    Long getActivationCountByActivationIdShort(Long applicationId, String activationIdShort);
 
     /**
      * Find the first activation associated with given application by the activation ID short.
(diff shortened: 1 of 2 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ActivationRepository.java
```diff
@@ -104,14 +104,18 @@
...
      *
      * @param applicationId     Application ID
-     * @param activationCode    Activation code
+     * @param activationIdShort Activation ID short
      * @return Count of activations matching the search criteria
      */
-    @Query("SELECT COUNT(a) FROM ActivationRecordEntity a WHERE a.application.id = ?1 AND a.activationCode = ?2")
-    Long getActivationCountByActivationCode(Long applicationId, String activationCode);
+    @Query("SELECT COUNT(a) FROM ActivationRecordEntity a WHERE a.application.id = ?1 AND a.activationCode LIKE ?2%")
+    Long getActivationCountByActivationIdShort(Long applicationId, String activationIdShort);
```
petrdvorak: I am not sure about this change. Crypto 3.0 does not have a "short activation ID", it only has an activation code. Therefore, I believe that the Java code should be written to cover Crypto 3.0. To check strong uniqueness, I would prefer to modify the HQL by using [`substring`](http://www.hplsql.org/substring) in the query...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -509,8 +509,9 @@ public InitActivationResponse initActivation(Long applicationId, String userId,
             String activationCode = null;
             for (int i = 0; i < powerAuthServiceConfiguration.getActivationGenerateActivationCodeIterations(); i++) {
                 String tmpActivationCode = powerAuthServerActivation.generateActivationCode();
-                Long activationCount = activationRepository.getActivationCountByActivationCode(applicationId, tmpActivationCode);
-                // Check that the temporary activation code is unique, otherwise generate a different code
+                String tmpActivationShortId = tmpActivationCode.substring(0, 11);
```
petrdvorak: dtto

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ActivationRepository.java
romanstrobl: We still support crypto v2 which contains short activation ID as a parameter.

See:
https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v2/ActivationServiceBehavior.java#L194

https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/resources/xsd/PowerAuth-2.0.xsd#L34

We will remove this method when crypto v2 is deprecated.

As for using `substring`, I would not recommend it, because it would degrade DB performance. When `LIKE ?2%` is used, index on activation code is leveraged in each database PowerAuth supports. In case of `substring` we would need to define a special index on the function, which can be tricky in some databases.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
romanstrobl: dtto

PR #260, opened 2019-01-31
Prepare release 0.21.0

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/SignatureServiceBehavior.java
```diff
@@ -361,8 +449,8 @@ private boolean handleInvalidApplicationVersion(ActivationRecordEntity activatio
             Long remainingAttempts = (activation.getMaxFailedAttempts() - activation.getFailedAttempts());
             if (remainingAttempts <= 0) {
                 activation.setActivationStatus(ActivationStatus.BLOCKED);
-                activationHistoryServiceBehavior.logActivationStatusChange(activation);
                 activation.setBlockedReason(AdditionalInformation.BLOCKED_REASON_MAX_FAILED_ATTEMPTS);
+                activationHistoryServiceBehavior.saveActivationAndLogChange(activation);
```
petrdvorak: Do we need to save the activation here? We are saving it a couple of lines below.

romanstrobl:
I fixed the issue with double save. I will proceed with the release.

PR #289, opened 2019-03-05
Fix #276: Startup issue on Wildfly with Java 11 and Spring Boot 2.1.2

### LINKED ISSUE #276: Startup issue on Wildfly with Java 11 and Spring Boot 2.1.2
The following error occurs in vanilla Wildfly installation with Java 11 and Spring boot `2.1.2`:

```
Caused by: java.lang.IllegalArgumentException: TransformerFactory does not recognise attribute 'http://javax.xml.XMLConstants/property/accessExternalDTD'."}}
```

The issue should be fixed in unreleased version of `spring-ws` project.

See also:
https://stackoverflow.com/questions/54186834/spring-boot-2-1-1-to-2-1-2-error-creating-bean-with-name-payloadrootannotation

We should check this issue after next upgrade of Spring dependencies.

The workaround is to comment out `org.apache.xalan`module in Wildfly configuration.

romanstrobl:
Changes:
- Fix #276: Startup issue on Wildfly with Java 11 and Spring Boot 2.1.2
- Reintroduce exclusions for ehcache and geronimo-javamail, they are still visible in OWASP report

The startup issues on Wildfly was resolved in spring-ws-security `3.0.7.RELEASE`.

### DIFF
--- a/powerauth-java-server/pom.xml
+++ b/powerauth-java-server/pom.xml
@@ -51,6 +51,8 @@
         <dependency>
             <groupId>org.springframework.ws</groupId>
             <artifactId>spring-ws-core</artifactId>
+            <!-- TODO - update/remove version shortly before final release, 3.0.6.RELEASE is the current version which causes startup issues on Wildfly -->
+            <version>3.0.7.RELEASE</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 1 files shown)


powerauth-java-server/pom.xml
```diff
@@ -51,6 +51,8 @@
         <dependency>
             <groupId>org.springframework.ws</groupId>
             <artifactId>spring-ws-core</artifactId>
+            <!-- TODO - update/remove version shortly before final release, 3.0.6.RELEASE is the current version which causes startup issues on Wildfly -->
```
petrdvorak: Is this a good idea, won't we really forget this? I would prefer downgrading the version and updating sooner before the release in case the issue is fixed.

romanstrobl: Let's keep the issue open then. I can test PA server on Wildfly in this issue branch, otherwise it would not start.

romanstrobl: The issue is fixed (I verified it), we just need to wait a bit for dependencies to line up with this `spring-ws-security` version.

romanstrobl:
I am still waiting for Spring boot release `2.1.4`, in this version we will not need to specify version of `spring-ws-security`, which seems to be a better fix.

PR #295, opened 2019-04-09
Fix #291: Implement support for recovery codes

### LINKED ISSUE #291: Implement support for recovery codes
Implement backend support for activation via recovery codes and management of recovery codes.

_Basically, a "recovery code" is associated with some activation and used as one-time credentials for creating a new activation._

_Use-case: I activate mobile token or mobile banking, a recovery code is displayed to me at the end of the process, and if I lose the phone or uninstall the app, I can activate a new instance using the recovery code._

romanstrobl:
This pull request contains implementation of recovery code support in PowerAuth server.


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/RecoveryServiceBehavior.java
```diff
@@ -0,0 +1,630 @@
...
+
+/**
+ * Behavior class implementing processes related to recovery codes.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Component
+public class RecoveryServiceBehavior {
+
+    private static final int PUK_COUNT = 10;
```
petrdvorak: Let's externalize this.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -928,4 +928,153 @@ public CommitUpgradeResponse commitUpgrade(CommitUpgradeRequest request) throws
         }
     }
 
+    @Override
+    @Transactional
+    public CreateRecoveryCodeResponse createRecoveryCode(CreateRecoveryCodeRequest request) throws Exception {
+        if (request.getApplicationId() <= 0L || request.getUserId() == null || request.getPukCount() < 1 || request.getPukCount() > 100) {
```
petrdvorak: Let's store the value 100 in a constant here in the class so that we can change the number easier.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/RecoveryServiceBehavior.java
romanstrobl: Good catch, I will make a DB column for this and update PowerAuth admin.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
romanstrobl: Yes, as it should be.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/RecoveryServiceBehavior.java
romanstrobl: Actually the PUK count is parameter for `CreateRecoveryCodeRequest`, so I think we can keep it this way. I just forgot to switch to the PUK count from request.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
romanstrobl: Fixed.

PR #309, opened 2019-05-12
Add application lookup by name

saalistaja:
Implemented by extending the current application query with new applicationName field. There has to be provided one of the `applicationId` or `applicationName` fields.

This change should be fully compatible with current requests.

I have few more questions, ideas:

* Should be the app name trimmed?
  * normalize value without trailing spaces when creating application or when querying by name
  * update current data in DB

* Should be the application name in DDL defined as not null?
  * I've fixed it in ddl for `pa_application` table, so it meets with the validation in `PowerAuthServiceImpl.createApplication`.
  * not sure if desired, but if yes I would like to prepare alter script for already existing schemas too?


docs/SOAP-Service-Methods.md
```diff
@@ -167,15 +167,16 @@ Get list of all applications that are present in this PowerAuth Server instance.
 
 ### Method 'getApplicationDetail'
 
-Get detail of application with given ID, including the list of versions.
+Get detail of application with given ID or name, including the list of versions.
```
romanstrobl: I am considering whether having a separate SOAP method would not be better than updating the `getApplicationDetail` method. I will think about it. @petrdvorak, what do you think?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/GetApplicationDetailTest.java
```diff
@@ -0,0 +1,100 @@
+package io.getlime.security.powerauth.app.server;
```
romanstrobl: Please add copyright

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/GetApplicationDetailTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.assertj.core.util.Lists;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@SpringBootTest
```
romanstrobl: Please add JavaDoc with @author.

saalistaja: resolved

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/GetApplicationDetailTest.java
saalistaja: resolved

docs/SOAP-Service-Methods.md
saalistaja: Well, for the current `getApplicationDetail` and its usage I see the id or name approach as the right one. But for sure as simple and fully clear approach two `getApplicationDetail` and `getApplicationDetailByName` will meet the requirements too. This case is a little bit not clear.

docs/PowerAuth-Server-0.22.0.md
```diff
@@ -51,6 +51,8 @@ Make sure you upgrade the web container to a version which supports Java 11 befo
 Following DB changes occurred between version 0.21.0 and 0.22.0:
 - Table `pa_activation_history` - added column `external_user_id`
 - Added tables, sequences and indexes for storage of Recovery Codes, Recovery PUKs and Recovery Configuration
+- Table `pa_application` - added not null constraint to column `name` (optional)
```
saalistaja: @romanstrobl Should we push the clients to apply those new constraints or there can be a case where they have already stored data which will block this (null name values, or not unique values). Then we should obviously prepare a fix and set default `''` for name?

saalistaja:
@romanstrobl 
The normalization (like trimming values) should imho happen both on UI and our service layer. It's a similar case like validation.

We should maybe also ensure trimming of all currently stored application names.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ApplicationRepository.java
```diff
@@ -20,11 +20,21 @@
...
  */
 public interface ApplicationRepository extends CrudRepository<ApplicationEntity, Long> {
 
+    /**
+     * Find application by name.
+     *
+     * @param applicationName Application name
+     * @return Optional application entity with given name, returns the first name match
+     */
+    Optional<ApplicationEntity> findFirstByName(String applicationName);
```
saalistaja: As currently we do allow not unique application names I had to put here `findFirst` otherwise it could end up in non unique result. If we will decide to push our clients to ensure unique already stored application names we can use simple `findByName`.

docs/PowerAuth-Server-0.22.0.md
romanstrobl: It is possible that clients have a null name in database (it occured in one case already). I do not think a default '' is a good name. I think it would be better to do a migration, something like: ```UPDATE pa_application SET application_name = "application_" + application_id WHERE application_name IS NULL```

romanstrobl:
@saalistaja I find the request normalization and especially trimming of existing DB values outside of scope of this pull request. It would be inconsistent with the rest of SOAP methods which in general do no trimming. I suggest we open another issue because there are many other SOAP methods where this would have to be done to get consistent behaviour of all SOAP methods.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/ApplicationRepository.java
romanstrobl: Yes, see my suggestion about updating existing DB values and adding the NOT NULL constraint.

docs/PowerAuth-Server-0.22.0.md
saalistaja: Brilliant, for sure that will be a better fix!

But still if there are already two same names now e.g. `MOB_APP` we will not succeed with the unique constraint. It will require manual DB update during the migration, is this acceptable? I could add a note about that.

romanstrobl: Yes, the UPDATE query would be used during migration right before activating the `NOT NULL` contstraint, that was my point.

romanstrobl: Ah I see, you are concerned about the UNIQUE index. We can prepare some check to avoid this case during migration, it should not happen.

saalistaja: Creation of the unique index will be the check itself but we should ensure the migration script was done (i suppose it is a manual script not ensured by the application start). Otherwise I would rather keep the `findFirstByName` query.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -638,7 +638,7 @@ public LookupApplicationByAppKeyResponse lookupApplicationByAppKey(LookupApplica
     }
 
     @Override
-    @Transactional
+    @Transactional(rollbackFor = GenericServiceException.class)
```
romanstrobl: @petrdvorak can you please check this? It seems that transaction rollbacks are not handeled correctly in PowerAuth server, if I understand this change.

petrdvorak: Good point...

PR #312, opened 2019-05-16
Fix #299: Allow use of biometry for off-line approvals

### LINKED ISSUE #299: Allow use of biometry for off-line approvals
Currently, the SOAP method for off-line code verification requires a signature type as an input. However, this does not work, because the calling system is not able to determine the signature type beforehand.

Instead, the SOAP method should accept a list of "allowed signature types" and validate signature against these types.

romanstrobl:
I updated the `verifyOfflineSignature` SOAP method to use the `allowBiometry` parameter instead of specifying signature type(s). The signature verification code was updated to handle multiple signature types and to store used signature type.


romanstrobl:
I have refactored the signature verification. The pull request is ready for review.

romanstrobl:
I simplified the code based on feedback from the pull request.

PR #340, opened 2019-08-08
Fix incomplete BC install instruction

petrvlcek:
Fixing incomplete Bouncy Castle installation tutorial for JDK8


petrdvorak:
I will make this fix a bit more detailed - apparently, there is a difference in setup between standalone and embedded Tomcat.

PR #353, opened 2019-09-20
Fix #348: Use IV in activation status blob encryption

### LINKED ISSUE #348: Crypto 3.1: Use IV in get activation status
In the next version of PowerAuth protocol, we're going to use initialization vector for status blob encryption. This change improves protection of status blob against possible replay attacks.

See wultra/powerauth-crypto#281 for more details.

hvge:


powerauth-java-client-axis/src/main/java/io/getlime/security/powerauth/soap/axis/client/PowerAuthServiceClient.java
```diff
@@ -336,14 +336,33 @@ private GregorianCalendar calendarWithDate(Date date) {
     }
 
     /**
-     * Call the getActivationStatus method of the PowerAuth 3.0 Server SOAP interface.
+     * Call the getActivationStatus method of the PowerAuth 3.0 Server SOAP interface. This method should be used only
+     * to acquire the activation status for other, than RESTful API purposes. If you're implementing the PowerAuth RESTful API,
```
romanstrobl: This sentence is a bit confusing, because PA server also has a RESTful API next to SOAP API. I believe you are referencing the powerauth-restful-integration project, right? We'd better make this distinction.

hvge: Fixed. Now I'm using naming from [our own documentation](https://developers.wultra.com/docs/2019.05/powerauth-crypto/Architecture-Overview) (e.g. PowerAuth standard RESTful API)

romanstrobl:
Perfect.

PR #355, opened 2019-09-23
Fix #354: Look ahead constant is part of status blob

### LINKED ISSUE #354: Crypto 3.1: Send look ahead constant in status blob
In the next version of PowerAuth protocol, we should send a value representing look ahead window back to the mobile client. The full explanation why we are going to change this is available in wultra/powerauth-mobile-sdk#236 ticket.

hvge:
This change also adding a basic validation to PowerAuthServiceConfiguration


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -148,6 +162,7 @@
      * Token timestamp validity in milliseconds, checked before validating the token.
      */
     @Value("${powerauth.service.token.timestamp.validity}")
+    @Min(0)
     private long tokenTimestampValidityInMilliseconds;
```
romanstrobl: The value should probably be non-zero, I suggest to use 1 as minimal value.

PR #356, opened 2019-09-28
Features/crypto 31

hvge:
This change adds support for PowerAuth protocol 3.1 to develop branch.


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -261,13 +262,14 @@ private VerifySignatureResponse verifySignatureImplNonTransaction(VerifySignatur
         String applicationKey = request.getApplicationKey();
         String dataString = request.getData();
         String signature = request.getSignature();
+        String signatureVersion = request.getSignatureVersion();
```
romanstrobl: In the beginning of the `verifySignature` method in this class there is a request validation logic. Since the `signatureVersion` is a required parameter, let's add it into the validation logic.

See:
https://github.com/wultra/powerauth-server/blob/5b7914229ce22990abebc4a7cb008d1a3bb346c0/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java#L278

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java
```diff
@@ -480,6 +482,7 @@ public VaultUnlockResponse vaultUnlock(VaultUnlockRequest request) throws Generi
             final String applicationKey = request.getApplicationKey();
             final String signature = request.getSignature();
             final SignatureType signatureType = request.getSignatureType();
+            final String signatureVersion = request.getSignatureVersion();
```
romanstrobl: In the beginning of the `vaultUnlock` method in this class there is a request validation logic. Since the `signatureVersion` is a required parameter, let's add it into the validation logic.

See:
https://github.com/wultra/powerauth-server/blob/5b7914229ce22990abebc4a7cb008d1a3bb346c0/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/v3/PowerAuthServiceImpl.java#L473

hvge:
> It would be nice to log the `signatureVersion` parameter into `pa_signature_audit` table, please consider this.

It's bigger change that I've expected. What about create a separate ticket for it? (Like add version to audit log)

romanstrobl:
> > It would be nice to log the `signatureVersion` parameter into `pa_signature_audit` table, please consider this.
> 
> It's bigger change that I've expected. What about create a separate ticket for it? (Like add version to audit log)

Yes, let's do it.

PR #377, opened 2019-12-03
Fix #375: Plain CTR_DATA is no longer in status.

### LINKED ISSUE #375: Crypto 3.1: Do not send CTR_DATA in status
The status blob should contain only a hash calculated from CTR_DATA, instead of the direct value of the counter.

Full details about the enhancement are available in wultra/powerauth-crypto#308

hvge:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -51,6 +51,7 @@
 import io.getlime.security.powerauth.crypto.lib.model.ActivationStatusBlobInfo;
 import io.getlime.security.powerauth.crypto.lib.model.RecoveryInfo;
 import io.getlime.security.powerauth.crypto.lib.model.exception.GenericCryptoException;
+import io.getlime.security.powerauth.crypto.lib.util.HashBasedCounterUtils;
```
romanstrobl: This seems unnecessary.

romanstrobl:
I am merging to pull request, otherwise we cannot deploy the build.

PR #384, opened 2020-01-10
Fix #352: Add application platform during activation

### LINKED ISSUE #352: Add application platform during activation
Currently, we do not store platform with the activation and it could be useful to have it...

romanstrobl:
Added platform and user device information on server.


docs/SOAP-Service-Methods.md
```diff
@@ -404,6 +406,8 @@ ECIES request should contain following data (as JSON):
  - `activationName` - Visual representation of the device, for example "Johnny's iPhone" or "Samsung Galaxy S".
  - `devicePublicKey` - Represents a public key `KEY_DEVICE_PUBLIC`  (base64-encoded).
  - `extras` - Any client side attributes associated with this activation, like a more detailed information about the client, etc.
+ - `platform` - User device platform: `ios`, `android`, `hw` or `unknown`.
+ - `deviceInfo` - Information about user device in format `{platform};{device-model-id}`, e.g. `ios;iPhone12,3`.
```
hvge: Please remove "ios" from the example. We have the platform already in the separate attribute.

powerauth-java-client-axis/src/main/resources/soap/wsdl/serviceV3.wsdl
```diff
@@ -1510,6 +1516,19 @@
...
+                <xs:annotation>
+                    <xs:documentation>Enum representing user device platforms.
+                    </xs:documentation>
+                </xs:annotation>
+                <xs:restriction base="xs:string">
+                    <xs:enumeration value="IOS"/>
+                    <xs:enumeration value="ANDROID"/>
+                    <xs:enumeration value="HW"/>
+                    <xs:enumeration value="UNKNOWN"/>
+                </xs:restriction>
```
hvge: Do not limit platform to the predefined set of platforms.

docs/SOAP-Service-Methods.md
romanstrobl: Yes, I will remove it.

powerauth-java-client-axis/src/main/resources/soap/wsdl/serviceV3.wsdl
romanstrobl: Ok, I will make a String out of the platform. I will just apply consistent case to avoid inconsistent values, e.g. `iOS` and `ios`

PR #406, opened 2020-04-14
Fix #396: OTP_USED state renamed to PENDING_COMMIT

### LINKED ISSUE #396: Rename OTP_USED state to PENDING_COMMIT
The `OTP_USED` state is conflicting with the activation OTP terminology. Let's investigate if the state name can be easily changed and if so, it is better to use a different name for this state.

hvge:


docs/PowerAuth-Server-0.24.0.md
```diff
@@ -40,6 +40,7 @@ PowerAuth server in version `0.24.0` slightly changed SOAP interface for protoco
 
 ### Support for Additional Activation OTP
 
+- Renamed `ActivationStatus.OTP_USED` enumeration to `ActivationStatus.PENDING_COMMIT`.
```
romanstrobl, line 43: We should highlight this change as an API change. I will do it.

romanstrobl:
@hvge please check the migration guide change.

hvge:
Looks OK

PR #410, opened 2020-04-15
Fix #400: "blocked_reason" renamed to "event_reason"

### LINKED ISSUE #400: Rename `block_reason` to `event_reason`
Currently, the `pa_activation_history` table has a column named `blocked_reason` that may contain a reason, why the activation was blocked. Due to changes related to an additional activation OTP (ticket #386), the activation history record can contain also reasons unrelated to the block. So, we should rename that column to `event_reason`.

hvge:


docs/SOAP-Service-Methods.md
```diff
@@ -1063,7 +1063,7 @@ Get the status change log for given activation and date range.
 | `Long` | `id` | Change ID |
 | `String` | `activationId` | An identifier of an activation |
 | `ActivationStatus` | `activationStatus` | An activation status at the moment of a signature verification |
-| `String` | `blockedReason` | Reason why activation was blocked (default: NOT_SPECIFIED) |
+| `String` | `eventReason` | Reason why this activation history record was created (default: null or NOT_SPECIFIED) |
```
romanstrobl, line 1066: It is a bit weird to provide 2 default values.

hvge: That's how it is in the database.

hvge: OK, I'll change that to `null`. It has now more wider meaning than before and for the most of events it's `null`

PR #441, opened 2020-06-09
Fix #372: Add activation flags

### LINKED ISSUE #372: Add activation flags
It would be nice to allow "flagging" each activation - basically, a CRUD of flags associated with any activation. As a result, activation could be `READONLY`, `PAYMENTS`, `FULLY_VERIFIED`, etc. depending on the requirements of the specific customer use-case.

romanstrobl:
This pull request implements CRUD for activation flags and exposes flags in existing methods that return activation information.

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
@@ -24,9 +24,7 @@ import com.google.common.io.BaseEncoding;
 import io.getlime.security.powerauth.app.server.configuration.PowerAuthServiceConfiguration;
 import io.getlime.security.powerauth.app.server.converter.v3.ActivationOtpValidationConverter;
 import io.getlime.security.powerauth.app.server.converter.v3.ActivationStatusConverter;
-import io.getlime.security.powerauth.app.server.converter.v3.RecoveryPukConverter;
-import io.getlime.security.powerauth.app.server.converter.v3.ServerPrivateKeyConverter;
-import io.getlime.security.powerauth.app.server.converter.v3.XMLGregorianCalendarConverter;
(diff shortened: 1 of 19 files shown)


romanstrobl:
I added basic tests for activation flags:
https://github.com/wultra/powerauth-tests/blob/develop/powerauth-backend-tests/src/test/java/com/wultra/security/powerauth/test/v31/PowerAuthActivationFlagsTest.java

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -300,27 +300,37 @@ public LookupActivationsResponse lookupActivations(List<String> userIds, List<Lo
...
-                activationServiceItem.setApplicationId(activation.getApplication().getId());
-                activationServiceItem.setApplicationName(activation.getApplication().getName());
-                // Unknown version is converted to 0 in SOAP
-                activationServiceItem.setVersion(activation.getVersion() == null ? 0L : activation.getVersion());
-                response.getActivations().add(activationServiceItem);
-            }
+        List<ActivationRecordEntity> filteredActivationList = new ArrayList<>();
+        // Filter activation by activation flags in case they are specified
+        if (activationFlags != null && !activationFlags.isEmpty()) {
+            List<ActivationRecordEntity> activationsWithFlags = activationsList.stream().filter(activation ->
```
hvge, line 306: Are we sure that `activationList` returned from `.lookupActivations(..)` is always non-null?

romanstrobl: Yes, the repository returns an empty list in case no match is found for the query criteria.

romanstrobl: It is a good point, we can skip all this processing in case the list is empty.

PR #449, opened 2020-06-26
Fix #445: Implement REST client based on WebClient

### LINKED ISSUE #445: Implement REST client based on WebClient
As part of migration to REST from SOAP, we should implement a REST client based on Spring WebClient. This client will be usable standalone as an alternative to the SOAP clients. The new REST client will be also used in the `powerauth-restful-integration` project for the Spring integration.

romanstrobl:
This draft pull request implements the REST client for PowerAuth server.

Notable changes:
- Client model classes are generated in new project `powerauth-client-model`. The XSD is used directly, so this avoids one usage of WSDL files.
- Model class packages were changed to `com.wultra`.
- Spring SOAP and REST clients share the `PowerAuthClient` interface as well as model classes. The SOAP and REST interfaces can be used interchangebly by autowiring a different client.

I will look into simplifying writing the client code, we have 3 clients now to update.


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/PowerAuthRequestWrapper.java
```diff
@@ -25,23 +25,23 @@
  * @param <T> Type of the request object instance.
  * @author Petr Dvorak, petr@wultra.com
  */
-public class RESTRequestWrapper<T> {
+public class PowerAuthRequestWrapper<T> {
```
petrdvorak, line 28: This is probably a leftover from the past - can we use the [`ObjectRequest`](https://github.com/wultra/lime-java-core/blob/develop/rest-model-base/src/main/java/io/getlime/core/rest/model/base/request/ObjectRequest.java) class? Or is this because of the JavaX Validation annotation?

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/PowerAuthResponseWrapper.java
```diff
@@ -25,7 +25,7 @@
  * @param <T> Type of the response object instance.
  * @author Petr Dvorak, petr@wultra.com
  */
-public class RESTResponseWrapper<T> {
+public class PowerAuthResponseWrapper<T> {
```
petrdvorak, line 28: Dtto as with request...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/controller/RESTControllerAdvice.java
```diff
@@ -44,14 +44,13 @@
      */
     @ResponseStatus(HttpStatus.BAD_REQUEST)
     @ExceptionHandler(value = GenericServiceException.class)
-    public @ResponseBody RESTResponseWrapper<List<RESTErrorModel>> returnGenericError(GenericServiceException ex) {
-        RESTErrorModel error = new RESTErrorModel();
+    public @ResponseBody
```
petrdvorak, line 47: Auto-formatting issue? (extra newline)

romanstrobl: Yes, I will fix it.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/PowerAuthRequestWrapper.java
romanstrobl: Yes, it seems like we can easily replace the wrappers with `ObjectRequest`/`ObjectResponse`. I though that there was some extra functionality, but there is not.

PR #452, opened 2020-07-31
Fix #451: Extended callbacks

### LINKED ISSUE #451: Extended callbacks
Currently, we only send an activation ID when calling the callback URLs. To improve the performance of some applications, it would be nice to allow extended callback configuration, to also send other attributes of activation.

The `CallbackUrlBehavior::notifyCallbackListeners` method that is responsible for handling the callbacks accepts application ID and activation ID. However, the activation ID is always set via `activation.getActivationId()` call. This means it should be simple to pass the entire activation instead and send multiple activation attributes to the call, depending on the configuration.

The `activationId` attribute would remain mandatory.

We would need to add a column with an encoded list of attributes to send in the callback (maybe as a JSON).

### UI Mockup

Screenshot 2020-07-22 at 20 49 53

romanstrobl:


romanstrobl:
The diff without migration to REST code is visible here:
https://github.com/wultra/powerauth-server/compare/issues/445-rest-client...issues/451-extended-callbacks

PR #460, opened 2020-09-09
Fix #459: Unclear logging message when activation expires

### LINKED ISSUE #459: Unclear logging message when activation expires
Message:

```
logger.warn("Activation not found with activation ID: {}", activationCode);
```

Should be:

```
logger.warn("Activation with the activation code: {} could not be obtained - it does not exist or it already expired.", activationCode);
```

petrdvorak:


romanstrobl (approved):
Looks good, thanks for fixing this one.

PR #480, opened 2020-10-13
Fix #479: Missing localization for error codes ERR0031 and ERR0032

### LINKED ISSUE #479: Missing localization for error codes ERR0031 and ERR0032
We need to add the localization strings to this [property file](https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/resources/i18n/errors.properties) to show a meaningful exception.

```java
/**
 * Activation OTP doesn't match value stored in database, or the OTP value is provided for the wrong validation mode.
 */
 public static final String INVALID_ACTIVATION_OTP = "ERR0031";

/**
 * Activation OTP operation is performed for the wrong validation mode.
 */
 public static final String INVALID_ACTIVATION_OTP_MODE = "ERR0032";
```

petrdvorak:


romanstrobl (approved):
Looks good, thanks for catching this one.

PR #483, opened 2020-11-09
Fix exceptions related to incorrect crypto provider initialization

romanstrobl:
Fixed new exceptions thrown when random generator is not initialized due to missing or invalid Bouncy Castle library initialization


petrdvorak (approved):
Looks OK

PR #504, opened 2020-12-06
Fix #503: Migrate WebClient to rest-client-base

### LINKED ISSUE #503: Migrate WebClient to rest-client-base
Migrate the following classes to the `rest-client-base` library which provides a common implementation of the REST client:
- `CallbackUrlBehavior`
- `PowerAuthRestClient`

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/CallbackUrlBehavior.java
```diff
@@ -91,9 +87,9 @@ public void setConfiguration(PowerAuthServiceConfiguration configuration) {
     }
 
     /**
-     * Configure WebClient HTTP connection parameters.
+     * Configure Rest client HTTP connection parameters.
      */
-    private void configureWebClient() {
+    private void configureRestClient() {
```
petrdvorak, line 92: Do we really need this method? It is only used once, right before initializing the client, and the only purpose is to store values from configuration to local properties, that I also think we no longer need.

petrdvorak (changes requested):
There is some left-over code that probably could be cleaned up a bit - let's do it with this change...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/CallbackUrlBehavior.java
romanstrobl: True, we can set the configuration directly.

petrdvorak (approved):
Looks OK

PR #522, opened 2020-12-17
Fix #521: Missing activation flags in the signature/token verification service

### LINKED ISSUE #521: Missing activation flags in the signature/token verification service

petrdvorak:


romanstrobl (approved):
Looks good

PR #533, opened 2020-12-21
Initial commit of the operations support

petrdvorak:
- [x] DB Schema for PostgreSQL
- [x] DB Schema for Oracle
- [x] Externalize cleanup scheduler configuration
- [x] Implement missing RESTful Client interface methods
- [x] Implement operation state change callbacks
- [x] Remove hardcoded strings by enums in operation service


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
```diff
@@ -0,0 +1,391 @@
...
+    private static final long serialVersionUID = -5284589668386509303L;
+
+    @Id
+    @Column(name = "id", updatable = false, length = 37)
+    private String id;
+
+    @Column(name = "user_id")
+    private String userId;
+
+    @Column(name = "application_id")
```
romanstrobl, line 50: Is there a reason why we do not use a `@JoinColumn` here?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
```diff
@@ -0,0 +1,391 @@
...
+    public void setTimestampFinalized(Date timestampFinalized) {
+        this.timestampFinalized = timestampFinalized;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof OperationEntity)) return false;
+        OperationEntity that = (OperationEntity) o;
+        return Objects.equals(id, that.id)
```
romanstrobl, line 337: I am not sure about this implementation of equals. For instance by increasing the failure counter the object becomes a different object. I think it is safer to use equals with fields which define the object and not use fields which are its attributes which change frequently.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
```diff
@@ -0,0 +1,391 @@
...
+                && Objects.equals(failureCount, that.failureCount)
+                && Objects.equals(maxFailureCount, that.maxFailureCount)
+                && Objects.equals(timestampCreated, that.timestampCreated)
+                && Objects.equals(timestampExpires, that.timestampExpires)
+                && Objects.equals(timestampFinalized, that.timestampFinalized);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(
```
romanstrobl, line 355: Same issue as with equals.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Behavior class implementing the operation related processes.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Service
+public class OperationBehavior {
```
romanstrobl, line 56: The class name is inconsistent with other class names, which use `*ServiceBehavior` pattern.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+                        final OperationDetailResponse operationDetailResponse = convertFromEntity(savedEntity);
+
+                        OperationUserActionResponse response = new OperationUserActionResponse();
+                        response.setResult(UserActionResult.OPERATION_FAILED);
+                        response.setOperation(operationDetailResponse);
+                        return response;
+                    }
+                }
+
+            } else {
```
romanstrobl, line 227: From style point of view, I think it is better to check for the negative cases in the beginning of the logic. This way two levels of nesting can be removed and the code is easier to read.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+                    final OperationDetailResponse operationDetailResponse = convertFromEntity(savedEntity);
+
+                    OperationUserActionResponse response = new OperationUserActionResponse();
+                    response.setResult(UserActionResult.OPERATION_FAILED);
+                    response.setOperation(operationDetailResponse);
+                    return response;
+                }
+
+            } else {
+                throw localizationProvider.buildExceptionForCode(ServiceError.OPERATION_INVALID_STATE);
```
romanstrobl, line 318: From style point of view, I think it is better to check for the negative cases in the beginning of the logic. This way two levels of nesting can be removed and the code is easier to read.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+        final String operationId = request.getOperationId();
+        final Optional<OperationEntity> operationOptional = operationRepository.findOperation(operationId);
+        if (operationOptional.isPresent()) {
+            final OperationEntity operationEntity = expireOperation(operationOptional.get(), currentTimestamp);
+            if (OperationStatusDo.PENDING.equals(operationEntity.getStatus())) {
+                operationEntity.setStatus(OperationStatusDo.CANCELED);
+                final OperationEntity savedEntity = operationRepository.save(operationEntity);
+                return convertFromEntity(savedEntity);
+            } else {
+                throw localizationProvider.buildExceptionForCode(ServiceError.OPERATION_INVALID_STATE);
```
romanstrobl, line 338: From style point of view, I think it is better to check for the negative cases in the beginning of the logic. This way two levels of nesting can be removed and the code is easier to read.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -1043,6 +1050,46 @@ public RemoveApplicationRolesResponse removeApplicationRoles(Long applicationId,
         return removeApplicationRoles(request);
     }
 
+    @Override
+    public OperationDetailResponse createOperation(OperationCreateRequest request) throws PowerAuthClientException {
+        return callV3RestApi("/operation/create", request, OperationDetailResponse.class);
+    }
+
+    @Override
+    public OperationDetailResponse operationDetail(OperationDetailRequest request) throws PowerAuthClientException {
```
romanstrobl, line 1059: Can we add a variant of this method with operation ID as a parameter?

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -1043,6 +1050,46 @@ public RemoveApplicationRolesResponse removeApplicationRoles(Long applicationId,
...
+        return callV3RestApi("/operation/list/pending", request, OperationListResponse.class);
+    }
+
+    @Override
+    public OperationDetailResponse operationCancel(OperationCancelRequest request) throws PowerAuthClientException {
+        return callV3RestApi("/operation/cancel", request, OperationDetailResponse.class);
+    }
+
+    @Override
+    public OperationUserActionResponse operationApprove(OperationApproveRequest request) throws PowerAuthClientException {
```
romanstrobl, line 1079: Can we add a variant of this method with individual parameters?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureTypeConverter.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Converter
+@Component
+public class SignatureTypeConverter implements AttributeConverter<PowerAuthSignatureTypes[], String> {
+
+    @Override
+    public String convertToDatabaseColumn(PowerAuthSignatureTypes[] powerAuthSignatureTypes) {
```
romanstrobl, line 39: We should check for null value here, I think.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureTypeConverter.java
```diff
@@ -0,0 +1,58 @@
...
+        return Arrays.stream(powerAuthSignatureTypes)
+                .map(PowerAuthSignatureTypes::toString)
+                .collect(Collectors.joining(","));
+    }
+
+    @Override
+    public PowerAuthSignatureTypes[] convertToEntityAttribute(String signatures) {
```
romanstrobl, line 46: We should check for null value here, I think.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationTemplateEntity.java
```diff
@@ -0,0 +1,236 @@
...
+    public void setExpiration(Long expiration) {
+        this.expiration = expiration;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof OperationTemplateEntity)) return false;
+        OperationTemplateEntity that = (OperationTemplateEntity) o;
+        return Objects.equals(id, that.id)
```
romanstrobl, line 209: I believe this equals implementation is incorrect. The object is different before it is persisted into database and after persistence because the ID changes during persistence.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationTemplateEntity.java
```diff
@@ -0,0 +1,236 @@
...
+                && Objects.equals(operationType, that.operationType)
+                && Objects.equals(dataTemplate, that.dataTemplate)
+                && Arrays.equals(signatureType, that.signatureType)
+                && Objects.equals(maxFailureCount, that.maxFailureCount)
+                && Objects.equals(expiration, that.expiration);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(id, templateName, operationType, dataTemplate, maxFailureCount, expiration);
```
romanstrobl, line 220: Same issue as for equals.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+    public OperationUserActionResponse attemptApproveOperation(OperationApproveRequest request) throws GenericServiceException {
+        final Date currentTimestamp = new Date();
+
+        final String operationId = request.getOperationId();
+        final String userId = request.getUserId();
+        final Long applicationId = request.getApplicationId();
+        final String data = request.getData();
+        final String signatureType = request.getSignatureType();
+
+        // TODO: We might need a lock here!!!
```
romanstrobl, line 170: Indeed, I think that the lock would be useful here because the processing changes the operation state.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -1043,6 +1050,46 @@ public RemoveApplicationRolesResponse removeApplicationRoles(Long applicationId,
...
+        return callV3RestApi("/operation/approve", request, OperationUserActionResponse.class);
+    }
+
+    @Override
+    public OperationUserActionResponse failApprovalOperation(OperationFailApprovalRequest request) throws PowerAuthClientException {
```
romanstrobl, line 1084: Can we add a variant of this method with individual parameters?

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -1043,6 +1050,46 @@ public RemoveApplicationRolesResponse removeApplicationRoles(Long applicationId,
...
+        return callV3RestApi("/operation/detail", request, OperationDetailResponse.class);
+    }
+
+    @Override
+    public OperationListResponse operationList(OperationListForUserRequest request) throws PowerAuthClientException {
+        return callV3RestApi("/operation/list", request, OperationListResponse.class);
+    }
+
+    @Override
+    public OperationListResponse operationPendingList(OperationListForUserRequest request) throws PowerAuthClientException {
```
romanstrobl, line 1069: Can we add a variant of this method with individual parameters?

powerauth-java-server/src/main/resources/i18n/errors.properties
```diff
@@ -49,3 +49,9 @@ ServiceError.ERR0029=Recovery code configuration is missing or incomplete.
 ServiceError.ERR0030=Token timestamp is too old.
 ServiceError.ERR0031=Activation OTP does not match value stored in database.
 ServiceError.ERR0032=Activation OTP used in the wrong OTP validation mode.
+ServiceError.ERR0033=Operation template was not found
```
romanstrobl, line 52: Missing period.

powerauth-java-server/src/main/resources/i18n/errors.properties
```diff
@@ -49,3 +49,9 @@ ServiceError.ERR0029=Recovery code configuration is missing or incomplete.
...
+ServiceError.ERR0034=Operation was not found
```
romanstrobl, line 53: Missing period.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+        if (OperationStatusDo.PENDING.equals(source.getStatus())
+                && source.getTimestampExpires().before(currentTimestamp)) {
+            logger.info("Operation {} expired.", source.getId());
+            source.setStatus(OperationStatusDo.EXPIRED);
+            return operationRepository.save(source);
+        }
+        return source;
+    }
+
+    private boolean factorsAcceptable(PowerAuthSignatureTypes[] allowedSignatureTypes, PowerAuthSignatureTypes usedFactors) {
```
romanstrobl, line 456: We should probably add a null check here.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    public OperationUserActionResponse rejectOperation(OperationRejectRequest request) throws GenericServiceException {
+        final Date currentTimestamp = new Date();
+
+        final String operationId = request.getOperationId();
+        final String userId = request.getUserId();
+        final Long applicationId = request.getApplicationId();
+
+        // TODO: We might need a lock here!!!
```
romanstrobl, line 242: Indeed, I think that the lock would be useful here because the processing changes the operation state.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+            throw localizationProvider.buildExceptionForCode(ServiceError.OPERATION_REJECT_FAILURE);
+        }
+    }
+
+    public OperationUserActionResponse failApprovalOperation(OperationFailApprovalRequest request) throws GenericServiceException {
+        final Date currentTimestamp = new Date();
+
+        final String operationId = request.getOperationId();
+
+        // TODO: We might need a lock here!!!
```
romanstrobl, line 283: Indeed, I think that the lock would be useful here because the processing changes the operation state.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+
+    }
+
+    public OperationUserActionResponse attemptApproveOperation(OperationApproveRequest request) throws GenericServiceException {
+        final Date currentTimestamp = new Date();
+
+        final String operationId = request.getOperationId();
+        final String userId = request.getUserId();
+        final Long applicationId = request.getApplicationId();
+        final String data = request.getData();
```
romanstrobl, line 167: We should either add validation to the requests or check for null value.

romanstrobl (changes requested):
Nice addition of operation functionality to the server. The implementation needs a bit of polishing, see comments in the pull request.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationTemplateEntity.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Objects;
+
+/**
+ * Entity representing an operation template.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Entity
+@Table(name = "pa_operation_template")
+public class OperationTemplateEntity implements Serializable {
```
romanstrobl, line 36: Can we add `nullable=false` for values which are mandatory?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
```diff
@@ -0,0 +1,391 @@
...
+import java.util.Objects;
+
+/**
+ * Entity representing an operation for approval.
+ *
+ * @author Petr Dvorak, petr@wutra.com
+ */
+@Entity
+@Table(name = "pa_operation")
+public class OperationEntity implements Serializable {
```
romanstrobl, line 39: Can we add `nullable=false` for values which are mandatory?

romanstrobl (changes requested):

docs/sql/mysql/create_schema.sql
```diff
@@ -204,6 +204,42 @@ CREATE TABLE `pa_recovery_config` (
...
+    PRIMARY KEY (id)
+) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+--
+-- Create table for operation templates
+--
+CREATE TABLE pa_operation_template (
+    id bigint(20) NOT NULL,
+    template_name varchar(255) NOT NULL,
+    operation_type varchar(255) NOT NULL,
```
romanstrobl, line 235: The `operation_type` field name is a bit unfortunate, because it creates an inconsistency between Next Step operations and PowerAuth operations:
- In Next Step, we use `operation_type` in steps, the value is either `CREATE` or `UPDATE`, see: https://github.com/wultra/powerauth-webflow/blob/fc2b4b8aeab2fb9dcb46c48b7d0e1867ce45f261/docs/sql/mysql/create_schema.sql#L169
- The `type` suggests that there are several operation types available, hence an enumeration can be used. However, the value cannot be typized because we do not know the value upfront.
- We use `operation_name` field for the same purpose in the rest of PowerAuth stack. While this field name is also not perfect, I think it is better to keep the terminology for consistency reasons.

I would like to ask, if it is still possible to rename this field? It will be a bit hard to explain the relationship between `operation_name` and `operation_type` in the documentation. I understand that we may have already deployed this DB structure, but I think a migration is possible before the final release and we can avoid headaches in the future. 

Thank you for considering this difficult change.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: Fixed.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: Makes sense.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: Makes sense.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
```diff
@@ -0,0 +1,469 @@
...
+        OperationDetailResponse destination = new OperationDetailResponse();
+        destination.setId(source.getId());
+        destination.setUserId(source.getUserId());
+        destination.setApplicationId(source.getApplicationId());
+        destination.setTemplateName(source.getTemplate().getTemplateName());
+        destination.setExternalId(source.getExternalId());
+        destination.setOperationType(source.getOperationType());
+        destination.setData(source.getData());
+        try {
+            final TypeReference<HashMap<String,String>> typeRef = new TypeReference<HashMap<String,String>>() {};
```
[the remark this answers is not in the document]
petrdvorak: OK

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: Makes sense.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -1043,6 +1050,46 @@ public RemoveApplicationRolesResponse removeApplicationRoles(Long applicationId,
...
+        return callV3RestApi("/operation/list/pending", request, OperationListResponse.class);
+    }
+
+    @Override
+    public OperationDetailResponse operationCancel(OperationCancelRequest request) throws PowerAuthClientException {
```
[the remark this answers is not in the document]
petrdvorak: I am inclined to avoid such methods, we will make a better API anyways.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureTypeConverter.java
petrdvorak: ok

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureTypeConverter.java
petrdvorak: ok

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationTemplateEntity.java
petrdvorak: Makes sense.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationTemplateEntity.java
petrdvorak: Yes.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
petrdvorak: Yes.

docs/sql/mysql/create_schema.sql
```diff
@@ -204,6 +204,42 @@ CREATE TABLE `pa_recovery_config` (
...
 ) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+--
+-- Create table for operations
+--
+CREATE TABLE pa_operation (
+    id varchar(37) NOT NULL,
+    user_id varchar(255) NOT NULL,
+    application_id bigint(20) NOT NULL,
+    template_id bigint(20) NULL,
```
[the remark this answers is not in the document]
petrdvorak: This is a "soft reference column", we do not need to lookup template based on ID in code.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationApproveRequest.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+public class OperationApproveRequest {
+
+    private String operationId;
+    private String userId;
+    private Long applicationId;
+    private String data;
+    private String signatureType;
```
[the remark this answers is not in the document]
petrdvorak: Yes. 👍

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/response/OperationDetailResponse.java
```diff
@@ -0,0 +1,169 @@
...
+    private String id;
+    private String userId;
+    private Long applicationId;
+    private String templateName;
+    private String externalId;
+    private String operationType;
+    private String data;
+    private Map<String, String> parameters;
+    private OperationStatus status;
+    private List<String> signatureType;
```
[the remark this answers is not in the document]
petrdvorak: Yes.

docs/sql/mysql/create_schema.sql
```diff
@@ -204,6 +204,42 @@ CREATE TABLE `pa_recovery_config` (
   CONSTRAINT `FK_RECOVERY_CONFIG_APP` FOREIGN KEY (`application_id`) REFERENCES `pa_application` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+--
+-- Create table for operations
+--
+CREATE TABLE pa_operation (
+    id varchar(37) NOT NULL,
+    user_id varchar(255) NOT NULL,
+    application_id bigint(20) NOT NULL,
```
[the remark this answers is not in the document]
petrdvorak: Yes.

docs/sql/mysql/create_schema.sql
```diff
@@ -204,6 +204,42 @@ CREATE TABLE `pa_recovery_config` (
   CONSTRAINT `FK_RECOVERY_CONFIG_APP` FOREIGN KEY (`application_id`) REFERENCES `pa_application` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+--
+-- Create table for operations
+--
+CREATE TABLE pa_operation (
+    id varchar(37) NOT NULL,
```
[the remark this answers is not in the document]
petrdvorak: We have it like this everywhere: activation ID, integrations, ... probably just allowing 1 character extra.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
petrdvorak: Not a big reason to have it here but fixed.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
petrdvorak: No.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: 👍 I used the lock function for all methods that modify an operation state.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: 👍 I used the lock function for all methods that modify an operation state.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: 👍 I used the lock function for all methods that modify an operation state.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationBehavior.java
petrdvorak: We will fix this when removing the SOAP interface by using well-annotated request objects. The good part is that these services are never called publically, always from trusted systems.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/model/ServiceError.java
```diff
@@ -197,8 +197,59 @@
...
+
+    /**
+     * Operation cannot be approved.
+     */
+    public static final String OPERATION_REJECT_FAILURE = "ERR0038";
+
+    /**
+     * Operation template already exists.
+     */
+    public static final String OPERATION_TEMPLATE_ALREADY_EXISTS = "ERR0039";
```
romanstrobl, line 233: Errors 39-42 are missing in documentation.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/model/ServiceError.java
```diff
@@ -197,8 +197,59 @@
...
+
+    /**
+     * Operation template related error occurred.
+     */
+    public static final String OPERATION_TEMPLATE_ERROR = "ERR0042";
+
+
     public static List<String> allCodes() {
-        List<String> list = new ArrayList<>(20);
+        List<String> list = new ArrayList<>(40);
```
romanstrobl, line 252: Probably should be 42.

romanstrobl (approved):
I found several more minor issues and possible improvements. If you disagree, feel free to ignore, these are just small nitpicks.

docs/PowerAuth-Server-1.1.0.md
```diff
@@ -42,11 +34,203 @@ ALTER TABLE pa_recovery_config
...
+    data_template varchar(255) NOT NULL,
+    signature_type varchar(255) NOT NULL,
+    max_failure_count bigint(20) NOT NULL,
+    expiration bigint(20) NOT NULL,
+    PRIMARY KEY (id)
+) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+CREATE INDEX pa_operation_user ON pa_operation(user_id);
+
+CREATE INDEX pa_operation_ts_created_idx ON pa_operation(timestamp_created DESC);
```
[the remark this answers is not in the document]
petrdvorak: We mostly query with `ORDER BY timestamp_created DESC` but I will remove this. On the single-column queries, the benefit is minimal but the readability is much worse.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/validator/OperationApproveRequestValidator.java
```diff
@@ -0,0 +1,61 @@
...
+package com.wultra.security.powerauth.client.model.validator;
+
+import com.wultra.security.powerauth.client.model.request.OperationApproveRequest;
+
+/**
+ * Validator for OperationApproveRequest class.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+public class OperationApproveRequestValidator {
```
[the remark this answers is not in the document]
petrdvorak: I think we will be able to replace it with annotations when we migrate to the Lombok / new REST. For now, this is consistent with how we did validation back then...

powerauth-java-server/src/main/resources/xsd/PowerAuth-3.0.xsd
```diff
@@ -1064,6 +1064,7 @@
             <xs:sequence>
                 <xs:element name="applicationId" type="xs:long" minOccurs="1" maxOccurs="1"/>
                 <xs:element name="name" type="xs:string" minOccurs="1" maxOccurs="1"/>
+                <xs:element name="type" type="xs:string" minOccurs="1" maxOccurs="1"/>
```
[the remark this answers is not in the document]
petrdvorak: Added

docs/PowerAuth-Server-1.1.0.md
```diff
@@ -42,11 +34,203 @@ ALTER TABLE pa_recovery_config
...
+### MySQL
+
+```sql
+CREATE TABLE pa_operation (
+    id varchar(37) NOT NULL,
+    user_id varchar(255) NOT NULL,
+    application_id bigint(20) NOT NULL,
+    template_id bigint(20) NULL,
+    external_id varchar(255) NULL,
+    operation_type varchar(255) NOT NULL,
```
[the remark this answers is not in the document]
petrdvorak: Where would the FK point?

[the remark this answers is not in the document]
romanstrobl: Sorry, I misunderstood the template table use case.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/model/ServiceError.java
petrdvorak: Good catch... we need to redo this error handling one day...

romanstrobl: Yes, please.

PR #537, opened 2020-12-21
Fix #536: Fix error code mappings

### LINKED ISSUE #536: Fix error code mappings

petrdvorak:


romanstrobl (approved):
Thanks for spotting and fixing this bug!

PR #542, opened 2020-12-28
Fix #541: Fix PostgreSQL DDL to allow run on PostgreSQL >= 10

### LINKED ISSUE #541: Fix PostgreSQL DDL to allow run on PostgreSQL >= 10
The MAXVALUE for sequences should fit to the boundaries or stay undefined and default to max value of the sequence type.

https://github.com/wultra/powerauth-server/blob/develop/docs/sql/postgresql/create_schema.sql

Official doc:
https://www.postgresql.org/docs/9.6/sql-createsequence.html

petrdvorak:


romanstrobl (approved):
Looks good

PR #560, opened 2021-04-13
Fix #558: Rename POSTCARD_PRIVATE_KEY_ENCRYPTION column

### LINKED ISSUE #558: Rename POSTCARD_PRIVATE_KEY_ENCRYPTION column
The Oracle database has a 30-character limit on the column name.

petrdvorak:


docs/PowerAuth-Server-1.1.0.md
```diff
@@ -0,0 +1,25 @@
...
+- Other web containers: follow instructions for installing a global library for the web container
+
+For more details about installation of the library see [Installing Bouncy Castle](./Installing-Bouncy-Castle.md).
+
+## Apply Database Hotfix
+
+We renamed the `POSTCARD_PRIVATE_KEY_ENCRYPTION` column to `POSTCARD_PRIV_KEY_ENCRYPTION` in 1.0.1 bugfix version to account for the 30-character limit in the Oracle databases. If you are upgrading directly from 1.0.0 version and still use the old column name, make sure to apply the following additional change:
+
+ ```sql
+ALTER TABLE pa_recovery_config CHANGE postcard_private_key_encryption postcard_priv_key_encryption INT DEFAULT 0 NOT NULL;
```
romanstrobl, line 24: This syntax unfortunately does not work in PostgreSQL and Oracle.

romanstrobl (changes requested):
The UPDATE query does not work on PostgreSQL and Oracle. Perhaps we can use instructions from here: https://github.com/wultra/powerauth-webflow/blob/develop/docs/Web-Flow-1.0.0.md

docs/PowerAuth-Server-1.1.0.md
petrdvorak: Good catch, fixed.

romanstrobl (approved):
I tested the UPDATE queries and they work fine on all databases now.

PR #566, opened 2021-04-22
Fix #565: No PasswordEncoder error

### LINKED ISSUE #565: No PasswordEncoder error
The following error occurs with `restrictAccess=true` when using the REST API: `There is no PasswordEncoder mapped for the id "null"`.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #577, opened 2021-05-24
Fix #576: Migrate to WebClient version 5.3

### LINKED ISSUE #576: Migrate to WebClient version 5.3
Migration of callbacks requires:
- use `ResponseEntity<String>` instead of `ClientResponse` due to migration of `exchange` -> `exchangeToMono`
- the `onSuccess` consumer is called for `200` responses only, so all error handling is done via `onError` consumer

romanstrobl:


petrdvorak (approved):
Looks OK

PR #579, opened 2021-05-27
Fix #552: Add validation for InitActivation call to disallow 0 as max. failed attempts

### LINKED ISSUE #552: Add validation for InitActivation call to disallow 0 as max. failed attempts
We should enforce value greater than 0.

petrdvorak:


romanstrobl (approved):
Looks good.

PR #582, opened 2021-05-27
Fix #544: Simplify notify callback listeners method

### LINKED ISSUE #544: Simplify notify callback listeners method
Currently, we use:

```java
callbackUrlBehavior.notifyCallbackListeners(activation.getApplication().getId(), activation);
```

We can have just:

```java
callbackUrlBehavior.notifyCallbackListeners(activation);
```

Sometimes, we have:

```java
callbackUrlBehavior.notifyCallbackListeners(application.getId(), activation);
```

But this does not make a difference - the application ID is still the same...

petrdvorak:


romanstrobl (approved):
Yes, better like this.

PR #583, opened 2021-05-27
Fix #404: Include BC library in web application

### LINKED ISSUE #404: Include BC library in web application
Currently, the BC library is provided by the web container. This is problematic because clients often forget to upgrade the BC library in the web container. 

Given that BC classloader issues have been fixed in https://github.com/wultra/powerauth-crypto/issues/271 without any known issues, we should consider providing the BC library in the individual war files.

Once it is confirmed that this approach works correctly, we should switch to this approach in all web applications related to PowerAuth (e.g. Web Flow, Enrollment server, etc.).

Following scenarios need to be considered:
- deployment to Tomcat
- deployment to Wildfly
- deployment to WebLogic
- standalone execution

We should also update documentation and release notes to make sure this change is properly documented.

petrdvorak:


romanstrobl (approved):
Looks ok, but we should probably update: https://github.com/wultra/powerauth-server/blob/develop/docs/Deploying-Wildfly.md#bouncy-castle-installation

PR #597, opened 2021-06-03
Fix #529: Generate serialVersionUID for entity classes

### LINKED ISSUE #529: Generate serialVersionUID for entity classes
Some of our entities are missing the `serialVersionUID` field - we should enable inspection for this in IDE and fix it.

petrdvorak:


romanstrobl (approved):
Looks good

PR #598, opened 2021-06-03
Fix #527: Verify annotation over repository classes

### LINKED ISSUE #527: Verify annotation over repository classes
Some of our repository interfaces use `@Repository` annotation, some use `@Component` annotation, and some use no annotation. We should probably decide which one is right and stick with it.

petrdvorak:


romanstrobl (approved):
Nice cleanup

PR #606, opened 2021-06-08
Fix #605: Return consistent error on request validations

### LINKED ISSUE #605: Return consistent error on request validations

petrdvorak:


romanstrobl (approved):
Thanks, looks ok.

PR #611, opened 2021-06-08
Fix #609: Missing callback type in documentation

### LINKED ISSUE #609: Missing callback type in documentation
We should add the callback type to the documentation:
https://github.com/wultra/powerauth-server/blob/develop/docs/Database-Structure.md#application-callback-url-table

romanstrobl:


petrdvorak (approved):
Looks good, thank you for catching this.

PR #616, opened 2021-06-11
Fix #614: Add notice about two package names in the @ComponentScan

### LINKED ISSUE #614: Add notice about two package names in the @ComponentScan

petrdvorak:


romanstrobl (approved):
Looks good

PR #618, opened 2021-06-16
Fix #617: Invalid migration instructions for Oracle

### LINKED ISSUE #617: Invalid migration instructions for Oracle
The table names must not contain the `"` character, Oracle treats these table names as case sensitive.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #622, opened 2021-06-27
Fix #621: Make sure to release response body when calling a callback

### LINKED ISSUE #621: Make sure to release response body when calling a callback
Current implementation may lead to memory leaks.

petrdvorak:


romanstrobl (approved):
Implemented as discussed.

PR #660, opened 2022-01-20
Fix #658: Add documentation for configuration properties

### LINKED ISSUE #658: Add documentation for configuration properties
We should document public configuration properties for PowerAuth server which are configurable through `application.properties`.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #670, opened 2022-02-21
Fix #661: Allow storing operation context

### LINKED ISSUE #661: Allow storing operation context
We need to add another column for the pa_operation table to store, for example, information about IP address of approval / rejection, or a detailed rejection reason. The column would contain a serialized JSON, in the following structure (example, can be different):

```json
{
  "request": {
    "ip": "10.122.5.12"
  },
  "resultInfo": {
    "rejectReason": "INVALID_DATA"
  }
}

### PR DESCRIPTION
**Pull request by petrdvorak**

### CODE REVIEW

File: powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationApproveRequest.java [hunk 1]
```diff
@@ -32,6 +34,7 @@
     private Long applicationId;
     private String data;
     private SignatureType signatureType;
+    private Map<String, String> additionalData;
```
romanstrobl, line 37: Small suggestion: perhaps we can make the map `final` and initialize it with an empty `LinkedHashMap`, this way we can avoid potential nullability issues?

romanstrobl (approved):
I have one very minor improvement idea.

romanstrobl (changes requested):
We will need to take a look at the error handling, because the postcard recovery test (v31) fails with:
```
PowerAuthRecoveryTest.recoveryPostcardInvalidPukIndexTest:767 expected: not <null>
```

romanstrobl (approved):
Perfect.

PR #671, opened 2022-02-22
Fix #668: Using networking layer object in DB entity

### LINKED ISSUE #668: Using networking layer object in DB entity
We are using network layer object (`HttpAuthenticationPrivate `) in database entity (`CallbackUrlEntity`):

https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlEntity.java#L65

This does not look right and it raises automatic warings because the network object is not Serializable.

petrdvorak:


romanstrobl (approved):
Thanks, the code is much cleaner now.

PR #675, opened 2022-03-29
Fix scheduling for PowerAuth auditing

romanstrobl:
The problem was that the configuration was done in `WebServiceConfig extends WsConfigurerAdapter` which caused following issue:

```
Bean 'databaseAuditWriter' of type [com.wultra.core.audit.base.database.DatabaseAuditWriter] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
```

Due to this issue the task was not scheduled. Moving configuration to a new configuration class solves the issue.


petrdvorak (approved):
Looks good

PR #681, opened 2022-04-12
Fix #680: Change default activation validity to 5 minutes

### LINKED ISSUE #680: Change default activation validity to 5 minutes
We are repeatedly having problems with testing because of the short activity expiration. While most users of the released apps do not have issues, developers (who occasionally need to set a breakpoint, or perform additional tasks) do not make it in the 2 minute interval.

petrdvorak:


romanstrobl (approved):
Good idea.

PR #686, opened 2022-05-04
Add alternate methods for compatibility with existing code 

romanstrobl:
Add alternate methods for compatibility with existing code which calls the client methods without query parameters and HTTP headers.


petrdvorak (approved):
Looks OK

Issue #501, opened 2020-11-20
Add a column for the original activation version


petrdvorak:
Currently, activation only stores the current version. We should:

- audit the activation version in table `pa_activation_history`
  - add a new `version` column which stores the current activation version and is specified whenever an activation history event is audited
  - add a new `AdditionalInformation` constant `ACTIVATION_VERSION_CHANGED` for activation history audit, use it:
    - during activation prepare (`null` version -> specific version)
    - during upgrade commit (version changes to a newer one)
  - for other activation changes the current activation version is stored, it is retrieved from activation entity

romanstrobl:
The issue will have to wait for the next release.

banterCZ:
Expected _level of done_: Create activation 2.1 and upgrade to 3.1, the new version should be recorded in the history.

PR #690, opened 2022-05-09
Fix #689: Do not log ACTIVATION_VERSION_CHANGED when preparing the activation

### LINKED ISSUE #689: Do not log ACTIVATION_VERSION_CHANGED when preparing the activation
Currently, we incorrectly log activation status change reason whenever the activation goes to PENDING_COMMIT:

Screenshot 2022-05-09 at 14 53 16

https://github.com/wultra/powerauth-server/blame/32df0accfe80f46581ea9936cd818747157c5dea/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java#L948

petrdvorak:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/AdditionalInformation.java
```diff
@@ -24,24 +24,63 @@
...
-     * <li>during upgrade commit (version changes to a newer one)</li>
-     * </ul>
-     */
-    public static final String ACTIVATION_VERSION_CHANGED = "ACTIVATION_VERSION_CHANGED";
+    public static class Key {
+        
+        /**
+         * Key describing reasons for blocked activation.
+         */
+        public static final String BLOCKED_REASON = "BLOCKED_REASON";
```
banterCZ, line 32: When refactoring, can we change it to enum (maybe in s separate PR)?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -945,7 +945,7 @@ public PrepareActivationResponse prepareActivation(String activationCode, String
             String mac = BaseEncoding.base64().encode(responseCryptogram.getMac());
 
             // Persist activation report and notify listeners
-            activationHistoryServiceBehavior.saveActivationAndLogChange(activation, null, AdditionalInformation.ACTIVATION_VERSION_CHANGED);
```
banterCZ, line 948: Ok, so it seems that the most important change is this one.

banterCZ (approved):
Seems ok.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
banterCZ: Just reminding that in the original issue #501 there was:

> during activation prepare (null version -> specific version)

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/AdditionalInformation.java
petrdvorak: Let's do this in a separate issue. The benefit of the change is minimal.

PR #707, opened 2022-06-07
Fix #706: Failing test testCreateApplicationWithDuplicateName

### LINKED ISSUE #706: Failing test testCreateApplicationWithDuplicateName
Following error occurs during build:

```
[ERROR] Failures: 
[ERROR]   CreateApplicationTest.testCreateApplicationWithDuplicateName:56 Unexpected exception type thrown ==> expected: <org.springframework.dao.DataIntegrityViolationException> but was: <io.getlime.security.powerauth.app.server.service.exceptions.GenericServiceException>
```

romanstrobl:
The build succeeds now.

### DIFF
--- a/powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/CreateApplicationTest.java
+++ b/powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/CreateApplicationTest.java
@@ -52,8 +52,8 @@ public class CreateApplicationTest {
         String testId = UUID.randomUUID().toString();
         CreateApplicationRequest request = new CreateApplicationRequest();
         request.setApplicationId(testId);
-        assertDoesNotThrow(()-> powerAuthService.createApplication(request));
-        assertThrows(GenericServiceException.class, ()-> powerAuthService.createApplication(request));
+        assertDoesNotThrow(() -> powerAuthService.createApplication(request));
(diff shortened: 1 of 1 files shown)


powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/CreateApplicationTest.java
```diff
@@ -53,7 +53,7 @@ public void testCreateApplicationWithDuplicateName() {
         CreateApplicationRequest request = new CreateApplicationRequest();
         request.setApplicationId(testId);
         assertDoesNotThrow(()-> powerAuthService.createApplication(request));
-        assertThrows(DataIntegrityViolationException.class, ()-> powerAuthService.createApplication(request));
+        assertThrows(GenericServiceException.class, ()-> powerAuthService.createApplication(request));
```
banterCZ, line 56: Are we really using default formatter? Because in that case it should be `() ->`

banterCZ (approved):

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/CreateApplicationTest.java
banterCZ: Great! If we are already touching the line, why not to make it perfect? ;)

petrdvorak (approved):
Looks good

PR #712, opened 2022-06-08
Make possible to run PowerAuth Java Server as standalone

banterCZ:


romanstrobl (approved):
Looks good in general, however I wonder whether we need to allow port specification because when we run standalone many apps, there will be conflicts on default port `8080`?

banterCZ:
Yes, the point about ports is very good. PA as the main app is using default 8080. Enrollment is prepared for 8081 https://github.com/wultra/powerauth-server/pull/712 and I am preparing admin with 8082

PR #713, opened 2022-06-08
Document version history

petrdvorak:


romanstrobl (changes requested):
I do not think this migration script will work for Oracle. Oracle has no `ALTER COLUMN` as far as I know and we usually use different data types for Oracle (e.g. `VARCHAR2` and `NUMBER`).

romanstrobl (approved):
Looks much better now.

PR #718, opened 2022-06-09
Fix #715: Fix admin creation of application version

### LINKED ISSUE #715: Failed to create application version from Admin
PowerAuth Admin is unable to create a new application version. It seems that there's a missing parameter `applicationVersionId` in the request. The parameter is required since we moved to string based application ids.

The following exception is reported:
```
org.springframework.web.bind.MissingServletRequestParameterException: Required request parameter 'applicationVersionId' for method parameter type String is not present
```

banterCZ:


romanstrobl (approved):
This should do it, thanks for the fix.

PR #719, opened 2022-06-09
Replace micrometer StringUtils by spring dependency

banterCZ:


powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
```diff
@@ -386,9 +386,9 @@ public String applicationUpdateCallbackAction(
             String callbackUrl = allParams.get("callbackUrl");
             String callbackId = allParams.get("callbackId");
             String error = null;
-            if (name == null || name.trim().isEmpty()) {
+            if (!StringUtils.hasText(name)) {
```
banterCZ, line 389: Improve other validations as well.

romanstrobl (approved):
The `hasText` method is less understandable to me than original code, but ok.

banterCZ:
> The `hasText` method is less understandable to me than original code, but ok.

Well, Apache Commons have little bit nicer API but I did not want to introduce a new dependency https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-

PR #720, opened 2022-06-09
Replace @RequestMapping by composed variants

banterCZ:


powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ActivationController.java
```diff
@@ -66,7 +63,7 @@ public ActivationController(PowerAuthClient client) {
      * @param model Model with passed parameters.
      * @return "activations" view.
      */
-    @RequestMapping(value = "/activation/list")
+    @RequestMapping("/activation/list")
```
romanstrobl, line 66: I am not sure these `@RequestMapping`s are correct in the original code. We should probably specify a HTTP method here. We don't want this endpoint to be accessible via `DELETE` method. This issue is present in other classes, too.

romanstrobl (approved):
Good improvement, but we should also revise the usage of `RequestMapping` without specified HTTP methods, this does not seem correct.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ActivationController.java
banterCZ: 👍 But let's do it in a separate PR. It will be easier to review the code and better for history tracking.

PR #722, opened 2022-06-09
Limit request methods

banterCZ:


petrdvorak:
@banterCZ This is all very nice but again, I have no idea why we work on it now...

petrdvorak (changes requested):
- Have you tested the change with complete stack integration?
- Does the client library need updates accordingly?

petrdvorak (approved):
Looks OK - did the testing.

banterCZ:
> * Does the client library need updates accordingly?

Only admin controllers affected. I do not see any need to update client library. JSP stay all the same.

petrdvorak:
Merging.

PR #724, opened 2022-06-09
Fix #710: Inherit activation flags when recovering activation

### LINKED ISSUE #710: Inherit activation flags when recovering activation
When using recovery codes, activation should inherit flags from the associated activation. The associated code:

https://github.com/wultra/powerauth-server/blob/develop/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java#L1697

petrdvorak:
The pull request addresses two areas:

- We inherit activation flags when creating a new operation using recovery codes that are linked to an activation.
- We allow specifying if recovery codes should be created using `prepareActivation`/`createActivation`/`createActivationUsingRecoveryCode` call.

The reason is the following:

- When we have activation that uses flags to mark it is "partially completed", we must make sure descendant activation is also marked like this.
- When recovering activation on old mobile client versions that does not support "partially completed" activation flag yet, we want to have a mechanism in place to assure that the "partially completed" activation does not create recovery codes.

### DIFF
--- a/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
+++ b/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
@@ -187,16 +187,18 @@ public interface PowerAuthClient {
     /**
      * Call the prepareActivation method of the PowerAuth 3.0 Server interface.
      *
-     * @param activationCode     Activation code.
-     * @param applicationKey     Application key.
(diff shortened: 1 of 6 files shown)


banterCZ (approved):

romanstrobl (approved):
Ok. I am a little bit unsure whether we should always inherit the activation flags, but right now I don't see a reason why not to do so.

PR #739, opened 2022-06-20
Fix #711: Add information about database dialect configuration into migration guide

### LINKED ISSUE #711: Add information about database dialect configuration into migration guide
Since release `1.2.5` the default DB dialect is specified, otherwise Spring application would not start.

We need to document the configuration of database dialect in migration guide for clients who use Oracle, otherwise the default PostgreSQL dialect is used which causes invalid SQL queries to be generated by Hibernate when using Oracle.

romanstrobl:


docs/powerauth-server/PowerAuth-Server-1.3.0.md
```diff
@@ -260,3 +260,18 @@ We consider the 5-minute interval to still be safe, since the relatively high ac
...
 powerauth.service.crypto.activationValidityInMilliseconds=120000
 ```
+
+## Database Dialect Configuration
+
+The latest release of PowerAuth requires configuration of database dialect.
+
+The dialect is specified using following configuration property:
+```properties
+spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
```
banterCZ, line 270: We are using a deprecated value in the example. Could we change it?

banterCZ (approved):

docs/powerauth-server/PowerAuth-Server-1.3.0.md
romanstrobl: Sure, good idea.

PR #750, opened 2022-06-27
Fix #744: Add scheduled activation cleanup

### LINKED ISSUE #744: Add scheduled activation cleanup
Currently, the activations are REMOVED lazilly, on access via API. We can also implement a cleanup job, so that the removal activation callbacks are sent as soon as the registration expires, even if the process was terminated and there is no API call...

petrdvorak:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -117,6 +117,13 @@ public class PowerAuthServiceConfiguration {
     @Min(0)
     private int activationValidityBeforeActive;
 
+    /**
+     * How many milliseconds should the activation cleanup job look to the past.
+     */
+    @Value("${powerauth.service.scheduled.job.activationsCleanup.lookBackInMilliseconds}")
+    @Min(0)
+    private int activationsCleanupLookBackInMilliseconds;
```
banterCZ, line 125: What about to start using `java.time.Period` and do not use units in the property name?

petrdvorak: For activation expiration related events, we typically use millseconds explicitly, to provide the required precision.

PR #753, opened 2022-06-29
Fix #747: Duplicate migration instructions for auditing tables

### LINKED ISSUE #747: Duplicate migration instructions for auditing tables
The auditing tables are added twice:
https://github.com/wultra/powerauth-server/blob/develop/docs/PowerAuth-Server-1.2.5.md#create-auditing-tables
https://github.com/wultra/powerauth-server/blob/develop/docs/PowerAuth-Server-1.3.0.md#create-auditing-tables

romanstrobl:


petrdvorak (approved):
Looks OK

PR #755, opened 2022-06-29
Fix #754: Failing build for PowerAuth server

### LINKED ISSUE #754: Failing build for PowerAuth server
The build fails due to error:

```
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'powerauth.service.scheduled.job.activationsCleanup.lookBackInMilliseconds' in value "${powerauth.service.scheduled.job.activationsCleanup.lookBackInMilliseconds}"
```

romanstrobl:


petrdvorak (commented):
Is this really needed?

- We do not use default values in the config class.
- We do have the properties in the develop branch already.

PR #757, opened 2022-07-11
Fix #756: Add support for Active Directory LDAP to Admin

### LINKED ISSUE #756: Add support for Active Directory LDAP to Admin

petrdvorak:


powerauth-admin/src/main/webapp/WEB-INF/jsp/header.jsp
```diff
@@ -44,6 +44,9 @@
                 <ul class="nav navbar-nav navbar-right">
                     <li><a href="${pageContext.request.contextPath}/logout">Logout</a></li>
                 </ul>
+                <p class="navbar-text navbar-right gray">
+                    <strong><c:out value="${pageContext.request.userPrincipal.name}"/></strong>
```
romanstrobl, line 48: What if the `header.jsp` file was included from a non-authenticated context, e.g. from `error.jsp` due to an LDAP/AD error? Won't this produce a null pointer exception, or is `userPrincipal` always defined even for non-authenticated contexts?

romanstrobl (approved):
Looks ok, I have just one concern related to a potential null pointer exception in non-authenticated context.

PR #765, opened 2022-10-19
Fix #764: Return activation status in ValidateTokenResponse

### LINKED ISSUE #764: Return activation status in ValidateTokenResponse
The `ValidateTokenResponse` response should return activation status to make it consistent with `VerifySignatureResponse`.

romanstrobl:
Unify response data and logic for signature verification and token verification.

### DIFF
--- a/docs/PowerAuth-Server-1.4.0.md
+++ b/docs/PowerAuth-Server-1.4.0.md
@@ -1,0 +1,9 @@
+# Migration from 1.3.x to 1.4.0
+
+This guide contains instructions for migration from PowerAuth Server version `1.3.x` to version `1.4.0`.
+
+## Change in PowerAuth Token Verification
+
+In earlier versions of PowerAuth Server, the token verification endpoint `/rest/v3/token/validate` returned an error in case the activation used by the token was not active. In order to always return activation status as part of the response, we changed the endpoint behaviour and removed the error handling for inactive activations. This change unifies the business logic with signature verification endpoint.
+
+Adaptation to this change is required only in case this endpoint is called directly on PowerAuth server. In case you use the @PowerAuthToken annotation for token validation, no changes are required.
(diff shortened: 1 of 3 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/TokenBehavior.java
```diff
@@ -274,31 +276,26 @@ public ValidateTokenResponse validateToken(ValidateTokenRequest request) throws
...
             } else {
-                final ValidateTokenResponse response = new ValidateTokenResponse();
-                response.setTokenValid(false);
-                return response;
+                isTokenValid = tokenVerifier.validateTokenDigest(nonce, timestamp, tokenSecret, tokenDigest);
             }
+
+            final ValidateTokenResponse response = new ValidateTokenResponse();
+            response.setTokenValid(isTokenValid);
+            response.setActivationStatus(activationStatusConverter.convert(ActivationStatus.ACTIVE));
```
banterCZ, line 290: Do I understand it correctly, that if the activation is not active, we still return activation status `ACTIVE` and `tokenValid=false`? Is that desired?

`!ActivationStatus.ACTIVE.equals(activation.getActivationStatus())`

romanstrobl: No, it's incorrect, fixed.

banterCZ (approved):

docs/PowerAuth-Server-1.4.0.md
```diff
@@ -0,0 +1,9 @@
+# Migration from 1.3.x to 1.4.0
+
+This guide contains instructions for migration from PowerAuth Server version `1.3.x` to version `1.4.0`.
+
+## Change in PowerAuth Token Verification
+
+In earlier versions of PowerAuth Server, the token verification endpoint `/rest/v3/token/validate` returned an error in case the activation used by the token was not active. In order to always return activation status as part of the response, we changed the endpoint behaviour and removed the error handling for inactive activations. This change unifies the business logic with signature verification endpoint.
+
+Adaptation to this change is required only in case this endpoint is called directly on PowerAuth server. In case you use the @PowerAuthToken annotation for token validation, no changes are required.
```
banterCZ, line 9:
```suggestion
Adaptation to this change is required only in case this endpoint is called directly on PowerAuth server. In case you use the `@PowerAuthToken` annotation for token validation, no changes are required.
```

hvge (approved):

PR #781, opened 2022-12-01
Fix #780: Add schema configuration for shedlock

### LINKED ISSUE #780: Add schema configuration for shedlock
Currently shedlock uses no schema on SQL queries. This does not work correctly in case schema is required by the database for each query.

romanstrobl:
Database schema support for shedlock.

### DIFF
--- a/lombok.config
+++ b/lombok.config
@@ -1,0 +1,1 @@
+lombok.log.fieldName=logger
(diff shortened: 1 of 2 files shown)


banterCZ (approved):
Well done!

saalistaja (approved):
All right.

PR #799, opened 2022-12-17
Issues/owasp dependency check

zcgandcomp:
Normalize naming
Add secrets for snapshot build
Add OWASP dependency check workflow

### DIFF
--- a/.github/workflows/maven-deploy.yml
+++ b/.github/workflows/maven-deploy.yml
@@ -41,10 +41,12 @@ jobs:
   maven-deploy-public:
     if: ${{ github.event_name == 'workflow_dispatch' }}
     name: Deploy to ${{ inputs.environment }}
-    uses: wultra/wultra-infrastructure/.github/workflows/maven-deploy.yml@develop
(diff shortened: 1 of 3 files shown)


.github/workflows/maven-deploy.yml
```diff
@@ -21,15 +21,13 @@ on:
...
-     - 'develop'
-     - 'test/ci'
-
-
+      - 'develop'
+      - 'test/ci'
 
 jobs:
-  maven-deploy-internal:
+  maven-deploy-push:
```
banterCZ, line 28: The name `push` confuses me.

banterCZ (approved):

.github/workflows/maven-deploy.yml
romanstrobl: I agree, it is confusing because we also use `push` for the push server component in our naming.

romanstrobl (approved):
Approved.

.github/workflows/maven-deploy.yml
zcgandcomp: I will gladly rename it - this is a naming used from crypto (https://github.com/wultra/powerauth-crypto/pull/439/files)

PR #818, opened 2022-12-19
Fix #817: Review business logic for activation code revocation

### LINKED ISSUE #817: Review business logic for activation code revocation
Review the business logic during activation removal related to activation code revocation:
- extract recovery code revocation functionality into a new method
- when removing activation in `CREATED` or `PENDING_COMMIT` state, recovery codes should always be revoked
- apply recovery code revocation logic to activations which are removed using background service

romanstrobl:


petrdvorak (approved):
Looks OK, with one minor style change suggestion.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -1452,26 +1452,14 @@ public RemoveActivationResponse removeActivation(String activationId, String ext
...
-                            puk.setStatus(RecoveryPukStatus.INVALID);
-                            puk.setTimestampLastChange(now);
-                        }
-                    }
-                    recoveryCodeRepository.save(recoveryCode);
-                }
-            }
+        // Recovery codes are revoked in case revocation is requested, or always when the activation is in CREATED or PENDING_COMMIT state
+        if (revokeRecoveryCodes ||
+                (activation.getActivationStatus() == ActivationStatus.CREATED ||
```
zcgandcomp, line 1461: are we using == style or ActivationStatus.CREATED.equals(activation.getActivationStatus())?

petrdvorak: Good catch, we typically use equals...

petrdvorak (changes requested):
Let's unify enum comparison.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -1976,8 +1992,8 @@ public void expireOperations() {
         final ImmutableSet<ActivationStatus> activationStatuses = ImmutableSet.of(ActivationStatus.CREATED, ActivationStatus.PENDING_COMMIT);
         try (final Stream<ActivationRecordEntity> abandonedActivations = activationRepository.findAbandonedActivations(activationStatuses, lookBackTimestamp, currentTimestamp)) {
             abandonedActivations.forEach(activation -> {
-                logger.info("Removing abandoned activation with ID: {}, revoking recovery codes.", activation.getActivationId());
-                deactivatePendingActivation(currentTimestamp, activation, false);
+                logger.info("Removing abandoned activation with ID: {}", activation.getActivationId());
+                removeActivation(activation, null, true);
```
zcgandcomp, line 1996: removeActivation does not have same timestap guarding mechanism as deactivatePendingActivation. Are we not removing activation before expiration? Or this was redundant, as there is a special query for finding abandoned/expired Activations

petrdvorak: This is redundant due to way we search for activations but I tend to agree we should keep `deactivatePendingActivation`, since we use this method elsewhere for deactivating activations that were abandoned.

Another thing to consider is to cleanup the code since the methods are quite similar:

```java
private void deactivatePendingActivation(Date timestamp, ActivationRecordEntity activation, boolean isActivationLocked) {
    if ((activation.getActivationStatus().equals(ActivationStatus.CREATED) || activation.getActivationStatus().equals(ActivationStatus.PENDING_COMMIT)) && (timestamp.getTime() > activation.getTimestampActivationExpire().getTime())) {
        logger.info("Deactivating pending activation, activation ID: {}", activation.getActivationId());
        if (!isActivationLocked) {
            // Make sure activation is locked until the end of transaction in case it was not locked yet
            activation = repositoryCatalogue.getActivationRepository().findActivationWithLock(activation.getActivationId());
        }
        activation.setActivationStatus(io.getlime.security.powerauth.app.server.database.model.ActivationStatus.REMOVED);
        revokeRecoveryCodes(activation.getActivationId());
        activationHistoryServiceBehavior.saveActivationAndLogChange(activation);
        callbackUrlBehavior.notifyCallbackListenersOnActivationChange(activation);
    }
}
```

```java
public RemoveActivationResponse removeActivation(@NotNull ActivationRecordEntity activation, String externalUserId, boolean revokeRecoveryCodes) {
    logger.info("Processing activation removal, activation ID: {}", activation.getActivationId());
    activation.setActivationStatus(io.getlime.security.powerauth.app.server.database.model.ActivationStatus.REMOVED);
    // Recovery codes are revoked in case revocation is requested, or always when the activation is in CREATED or PENDING_COMMIT state
    if (revokeRecoveryCodes ||
            (activation.getActivationStatus() == ActivationStatus.CREATED ||
                    activation.getActivationStatus() == ActivationStatus.PENDING_COMMIT)) {
        revokeRecoveryCodes(activation.getActivationId());
    }
    activationHistoryServiceBehavior.saveActivationAndLogChange(activation, externalUserId);
    callbackUrlBehavior.notifyCallbackListenersOnActivationChange(activation);
    RemoveActivationResponse response = new RemoveActivationResponse();
    response.setActivationId(activation.getActivationId());
    response.setRemoved(true);
    return response;
}
```

petrdvorak: But for this change, I would just use the original method, only to provide the correct logging message to the log ("Deactivating pending activation, ...").

romanstrobl:
> Let's unify enum comparison.

Enums are null safe, there is no reason for using `equals`. @banterCZ Please add your comment

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
romanstrobl: Ok, I will look into this.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
romanstrobl: There is no point in using equals for enums, enumerations are null safe. @banterCZ Please add your comment.

banterCZ:
> > Let's unify enum comparison.
> 
> Enums are null safe, there is no reason for using `equals`. @banterCZ Please add your comment.

Exactly
> Testing equality of an enum value with equals is perfectly valid because an enum is an Object and every Java developer knows "==" should not be used to compare the content of an Object. At the same time, using "==" on enums:
>
> - provides the same expected comparison (content) as equals
> - is more null-safe than equals()
> - provides compile-time (static) checking rather than runtime checking
> For these reasons, use of "==" should be preferred to equals.
https://rules.sonarsource.com/java/RSPEC-4551

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
banterCZ: Exactly
> Testing equality of an enum value with equals is perfectly valid because an enum is an Object and every Java developer knows "==" should not be used to compare the content of an Object. At the same time, using "==" on enums:
>
> - provides the same expected comparison (content) as equals
> - is more null-safe than equals()
> - provides compile-time (static) checking rather than runtime checking
> For these reasons, use of "==" should be preferred to equals.

https://rules.sonarsource.com/java/RSPEC-4551

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -1962,12 +1951,39 @@ private ActivationRecovery createRecoveryCodeForActivation(ActivationRecordEntit
...
+     * Revoke recovery codes for an activation entity.
+     * @param activationId Activation identifier.
+     */
+    private void revokeRecoveryCodes(String activationId) {
+        logger.info("Revoking recovery codes for activation ID: {}", activationId);
+        final RecoveryCodeRepository recoveryCodeRepository = repositoryCatalogue.getRecoveryCodeRepository();
+        final List<RecoveryCodeEntity> recoveryCodeEntities = recoveryCodeRepository.findAllByActivationId(activationId);
+        final Date now = new Date();
+        for (RecoveryCodeEntity recoveryCode : recoveryCodeEntities) {
+            logger.debug("Revoking recovery code: {} for activation ID: {}", recoveryCode.getRecoveryCode(), activationId);
```
banterCZ, line 1964: Ok, I double checked that `RecoveryCodeEntity#toString` does not contain `recoveryCode` itself. So we may log the whole entity if the ID does not give enough context.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
romanstrobl: I cleaned up the code and improve code reuse.

petrdvorak (approved):
Looks good

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationServiceBehavior.java
```diff
@@ -1962,12 +1940,58 @@ private ActivationRecovery createRecoveryCodeForActivation(ActivationRecordEntit
...
+     */
+    private void removeActivationInternal(ActivationRecordEntity activation, String externalUserId, boolean revokeRecoveryCodes) {
+        activation.setActivationStatus(io.getlime.security.powerauth.app.server.database.model.ActivationStatus.REMOVED);
+        // Recovery codes are revoked in case revocation is requested, or always when the activation is in CREATED or PENDING_COMMIT state
+        revokeRecoveryCodes = revokeRecoveryCodes
+                || activation.getActivationStatus() == ActivationStatus.CREATED
+                || activation.getActivationStatus() == ActivationStatus.PENDING_COMMIT;
+        if (revokeRecoveryCodes) {
+            revokeRecoveryCodes(activation.getActivationId());
+        }
```
banterCZ, line 1957: Please, let's not assign values to the parameter.

```suggestion
    private void removeActivationInternal(final ActivationRecordEntity activation, final String externalUserId, final boolean revokeRecoveryCodes) {
        activation.setActivationStatus(io.getlime.security.powerauth.app.server.database.model.ActivationStatus.REMOVED);
        // Recovery codes are revoked in case revocation is requested, or always when the activation is in CREATED or PENDING_COMMIT state
        if (revokeRecoveryCodes
                || activation.getActivationStatus() == ActivationStatus.CREATED
                || activation.getActivationStatus() == ActivationStatus.PENDING_COMMIT) {
            revokeRecoveryCodes(activation.getActivationId());
        }
```

banterCZ (approved):
Looks good.

PR #819, opened 2022-12-21
Fix #759: Optimize query for getting expired activations

### LINKED ISSUE #759: Optimize query for getting expired activations
Currently, getting the list of expired activations is slow because the planning goes via a sequential scan. Adding an index over expiration timestamp (and possibly also activation status) should help significantly.

```sql
create index pa_activation_expiration on pa_activation (timestamp_activation_expire, status)
```

romanstrobl:


petrdvorak (approved):
Looks OK

banterCZ (approved):
Looks good.

PR #820, opened 2022-12-21
Fix #760: Optimize query for getting expired operations

### LINKED ISSUE #760: Optimize query for getting expired operations
The current query is very inefficient since it looks for distinct records and orders the results by `timestamp_created`. We should optimize the query and add appropriate indexes to the table.

Also, see #591.

romanstrobl:


docs/sql/oracle/create_schema.sql
```diff
@@ -399,11 +399,13 @@ CREATE UNIQUE INDEX PA_APPLICATION_NAME ON PA_APPLICATION(NAME);
 
 CREATE INDEX PA_OPERATION_USER ON PA_OPERATION(USER_ID);
 
-CREATE INDEX PA_OPERATION_TS_CREATED_IDX ON PA_OPERATION(TIMESTAMP_CREATED);
+CREATE INDEX PA_OPERATION_TS_CREATED ON PA_OPERATION(TIMESTAMP_CREATED);
```
banterCZ, line 402: May we edit existing indexes?

banterCZ (approved):
Looks good.

docs/sql/oracle/create_schema.sql
romanstrobl: Good point, I'll revert the change and leave the inconsistency. We can deal with it later when introducing Liquibase.

Issue #824, opened 2022-12-22
Maven 'version' contains an expression but should be a constant


banterCZ:
```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for io.getlime.security:powerauth-client-model:jar:1.5.0-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ io.getlime.security:powerauth-server-parent:${powerauth.version}, /Users/macmini/.jenkins/jobs/PowerAuth Build/jobs/Build powerauth-server/workspace/pom.xml, line 29, column 14
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for io.getlime.security:powerauth-server-parent:pom:1.5.0-SNAPSHOT
[WARNING] 'version' contains an expression but should be a constant. @ io.getlime.security:powerauth-server-parent:${powerauth.version}, /Users/macmini/.jenkins/jobs/PowerAuth Build/jobs/Build powerauth-server/workspace/pom.xml, line 29, column 14
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

PR #829, opened 2023-01-03
Stop using revision property for project version

banterCZ:
Follow up to issue #824


romanstrobl (approved):
Alright, let's switch back.

PR #839, opened 2023-01-26
Update CodeQL to v2

banterCZ:


.github/workflows/codeql-analysis.yml
```diff
@@ -18,50 +18,59 @@ on:
...
+      security-events: write
 
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'java', 'javascript' ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
-        # Learn more:
-        # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
+        language: [ 'java' ]
```
romanstrobl, line 35: We may want to keep `javascript`, it is used lightly in PowerAuth Admin.

romanstrobl (approved):
Approved, one suggestion related to JavaScript.

.github/workflows/codeql-analysis.yml
banterCZ: Sorry, it was just a copy & paste

PR #854, opened 2023-02-17
Fix #853: ActivationFlagsServiceBehavior does not set timestampLastChange

### LINKED ISSUE #853: ActivationFlagsServiceBehavior does not set timestampLastChange

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/ActivationFlagsServiceBehavior.java
```diff
@@ -191,6 +193,7 @@ public RemoveActivationFlagsResponse removeActivationFlags(String activationId,
                 .build();
         audit.log(AuditLevel.INFO, "Removing activation flags: {} from activation {}", auditDetail, activationFlags, activationId);
         activation.getFlags().removeAll(activationFlags);
+        activation.setTimestampLastChange(new Date());
```
banterCZ, line 196: - [ ] As discussed with @petrdvorak, this change is not enough. It has to be connected with ActivationHistory.

romanstrobl (changes requested):
We'll need to finish work on this pull request first, then I can accept it.

PR #855, opened 2023-02-19
Fix #830: Allow overriding operation expiration timestamp

### LINKED ISSUE #830: Allow overriding operation expiration timestamp
Currently, the operation expiration timestamp is determined by the operation template. We should allow overriding the value by adding a new attribute `expirationTimestamp`. In case the attribute is `null`, the template value is applied.

```json
{
  "requestObject": {
    "userId": "string",
    "applications": [
      "string"
    ],
    "activationFlag": "string",
    "templateName": "string",
    "expirationTimestamp": "timestamp",
    "externalId": "string",
    "parameters": {
      "additionalProp1": "string",
      "additionalProp2": "string",
      "additionalProp3": "string"
    }
  }
}
```

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
```diff
@@ -102,6 +102,7 @@ public OperationDetailResponse createOperation(OperationCreateRequest request) t
         final List<String> applications = request.getApplications();
         final String activationFlag = request.getActivationFlag();
         final String templateName = request.getTemplateName();
+        final Date timestampExpiresRequest = request.getTimestampExpires();
```
petrdvorak, line 105: Validate the timestamp is not in the past.

petrdvorak (approved):
With one minor issue, this is good to go.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
romanstrobl: Good idea.

petrdvorak (changes requested):
The expiration timestamp in request should be validated early. Also, we should use the `currentTimestamp` value instead of `new Date()`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
```diff
@@ -139,9 +140,19 @@ public OperationDetailResponse createOperation(OperationCreateRequest request) t
...
+        // Resolve the operation expiration date
+        final Date timestampExpires;
+        if (timestampExpiresRequest != null) {
+            timestampExpires = timestampExpiresRequest;
+        } else {
+            final long expiration = templateEntity.getExpiration() * 1000L;
+            timestampExpires = new Date(currentTimestamp.getTime() + expiration);
+        }
+
+        if (timestampExpires.before(new Date())) {
```
petrdvorak, line 152: Use `currentTimestamp` value instead of `new Date()`.

petrdvorak: Also, the validation can probably happen early on in the request.

romanstrobl: Good points.

petrdvorak (approved):
Looks OK

PR #859, opened 2023-02-28
Fix #858: Prefer Java 8 APIs over Guava

### LINKED ISSUE #858: Prefer Java 8 APIs over Guava
Use `java.util.Base64` instead of `com.google.common.io.BaseEncoding#base64()`

See https://rules.sonarsource.com/java/tag/java8/RSPEC-4738

banterCZ:
Use `java.util.Base64` instead of `com.google.common.io.BaseEncoding#base64()`.

See https://rules.sonarsource.com/java/tag/java8/RSPEC-4738

### DIFF
--- a/powerauth-admin/src/main/java/io/getlime/security/app/admin/converter/SignatureAuditItemConverter.java
+++ b/powerauth-admin/src/main/java/io/getlime/security/app/admin/converter/SignatureAuditItemConverter.java
@@ -16,11 +16,11 @@
 
 package io.getlime.security.app.admin.converter;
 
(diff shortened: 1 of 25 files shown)


romanstrobl (changes requested):
I do not like this change, because we wanna stick to one version of Base-64 encoding across of all projects. Is it really necessary?

romanstrobl:
We decided to introduce this change gradually in all projects, where applicable, in order to move to the standard Base64 encoder.

romanstrobl (approved):
Let's move to Java 8 API for Base64 encoding.

PR #873, opened 2023-04-03
Fix #871: Signature audit log stores the next signature value when recording verification

### LINKED ISSUE #871: Signature audit log stores the next signature value when recording verification
Currently, the signature audit log contains the updated signature value with each record, not the value that was used for signature verification.

The previous value is still stored in the previous records, but it would be nice to store the counter values aligned. In principle, the issue is that the signature audit log is handled [here](https://github.com/wultra/powerauth-server/blob/eb12d146a176179734bbf6aa169ba696551b9aaf/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/SignatureSharedServiceBehavior.java#L443), just before we prepare an updated record and store it:

banterCZ:


romanstrobl (changes requested):
This is much better than previous commit. However, I still have a problem with this change. 
The audit record is created before the change is propagated to database and before the signature is actually verified.

I recommend to keep the audit log write after all signature verification tasks have been completed, just use the original counter values (numeric / hash-based) before their increment. We can use some temporary DTO for this, or just keep them as temporary variables.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/v3/SignatureSharedServiceBehavior.java
```diff
@@ -400,7 +400,8 @@ private void handleInvalidApplicationVersionImpl(ActivationRecordEntity activati
         }
 
         // Create the audit log record
-        auditingServiceBehavior.logSignatureAuditRecord(activation, signatureData, signatureType, false, activation.getVersion(),
+        final AuditingServiceBehavior.ActivationRecordDto activationDto = createActivationDtoFrom(activation);
```
banterCZ, line 403: - [ ] I have to decide what to do with`setActivationStatus`. I prefer to audit the value before the change.

banterCZ: Changed in a separate commit to better review.

romanstrobl (approved):
Yes, I was expecting a solution similar to this.

PR #880, opened 2023-04-12
Fix #869: Upgrade Spring Boot 3

### LINKED ISSUE #869: Upgrade Spring Boot 3
Analyze migration to Spring Boot 3 and upgrade all required project dependencies.

banterCZ:


pom.xml
```diff
@@ -80,34 +80,22 @@
     </modules>
 
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-
-        <!-- Maven Properties and Dependencies -->
-        <maven.compiler.source>1.8</maven.compiler.source>
```
banterCZ, line 86: Let's use implicit properties and versions from Spring Boot parent.

powerauth-java-server/src/main/webapp/WEB-INF/weblogic.xml
```diff
@@ -2,7 +2,7 @@
 <weblogic-web-app>
     <container-descriptor>
         <prefer-application-packages>
-            <package-name>javax.validation.*</package-name>
+            <package-name>jakarta.validation.*</package-name>
```
banterCZ, line 5: Very hypothetic change for Weblogix, since now JDK 17 is required.

romanstrobl (approved):
Excellent! We may need to look into increasing the `allocationSize` for generated identifiers using DB sequences later.

PR #882, opened 2023-04-12
Fix #881: Invalid JPA lock timeout setting

### LINKED ISSUE #881: Invalid JPA lock timeout setting
The following setting is incorrect:

```properties
# Database Lock Timeout Configuration
spring.jpa.properties.lock.timeout=10000
```

The correct configuration is:
```properties
# Database Lock Timeout Configuration
spring.jpa.properties.javax.persistence.lock.timeout=10000
```

romanstrobl:


petrdvorak (approved):
Looks OK

docs/PowerAuth-Server-0.18.0.md
```diff
@@ -90,7 +90,7 @@ We added row level DB locking for activations when handling concurrent signature
 
 ```properties
 # Database Lock Timeout Configuration
-spring.jpa.properties.lock.timeout=10000
+spring.jpa.properties.javax.persistence.lock.timeout=10000
```
banterCZ, line 93: What about keeping the wrong key in the old migration guide and mentioning it in `docs/PowerAuth-Server-1.5.0.md`?

banterCZ (approved):
Great!

docs/PowerAuth-Server-0.18.0.md
romanstrobl: I think we do not need this value in new migration guide, it's not being changed by any client. We only mentioned it when introducing the value.

PR #889, opened 2023-04-17
Fix #885: PowerAuthRestClient#prepareActivation throws NPE

### LINKED ISSUE #885: PowerAuthRestClient#prepareActivation throws NPE
Change `shouldGenerateRecoveryCodes` at `PowerAuthRestClient#prepareActivation` to primitive `boolean` to avoid `NullPointerException`. Double-check how was the `null` value handled before.

https://github.com/wultra/powerauth-server/blob/c7a23fa3ff4ee3d6dc8e8a8fbe02bceeeb935899/powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java#L250

banterCZ:


romanstrobl (approved):
Looks OK.

PR #891, opened 2023-04-17
Fix #890: AttributeConverter dirty checking does not work at Hibernate 6

### LINKED ISSUE #890: AttributeConverter dirty checking does not work at Hibernate 6
See https://hibernate.atlassian.net/browse/HHH-16081

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ListConverterMutabilityPlan.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * // TODO (racansky, 2023-04-17, https://hibernate.atlassian.net/browse/HHH-16081) workaround for non working-dirty checking
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public class ListConverterMutabilityPlan implements MutabilityPlan<List<String>> {
```
banterCZ, line 32: - [ ] Better rewrite to `MutableMutabilityPlan`.

romanstrobl (approved):
Good solution.

PR #893, opened 2023-04-18
Fix #892: Change shouldCreateRecoveryCodes to boolean

### LINKED ISSUE #892: Change shouldCreateRecoveryCodes to boolean
Follow-up to https://github.com/wultra/powerauth-restful-integration/issues/395 and #885

banterCZ:


powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -286,6 +286,7 @@ public CreateActivationResponse createActivation(String userId, Date timestampAc
         request.setEncryptedData(encryptedData);
         request.setMac(mac);
         request.setNonce(nonce);
+        request.setGenerateRecoveryCodes(true);
```
petrdvorak, line 289: I think we should leave this to default (`false`).

petrdvorak (changes requested):
We are probably changing the default behavior. Originally, if the flag was not included in the request (and was `null`), the recovery codes were generated. I don't have a strong opinion about the choices we have, but we should remain consistent if we make one.

- Currently, we have changed the default behavior. We should document it in migration guides to mention "If you want recovery codes to be generated, you need to explicitly say it."
- As an alternative, we can rename the property to `suppressRecoveryCode` which will allow us to keep the original behavior. This will also need to be documented.

My recommendation would be to just document the current behavior better, then this is good to go.

romanstrobl (approved):
Nice improvement, however I also agree default value should be `false` in PowerAuthRestClient.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
banterCZ: Set default `CreateActivationRequest#generateRecoveryCodes` to `true`.

PR #900, opened 2023-04-21
Remove OWASP Dependency Check GitHub Action

banterCZ:


romanstrobl (approved):
Alright, let's switch completely to Snyk and Dependabot.

PR #904, opened 2023-04-24
Fix #903: Clean-up maven dependencies

### LINKED ISSUE #903: Clean-up maven dependencies
Investigate not used dependencies via `mvn dependency:analyze -DignoreNonCompile`.

banterCZ:


powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
```diff
@@ -121,7 +117,7 @@ public String applicationDetail(@PathVariable("applicationId") String id, Map<St
             model.put("allowMultipleRecoveryCodes", recoveryConfig.isAllowMultipleRecoveryCodes());
             model.put("postcardPublicKey", recoveryConfig.getPostcardPublicKey());
             model.put("remotePostcardPublicKey", recoveryConfig.getRemotePostcardPublicKey());
-            model.put("versions", Lists.reverse(applicationDetails.getVersions()));
```
banterCZ, line 124: This was the only Guava function in PAS that has no direct equivalent in JDK.

romanstrobl (approved):
Nice cleanup!

PR #906, opened 2023-04-26
Fix #905: Add support for proximity anti-fraud check

### LINKED ISSUE #905: Add support for proximity anti-fraud check
- Generate and store TOTP seed when creating an operation (based on the template, may be overridden similar to timestamp, see #830)
- Return OTP for operation detail

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +726,57 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+            }
+        }
+        return null;
+    }
+
+    private static boolean proximityCheckPassed(final OperationEntity operation, final OperationApproveRequest request) {
+        if (operation.getTotpSeed() == null) {
+            return true;
+        }
+        final String otp = request.getAdditionalData().get("proximity_otp"); // TODO Lubos - constant, better name, import static additional data correct?
```
banterCZ, line 765: - [ ] Clarify, how to pass OTP value

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
         return m3;
     }
 
+    @SneakyThrows(GenericServiceException.class)
+    private void generateAndSetOtpToOperationDetail(final OperationEntity operation, final OperationDetailResponse operationDetailResponse) {
+        operationDetailResponse.setTotp(generateTotp(operation, powerAuthServiceConfiguration.getProximityCheckOtpLength()));
```
romanstrobl, line 734: Can we split the calls into two lines?

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationTemplateCreateRequest.java
```diff
@@ -32,12 +32,44 @@
...
+
+    /**
+     * Risk flags for offline QR code. Uppercase letters without separator, e.g. {@code XFC}.
+     */
     private String riskFlags;
 
+    /**
+     * Whether proximity check is enabled and TOTP seed should be generated
+     */
+    private boolean proximityCheckEnabled;
```
romanstrobl, line 73: Is the `boolean` type intentional? The `OperationCreateRequest` uses the `Boolean` type for same parameter name.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+    }
+
+    private static String generateTotp(final OperationEntity operation, final int otpLength) throws GenericServiceException {
+        final String seed = operation.getTotpSeed();
+        if (seed == null) {
+            return null;
+        }
+
+        try {
+            return new String(Totp.generateTotpSha256(Base64.getDecoder().decode(seed), LocalDateTime.now(), otpLength));
```
romanstrobl, line 744: Can we split these calls into multiple lines?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+            }
+        }
+        return null;
+    }
+
+    private static boolean proximityCheckPassed(final OperationEntity operation, final OperationApproveRequest request) {
+        if (operation.getTotpSeed() == null) {
+            return true;
+        }
+        final String otp = request.getAdditionalData().get("proximity_otp"); // TODO Lubos - constant, better name, is additional data correct?
```
romanstrobl, line 773: Can we fix this TODO before merging?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+        if (otp == null) {
+            logger.info("Proximity check enabled for operation ID: {} but OTP not sent", operation.getId());
```
romanstrobl, line 775: Shouldn't this log level be a warning? It seems to be an unusual state.

powerauth-java-server/src/main/resources/application.properties
```diff
@@ -107,4 +107,6 @@ powerauth.service.correlation-header.enabled=false
 powerauth.service.correlation-header.name=X-Correlation-ID
 powerauth.service.correlation-header.value.validation-regexp=[a-zA-Z0-9\\-]{8,1024}
 # For logging correlation HTTP headers enable the pattern and update correlation header name in the pattern
-#logging.pattern.console=%clr(%d{${LOG_DATEFORMAT_PATTERN:yyyy-MM-dd HH:mm:ss.SSS}}){faint} %clr(${LOG_LEVEL_PATTERN:%5p}) [%X{X-Correlation-ID}] %clr(%5p) %clr(${PID: }){magenta} %clr(---){faint} %clr([%15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n${LOG_EXCEPTION_CONVERSION_WORD:%wEx}
\ No newline at end of file
+#logging.pattern.console=%clr(%d{${LOG_DATEFORMAT_PATTERN:yyyy-MM-dd HH:mm:ss.SSS}}){faint} %clr(${LOG_LEVEL_PATTERN:%5p}) [%X{X-Correlation-ID}] %clr(%5p) %clr(${PID: }){magenta} %clr(---){faint} %clr([%15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n${LOG_EXCEPTION_CONVERSION_WORD:%wEx}
+
+powerauth.service.proximity-check.otp.length=6
```
romanstrobl, line 112: Can we use a longer OTP for proximity check by default?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+            } catch (CryptoProviderException e) {
+                logger.error("Unable to generate proximity OTP seed for operation, user ID: {}", request.getUserId(), e);
+                throw new GenericServiceException(ServiceError.OPERATION_ERROR, e.getMessage(), e.getLocalizedMessage());
+            }
+        }
+        return null;
+    }
+
+    private static boolean proximityCheckPassed(final OperationEntity operation, final OperationApproveRequest request) {
+        if (operation.getTotpSeed() == null) {
```
romanstrobl, line 770: Can we also add a paranoid check for an empty string as TOTP seed?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+            final KeyGenerator keyGenerator = new KeyGenerator();
+            try {
+                final byte[] seed = keyGenerator.generateRandomBytes(PROXIMITY_OTP_SEED_LENGTH);
+                return Base64.getEncoder().encodeToString(seed);
+            } catch (CryptoProviderException e) {
+                logger.error("Unable to generate proximity OTP seed for operation, user ID: {}", request.getUserId(), e);
+                throw new GenericServiceException(ServiceError.OPERATION_ERROR, e.getMessage(), e.getLocalizedMessage());
+            }
+        }
+        return null;
```
romanstrobl, line 766: The case when `request.getProximityCheckEnabled() == null` is not logged, perhaps we should log it as well?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+        if (operation.getTotpSeed() == null) {
+            return true;
+        }
+        final String otp = request.getAdditionalData().get("proximity_otp"); // TODO Lubos - constant, better name, is additional data correct?
+        if (otp == null) {
+            logger.info("Proximity check enabled for operation ID: {} but OTP not sent", operation.getId());
+            return false;
+        }
+        try {
+            boolean result = Totp.validateTotpSha256(otp.getBytes(), Base64.getDecoder().decode(operation.getTotpSeed()), LocalDateTime.now());
```
romanstrobl, line 779: What happens when TOTP seed could not be decoded from base64?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +729,62 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
     }
 
+    @SneakyThrows(GenericServiceException.class)
+    private void generateAndSetOtpToOperationDetail(final OperationEntity operation, final OperationDetailResponse operationDetailResponse) {
+        operationDetailResponse.setTotp(generateTotp(operation, powerAuthServiceConfiguration.getProximityCheckOtpLength()));
+    }
+
+    private static String generateTotp(final OperationEntity operation, final int otpLength) throws GenericServiceException {
+        final String seed = operation.getTotpSeed();
+        if (seed == null) {
```
romanstrobl, line 739: I suggest we add a debug log message for this state.

romanstrobl (approved):
Looks good in general, however I have several minor comments and concerns for discussion.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Sure.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Sounds reasonable.

powerauth-java-server/src/main/resources/application.properties
banterCZ: Ok, let's increase it to 8.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Well, I used to have it there but we may pass an empty string to `validateTotpSha256` that returns `false`. So you want to interpret an empty string as `null` and return `true`?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Good idea.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
romanstrobl: I would like to add a check that when the TOTP value is shorter than some minimal configuration (e.g. 6 digits) we reject this value immediately as invalid.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationTemplateCreateRequest.java
banterCZ: Yes, for `OperationCreateRequest` it is optional, three-state logic, for `OperationTemplateCreateRequest`, you have to decide. Would you prefer the same three-state logic?

romanstrobl: I just wanted to confirm this is interntional.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: I have added startup validation to `PowerAuthServiceConfiguration`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: - [ ] Clarify, how to pass OTP value

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +734,102 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+            audit.log(AuditLevel.INFO, "Proximity OTP generated for operation ID: {}", auditDetail, operationId);
+
+            return new String(totp);
+        } catch (CryptoProviderException | IllegalArgumentException e) {
+            logger.error("Unable to generate OTP for operation ID: {}, user ID: {}", operationId, operation.getUserId(), e);
+            throw new GenericServiceException(ServiceError.OPERATION_ERROR, e.getMessage(), e.getLocalizedMessage());
+        }
+    }
+
+    private static String generateTotpSeed(final OperationCreateRequest request, final OperationTemplateEntity template) throws GenericServiceException {
```
petrdvorak, line 767: Opinion: I think logging inside this method is overly excessive - do we really need this much information and add two parameters to the method that can work with a simple boolean?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -716,6 +734,102 @@ private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String>
...
+        }
+
+        final String otp = request.getAdditionalData().get(PROXIMITY_OTP);
+        if (otp == null) {
+            logger.warn("Proximity check enabled for operation ID: {} but proximity OTP not sent", operation.getId());
+            return false;
+        }
+        try {
+            final LocalDateTime now = LocalDateTime.now();
+            boolean result = Totp.validateTotpSha256(otp.getBytes(), Base64.getDecoder().decode(seed), now);
```
petrdvorak, line 799: I am missing the expected length in this method, it probably will just result in one if-else check.

petrdvorak (changes requested):
The length check must be in validation. Not sufficient to only enforce it during the generating of the OTP.

petrdvorak (changes requested):
Looks OK but cannot approve until the crypto methods for TOTP validation are extended by the length parameter.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Validation of the length added.

petrdvorak (approved):
Looks OK

PR #910, opened 2023-05-16
Fix #909: Coverity: Found reliance on default encoding

### LINKED ISSUE #909: Coverity: Found reliance on default encoding
`OperationServiceBehavior` creates TOTP string and does not explicitly specify the encoding.

Follow-up to #905

banterCZ:


romanstrobl (approved):
Good catch!

PR #912, opened 2023-05-17
Fix #874: Evaluate adding failed attempts counter and maximum failures to the activation status response

### LINKED ISSUE #874: Evaluate adding failed attempts counter and maximum failures to the activation status response
Currently, we do not return failed counter value in the activation status service:

https://developers.wultra.com/components/powerauth-server/1.4.x/documentation/WebServices-Methods#method-getactivationstatus

Missing values are in:

https://developers.wultra.com/components/powerauth-server/1.4.x/documentation/Database-Structure#table-pa_activation

- `failed_attempts`
- `max_failed_attempts`

We should allow adding this as the risk of incorrectly using the values is greatly decreased by using Enrollment Server for mobile activation status (people will unlikely use it to pass the information in plaintext), and it would help us implement several server-side scenarios.

petrdvorak:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/response/GetActivationStatusResponse.java
```diff
@@ -45,6 +45,8 @@ public class GetActivationStatusResponse {
     private String platform;
     private String deviceInfo;
     private String applicationId;
+    private Long failedAttempts;
```
banterCZ, line 48: Shouldn't it be mentioned in `WebServices-Methods.md`?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -438,6 +441,7 @@ public GetActivationStatusResponse getActivationStatus(String activationId, Stri
                     response.setPlatform(activation.getPlatform());
                     response.setDeviceInfo(activation.getDeviceInfo());
                     response.getActivationFlags().addAll(activation.getFlags());
+                    response.getApplicationRoles().addAll(application.getRoles());
```
banterCZ, line 444: 🤔 This was not mentioned in the issue nor it the commit message.

banterCZ (approved):
Looks good.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/response/GetActivationStatusResponse.java
petrdvorak: Good catch

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
petrdvorak: This is the result of inconsistency I fixed along the way - we return this value in list but not in detail, seemed like an oversight.

PR #913, opened 2023-05-17
Fix #785: Add billing telemetry to the server

### LINKED ISSUE #785: Add billing telemetry to the server
- Query for how many users used the system in a given period
- Query for how many authentications happened in a given period

petrdvorak:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/TelemetryServiceBehavior.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        if (days > 365) {
+            throw new UnknownTelemetryReportNameException("The parameter 'days' must be smaller than 365. Provided: " + days);
+        }
+        if (applicationId == null) {
+            throw new UnknownTelemetryReportNameException("Missing parameter 'application'.");
+        }
+
+        final Date toDate = new Date();
+        final Date fromDate = new Date(Instant.now().minus(days, ChronoUnit.DAYS).toEpochMilli());
```
banterCZ, line 102:
```suggestion
        final Date fromDate = Date.from(Instant.now().minus(days, ChronoUnit.DAYS));
```

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/TelemetryServiceBehavior.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param parameters Report parameters.
+     * @return Number of unique app users in specified period.
+     */
+    private TelemetryReportResponse reportUsersInPastDays(String reportName, Map<String, Object> parameters) throws UnknownTelemetryReportNameException {
+        final Integer days = (Integer) parameters.get(PARAM_DAYS);
+        final String applicationId = (String) parameters.get(PARAM_APPLICATION);
+
+        // Validate report parameters are present
+        if (days == null) {
+            throw new UnknownTelemetryReportNameException("Missing parameter 'days'.");
```
banterCZ, line 92: I would rather throw `jakarta.validation.ValidationException`, definitely not `UnknownTelemetryReportNameException`, or let's rename it to `TelemetryReportException`.

banterCZ (approved):
I suggest renaming the exception, otherwise good job!

romanstrobl (approved):
Nice feature!

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/TelemetryServiceBehavior.java
petrdvorak: Good catch, this is a glitch in naming, I will fix it.

PR #914, opened 2023-05-22
Fix #907: Crypto 3.2: Simplified configuration

### LINKED ISSUE #907: Crypto 3.2: Simplified configuration
Encode configuration for Mobile Token SDK using the new format specified below and encode it into Base64.

### Configuration

| Field name | Data type | Description |
|------------|-----------|-------------|
| `VERSION`     | Byte(1)   | Version of data format. Constant `0x01` |
| `APP_KEY`     | Byte(16)  | Binary data for APP_KEY constant |
| `APP_SECRET`  | Byte(16)  | Binary data for APP_SECRET constant |
| `KEYS_COUNT`  | Count     | Number of keys encoded in the next few bytes |
| 1st key       | Key       | First key |
| Nth key       | Key       | N-th key  |

The sequence of bytes described above is then encoded to one Base64 string for the use in the mobile SDK.

### Key identifiers

| Key ID | Description |
|--------|-------------|
| `0x01` | MASTER_SERVER_PUBLIC_KEY on P-256 curve, required for protocol 3.2+ |
| `0x02` | MASTER_SERVER_PUBLIC_KEY on EC-NEXT curve, required for protocol 4.0+ |

Example implementation in C++:

- [DataWriter.h](https://github.com/wultra/powerauth-mobile-sdk/blob/develop/src/PowerAuth/utils/DataWriter.h)
- [DataWriter.cpp](https://github.com/wultra/powerauth-mobile-sdk/blob/develop/src/PowerAuth/utils/DataWriter.cpp)  
- [DataReader.h](https://github.com/wultra/powerauth-mobile-sdk/blob/develop/src/PowerAuth/utils/DataReader.h)
- [DataReader.cpp](https://github.com/wultra/powerauth-mobile-sdk/blob/develop/src/PowerAuth/utils/DataReader.cpp)

romanstrobl:


banterCZ (approved):
Looks good.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/util/SdkDataWriterTest.java
```diff
@@ -0,0 +1,100 @@
...
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class SdkDataWriterTest {
+
+    @Test
+    public void testCountWrite() {
+        final String countsGeneratedBySdk = "AAF/gICA/4EAgQG//8AAQADAAP//wAEAAMD////BAAAAwQIDBNAgMED/////";
+        final SdkDataReader readerSdk = new SdkDataReader(Base64.decode(countsGeneratedBySdk));
+        final List<Long> expectedCounts = new ArrayList<>();
+        expectedCounts.add(0L);
```
hvge, line 42: I think that you can use `Integer` here. The mobile SDK is using `size_t` for "count" because its bit-length depends on the architecture. I would get a ton of warnings if I use `int` or `long` instead, because the C++ standard template library is using `size_t` for "size" or "count"

The maximum value for our serialized count never exceeds 32-bit integer, so we're OK with int.

hvge (approved):
Looks OK.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/util/SdkDataWriterTest.java
romanstrobl: Ah yes, I will fix this.

PR #919, opened 2023-05-23
Fix #915: Auditing of proximity anti-fraud check

### LINKED ISSUE #915: Auditing of proximity anti-fraud check
Improve auditing of proximity anti-fraud check. Generating seed and veryfying OTP should not be a separate audit entry, but an attribute of operation creation and operation approval. Generating OTP should not be audited so far.

Follow-up to #905

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -784,54 +790,32 @@ private static String generateTotpSeed(final OperationCreateRequest request, fin
         return null;
     }
 
-    private boolean proximityCheckPassed(final OperationEntity operation, final OperationApproveRequest request) {
+    private static boolean proximityCheckPassed(final ProximityCheckResult proximityCheckResult) {
+        return proximityCheckResult == ProximityCheckResult.SUCCESS || proximityCheckResult == ProximityCheckResult.DISABLED;
```
romanstrobl, line 794: Can we log the disabled proximity check in a different way than when it is disabled?

romanstrobl (approved):
I think the change is fine. Let's just not audit proximity check as successful in case it is disabled.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
petrdvorak: Good observation, we should probably just log the enum value there.

banterCZ: The enum value is already in the audit record `Operation approved with ID: bc4e8032-4dc5-4741-b4a4-0c5cf188db9b`
```json
{
  "id": "bc4e8032-4dc5-4741-b4a4-0c5cf188db9b",
  "userId": "TestUserV31_f2d593e7-08e8-41e7-9310-96361a74050f",
  "appId": "PA_Tests",
  "status": "APPROVED",
  "additionalData": {
    "proximity_otp": "89569078"
  },
  "failureCount": 0,
  "proximityCheckResult": "SUCCESS",
  "currentTimestamp": "2023-05-23T14:40:26.893+00:00"
}
```
So, we may also expect `"proximityCheckResult": "DISABLED"`. Is that acceptable?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -219,6 +217,9 @@ public OperationDetailResponse createOperation(OperationCreateRequest request) t
 
     public OperationUserActionResponse attemptApproveOperation(OperationApproveRequest request) throws GenericServiceException {
         final Date currentTimestamp = new Date();
+        final LocalDateTime currentLocalDateTime = currentTimestamp.toInstant()
```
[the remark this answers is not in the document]
banterCZ: As far as I know, the timezone should be set via VM to UTC, but if not, there is an issuer in `wultra-crypto`, which I have fixed here https://github.com/wultra/powerauth-crypto/pull/484

PR #931, opened 2023-07-12
Fix #930: Restrict token validation for timestamps in future

### LINKED ISSUE #930: Restrict token validation for timestamps in future
When checking token digests, we check for expired tokens. To improve system resiliency, we can also add validation that the timestamp is not too much in the future.

petrdvorak:


hvge (approved):
Looks OK

PR #933, opened 2023-07-17
Fix #932: Coverity: Dereference null return value

### LINKED ISSUE #932: Coverity: Dereference null return value
394602 [Dereference null return value](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_NULL_RETURNS)
If the function actually returns a null value, a NullPointerException will be thrown.

In io.​getlime.​security.​powerauth.​app.​server.​service.​util.​SdkConfigurationSerializer.​deserialize(java.​lang.​String): Return value of function which returns null is dereferenced without checking ([CWE-476](http://cwe.mitre.org/data/definitions/476.html))

banterCZ:


petrdvorak (approved):
Looks OK

Issue #950, opened 2023-08-01
Crypto 3.2: Do not generate nonce for older requests


hvge:
The current implementation in `develop` branch use `nonce` for response data encryption also for older client running on older protocol V3.1. The result is that such client is basically unable to communicate with the PowerAuth Server.

The expected behavior is that PowerAuth server support wide range of older mobile clients, still running on older protocol versions.

Issue #938, opened 2023-07-21
Crypto 3.2: Consider masking out ephemeral public key in ECIES responses


romanstrobl:
With unified logic for encryption of requests and responses, `ephemeralPublicKey` is always returned. The client should not rely on this parameter. We should consider masking out the parameter in responses as a slight performance optimization, the value is not required for responses, only requests.

hvge:
Changed from "enhancement" to a "bug".

PR #952, opened 2023-08-01
Broken ECIES 3.2

hvge:
This PR fixes two major problems in ECIES 3.2 implementation:

- Do not use nonce in encrypted response (#950)
- Do not return `ephemeralPublicKey` in encrypted responses (#938)


petrdvorak (approved):
Looks OK

banterCZ (approved):

PR #953, opened 2023-08-02
Use Duration instead of int in ConfigurationProperties

jandusil:


jandusil:
I am not sure about whether to create new migration file, or include it to PowerAuth-Server-1.5.0.md

docs/Configuration-Properties.md
```diff
@@ -28,9 +28,9 @@ The PowerAuth Server uses the following public configuration properties:
 
 | Property                                                      | Default   | Note                                                                                    |
 |---------------------------------------------------------------|-----------|-----------------------------------------------------------------------------------------|
-| `powerauth.service.crypto.activationValidityInMilliseconds`   | `120000`  | Default activation validity period in miliseconds                                       |
+| `powerauth.service.crypto.activationValidityTime`   | `PT2M`  | Default activation validity period (ISO 8601 Duration format) |
```
banterCZ, line 31: May we use the variant without `PT` prefix? Spring Boot can do that, see https://docs.spring.io/spring-boot/docs/2.1.13.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -196,21 +206,26 @@ public class PowerAuthServiceConfiguration {
...
     /**
-     * Token timestamp validity to future in milliseconds, checked before validating the token.
+     * Specifies the validity duration to future of token timestamps, checked before token validation.
+     * <p>
+     * The value is specified in the ISO-8601 duration format PTnHnMn.nS.
      */
     @Value("${powerauth.service.token.timestamp.forward.validity}")
-    @Min(0)
-    private long tokenTimestampForwardValidityInMilliseconds;
+    @DurationMin(millis = 0)
```
banterCZ, line 226: What is the difference between this and the previous format `@Min(0)`?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -686,7 +687,7 @@ public InitActivationResponse initActivation(String applicationId, String userId
             // Get activation expiration date from request or from constants, if not provided
             Date timestampExpiration = activationExpireTimestamp;
             if (timestampExpiration == null) {
-                timestampExpiration = new Date(timestamp.getTime() + powerAuthServiceConfiguration.getActivationValidityBeforeActive());
+                timestampExpiration = Date.from(Instant.now().plus(powerAuthServiceConfiguration.getActivationValidityBeforeActive()));
```
banterCZ, line 690: Are you sure that `timestamp` (before) is the same as `Instant.now()`? I think that there is a little time shift.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -431,7 +431,7 @@ private synchronized RestClient getRestClient(CallbackUrlEntity callbackUrlEntit
     private RestClient initializeRestClient(CallbackUrlEntity callbackUrlEntity) throws RestClientException {
         final DefaultRestClient.Builder builder = DefaultRestClient.builder();
         if (configuration.getHttpConnectionTimeout() != null) {
-            builder.connectionTimeout(configuration.getHttpConnectionTimeout());
+            builder.connectionTimeout((int) configuration.getHttpConnectionTimeout().toMillis());
```
banterCZ, line 434: Not much dangerous here, but better to use `Math#toIntExact()`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
jandusil: The @Min apparently wont work on type Duration

docs/Configuration-Properties.md
jandusil: While it is true that we can omit the 'PT' prefix and still have Spring Boot interpret the duration correctly, I believe we should consider the implications for broader public documentation and usage. By adhering to the ISO 8601 Duration format in its complete form, including the 'PT' prefix, we ensure that our format is standard-compliant and immediately recognizable to anyone familiar with ISO 8601.

banterCZ: Standards are fine. Let's wait for another opinion and we will decide.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -431,7 +431,7 @@ private synchronized RestClient getRestClient(CallbackUrlEntity callbackUrlEntit
     private RestClient initializeRestClient(CallbackUrlEntity callbackUrlEntity) throws RestClientException {
         final DefaultRestClient.Builder builder = DefaultRestClient.builder();
         if (configuration.getHttpConnectionTimeout() != null) {
-            builder.connectionTimeout(configuration.getHttpConnectionTimeout());
+            builder.connectionTimeout(Math.toIntExact(configuration.getHttpConnectionTimeout().toMillis()));
```
banterCZ, line 434: 🤔 I have opened another issue, what we may improve in the next step. https://github.com/wultra/lime-java-core/issues/201

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/ReplayPersistenceService.java
```diff
@@ -70,7 +70,7 @@ public boolean uniqueValueExists(final String uniqueValue) {
      * @return Whether unique value was added successfully.
      */
     public boolean persistUniqueValue(final UniqueValueType type, final String uniqueValue) {
-        final Instant expiration = Instant.now().plus(config.getRequestExpirationInMilliseconds(), ChronoUnit.MILLIS);
+        final Instant expiration = Instant.now().plus(config.getRequestExpiration().toMillis(), ChronoUnit.MILLIS);
```
banterCZ, line 73: Doesn't it work in a simpler form?
```suggestion
        final Instant expiration = Instant.now().plus(config.getRequestExpiration());
```

banterCZ (approved):
Good job (except my personal opinion about the duration format).
Let's wait for the fix of crypto 3.1 and run backend tests before merging.

PR #955, opened 2023-08-03
Fix #954: Use proper response nonce for 3.1 protocol

### LINKED ISSUE #954: Crypto 3.2: Use nonce from request for response
The current server implementation doesn't handle protocol 3.1 well. The problem is that the current implementation doesn't use nonce received with the request for the response encryption when does setup for `EciesEncryptor` object.

hvge:
This PR configures `EciesEncryptor` for the response encryption to use proper nonce for protocol V3.1


banterCZ (approved):
Looks good.

Issue #957, opened 2023-08-04
Crypto 3.2: Use high level encryptor and decryptor


hvge:
Use high level encryptor and decryptor created in wultra/powerauth-crypto#493 This will fix V3.2 protocol support

Issue #956, opened 2023-08-04
Crypto 3.2: /ecies/decryptor doesn't use ReplayVerificationService


hvge:
The endpoint implementing PowerAuth Server RESTapi for getting parameters for ECIES decryption `/rest/v3//ecies/decryptor` doesn't use `ReplayVerificationService` at all. That makes all custom application encrypted requests unprotected to replay attacks.

PR #965, opened 2023-08-10
Use general encryptor for E2EE

hvge:
This PR replaces usage of low level ECIES with a new general encryptor facility introduced in wultra/powerauth-crypto#493 

This PR will fix:
- #957
- #956

TODO:
- We have to properly check usage of replay verification service. It's clear that such service store data to database in case that entry is new (e.g. there's no replay attack), so all our comments about transaction rollback in exceptions appears to be inaccurate.
- Documentation

### DIFF
--- a/docs/PowerAuth-Server-1.5.0.md
+++ b/docs/PowerAuth-Server-1.5.0.md
@@ -127,3 +127,52 @@ Since version `1.5.0`, MySQL database is not supported anymore.
 
 PostgreSQL JDBC driver is already included in the WAR file.
 Oracle JDBC driver remains optional and must be added to your deployment if desired.
+
+## Configuration properties
+
+In previous versions of PowerAuth Server, time-based properties were represented as a long value, specifying the
+duration in milliseconds. From version `1.5.0` onwards, we've transitioned to using `java.time.Duration` for these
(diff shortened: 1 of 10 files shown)


hvge:
> I like the changes - still a bit anxious about the uniqueness verification service. Should we maybe allow configuring no-op service so that if we hit a performance issue, we can disable this?

We should address this in a separate ticket. There's still missing small fix that set the tolerance for timestamp validation to 1 minute (according to my original design). We'll still reduce the time window for the replay attack from infinity to 1 minute in case that verification service is off. That 1 minute tolerance can also be adjusted by configuration.

romanstrobl (approved):
Looks OK.

docs/PowerAuth-Server-1.5.0.md
```diff
@@ -127,3 +127,52 @@ Since version `1.5.0`, MySQL database is not supported anymore.
 
 PostgreSQL JDBC driver is already included in the WAR file.
 Oracle JDBC driver remains optional and must be added to your deployment if desired.
+
+## Configuration properties
```
banterCZ, line 131: Wrong resolution of merge conflict, fixed by https://github.com/wultra/powerauth-server/pull/1000

PR #979, opened 2023-08-22
Fix #978: Revert connection timeout

### LINKED ISSUE #978: Revert connection timeout properties
Revert changes introduced by #940 to still support API https://github.com/wultra/lime-java-core/issues/201 but keep original configuration properties.

banterCZ:


powerauth-java-server/src/main/resources/application.properties
```diff
@@ -65,25 +65,26 @@ powerauth.service.http.proxy.username=
 powerauth.service.http.proxy.password=
 
 # HTTP Connection Configuration
-powerauth.service.http.connection.timeout=5s
+# Without unit bind to milliseconds
+powerauth.service.http.connection.timeout=5000
```
banterCZ, line 69: 👀 Java type is `Duration`. Tested that this binding works.

romanstrobl (approved):
Better like this for now.

PR #982, opened 2023-08-23
Fix #981: Option to disable verification of replay attacks

### LINKED ISSUE #981: Option to disable verification of replay attacks
Add an option to disable verification of replay attacks due to performance reasons. Verification should be enabled by default.

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/ReplayVerificationService.java
```diff
@@ -41,6 +42,10 @@
  */
 @Service
 @Slf4j
+@ConditionalOnProperty(
```
banterCZ, line 45: Why not. Personally, I would prefer a solution with NoOpt implementation annotated something like `@ConditionalOnMissingBean(ReplayVerificationService.class)`. It helps us to avoid null-checks.

banterCZ (approved):
I am ok with the solution if IoC is working but I suggested another idea to consider.

romanstrobl:
I will close this pull request and implement this feature another way using an interface with multiple service implementations to avoid null checks.

PR #985, opened 2023-08-24
Fix #971: Add column for signature metadata

### LINKED ISSUE #971: Add column for signature metadata
PowerAuth server should support mutliple signature types, hence columns `signature_data_method` and `signature_data_uri_id` in table `pa_signature_audit` should be generified into a single JSON column e.g. `signature_metadata`.

Sample structure of contents of column `signature_metadata` after migration to JSON format:
```json
{
    "signature_data_method": "POST",
    "signature_data_uri_id": "/authorize"
}
```

Other signature types will use dedicated JSONs. Let's prepare classes which model various signature types, for this release only the PowerAuth signature model class type.

jandusil:
- Add new column signature_metadata, remove merged columns signature_data_method and signature_data_uri_id, document it in liquibase
- Edit entity class, introduce new converter


docs/db/changelog/changesets/powerauth-java-server/1.5.x/20230824-update-signature-metadata-columns.xml
```diff
@@ -0,0 +1,53 @@
...
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <columnExists tableName="pa_signature_audit" columnName="signature_data_method"/>
+                <columnExists tableName="pa_signature_audit" columnName="signature_data_uri_id"/>
+            </and>
+        </preConditions>
+        <comment>Migrate data from signature_data_method and signature_data_uri_id to signature_metadata</comment>
+        <sql>
+            UPDATE pa_signature_audit
+            SET signature_metadata = jsonb_build_object(
```
romanstrobl, line 32: This SQL query will only work in PostgresSQL. What about Oracle? I suggest to drop this migration.

romanstrobl (approved):
Good job. Let's review the Liquibase script, because it will not work on other databases such as Oracle.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(SignatureMetadataConverter.class);
+
+    private final ObjectMapper objectMapper;
+
+    /**
+     * Converter constructor.
+     * @param objectMapper Object mapper.
+     */
+    public SignatureMetadataConverter(ObjectMapper objectMapper) {
```
romanstrobl, line 48: I suggest to add an `@Autowired` annotation. I know it works also without the annotation, however it will be more clear how the ObjectMapper is set up.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureMetadata.java
```diff
@@ -0,0 +1,15 @@
+package io.getlime.security.powerauth.app.server.database.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class SignatureMetadata {
```
petrdvorak, line 10: I think this should be an implementation of the interface - we need to use different types eventually.

petrdvorak: The specific class might need to have also specific JSON serialization / deserialization annotation, plus i.e., a discriminator transient property.

petrdvorak (changes requested):
Please let's introduce abstraction on signature metadata types. We currently need to have only PowerAuth signature metadata with method and URI ID, but the reason we are making this change is that next release, we will also need to have FIDO2 signature metadata with completely different attributes.

As a result, we need to introduce facilities that allow serialization / deserialization of the specific classes.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
```diff
@@ -157,15 +158,15 @@ public void logSignatureAuditRecord(ActivationRecordDto activation, SignatureDat
...
         signatureAuditRecord.setActivation(activationRepository.getReferenceById(activation.getActivationId()));
         signatureAuditRecord.setActivationCounter(activation.getCounter());
         signatureAuditRecord.setActivationCtrDataBase64(activation.getCtrDataBase64());
         signatureAuditRecord.setActivationStatus(activation.getActivationStatus());
         signatureAuditRecord.setAdditionalInfo(additionalInfo);
         signatureAuditRecord.setDataBase64(data);
         signatureAuditRecord.setSignature(signatureData.getSignature());
-        signatureAuditRecord.setSignatureDataMethod(signatureData.getRequestMethod());
-        signatureAuditRecord.setSignatureDataUriId(signatureData.getRequestUriId());
+        signatureAuditRecord.setSignatureMetadata(signatureMetadata);
```
jandusil, line 169: Do we want to adjust the AuditDetail as well? Probably yes, right?

docs/db/changelog/changesets/powerauth-java-server/1.5.x/20230323-add-column-signature-metadata.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <changeSet id="1" logicalFilePath="powerauth-java-server/1.5.x/20230323-add-column-signature-metadata" author="Lubos Racansky">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="pa_signature_audit" columnName="signature_metadata"/>
+            </not>
+        </preConditions>
+        <comment>Add signature_metadata column of type JSONB</comment>
+        <addColumn tableName="pa_signature_audit">
+            <column name="signature_metadata" type="JSONB"/>
```
banterCZ, line 14: Does it work even for Oracle?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
```diff
@@ -0,0 +1,94 @@
...
+ * A JPA attribute converter for converting SignatureMetadata objects to and from JSON representations.
+ * This class enables storing SignatureMetadata in the database as a JSON column.
+ *
+ * @author Jan Dusil
+ */
+@Converter
+@Component
+public class SignatureMetadataConverter implements AttributeConverter<SignatureMetadata<?, ?>, String> {
+
+    private static final Logger logger = LoggerFactory.getLogger(SignatureMetadataConverter.class);
```
banterCZ, line 42: May be replaced by `@Slf4j`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Converts a JSON string representation to a SignatureMetadata object.
+     *
+     * @param s The JSON string to convert.
+     * @return The converted SignatureMetadata object.
+     */
+    @Override
+    public SignatureMetadata<?, ?> convertToEntityAttribute(String s) {
+        if (s == null || s.isEmpty()) {
```
banterCZ, line 83: May be replaced by `StringUtils`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/PowerAuthSignatureMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+
+package io.getlime.security.powerauth.app.server.database.model;
+
+import lombok.AllArgsConstructor;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.NoArgsConstructor;
+
+
+/**
+ * Concrete implementation of the SignatureMetadata interface for PowerAuth.
```
banterCZ, line 27: Could be linked
```suggestion
 * Concrete implementation of the {@link SignatureMetadata} interface for PowerAuth.
```

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/model/signature/SignatureMetadataConverterTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void setUp() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        converter = new SignatureMetadataConverter(objectMapper);
+    }
+
+    /**
+     * Tests the conversion of a PowerAuthSignatureMetadata object to its serialized JSON form.
+     */
+    @Test
+    public void convertToDatabaseColumnTest() {
```
banterCZ, line 55: Nitpick? Let's use package private classes and methods for JUnit 5.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/model/signature/SignatureMetadataConverterTest.java
```diff
@@ -0,0 +1,99 @@
...
+        assertEquals(originalMetadata.getMetadataParam2(), convertedMetadata.getMetadataParam2());
+    }
+
+    /**
+     * Tests the converter's behavior when provided with an invalid JSON string.
+     */
+    @Test
+    public void testInvalidJsonInput() {
+        String invalidJson = "{\"invalidField\":\"someValue\"}";
+        SignatureMetadata<?, ?> metadata = converter.convertToEntityAttribute(invalidJson);
```
banterCZ, line 96: 🤔 I know that Java generics are tricky. But don't we have anything better than unbounded wildcard type?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/model/signature/SignatureMetadataConverterTest.java
```diff
@@ -0,0 +1,99 @@
...
+import io.getlime.security.powerauth.app.server.database.model.PowerAuthSignatureMetadata;
+import io.getlime.security.powerauth.app.server.database.model.SignatureMetadata;
+import io.getlime.security.powerauth.app.server.database.model.converter.SignatureMetadataConverter;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Test class for SignatureMetadataConverter.
```
banterCZ, line 30: dtto

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/PowerAuthSignatureMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+ * Contains metadata parameters specific to PowerAuth Signature.
+ *
+ * @author Jan Dusil
+ */
+@AllArgsConstructor
+@NoArgsConstructor
+public class PowerAuthSignatureMetadata implements SignatureMetadata<String, String> {
```
banterCZ, line 34: Could you please double-check and clarify dirty-check behavior for this type? Especially when missing `equals`. We have faced some issues even for enums when upgrading Hibernate.

banterCZ: Moreover, shouldn't be the class or even the interface _serializable_? Because `SignatureEntity` is.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+})
+public interface SignatureMetadata<T1, T2> {
+
+    /**
+     * Retrieves the first metadata parameter.
+     *
+     * @return The first metadata parameter.
+     */
+    @JsonIgnore
+    T1 getMetadataParam1();
```
banterCZ, line 47: Probably already discussed with @petrdvorak, but cannot it be just a marker interface? At first glance, I just don't like the very generic name `metadataParam1`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/SignatureMetadataType.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Jan Dusil
+ */
+public class SignatureMetadataType {
+
+    /**
+     * Represents PowerAuth signature metadata.
+     * This value is used for identifying the type of metadata in JSON serialization/deserialization.
+     */
+    public static final String POWER_AUTH = "PowerAuthSignatureMetadata";
```
romanstrobl, line 37: We usually write `POWERAUTH` without underscore.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * A JPA attribute converter for converting SignatureMetadata objects to and from JSON representations.
+ * This class enables storing SignatureMetadata in the database as a JSON column.
+ *
+ * @author Jan Dusil
+ */
+@Converter
+@Component
+@Slf4j
+public class SignatureMetadataConverter implements AttributeConverter<SignatureMetadata<?, ?>, String> {
```
romanstrobl, line 41: Why do we use the `<?, ?>` type specification?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/SignatureMetadataType.java
```diff
@@ -0,0 +1,39 @@
...
+ * Enumeration constants representing the types of signature metadata supported in the application.
+ * These constants are used as names in JSON subtypes and therefore must be kept in sync.
+ * <p>
+ * Note: Although it is represented as a class with static final String fields, it serves
+ * the purpose of an enum type for specific use cases that require constant String values.
+ * </p>
+ *
+ * @author Jan Dusil
+ */
+public class SignatureMetadataType {
```
romanstrobl, line 31: If this is an enumeration, cannot we use an `enum` and map it to JSON?

romanstrobl (changes requested):
It should work as expected, however I suggest we improve the types. In my opinion there is no need to use the `?` (any type) in generics, or am I missing something? E.g. we know the resulting value in DB is a String.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/SignatureMetadataType.java
jandusil: The problem would appear here-> JsonSubTypes.Type(name="X"). I could not figure out a better way and I also found some enums in enrollment server project (see ErrorCode.class), where the "enum" was specified exactly like here.

Could I leave it like this? Or could you suggest a workaround? Thanks


@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = PowerAuthSignatureMetadata.class,
                name = SignatureMetadataType.POWER_AUTH)
})

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
jandusil: My general idea was that this converter could be used for different types of SignatureMetadata implementations not just the SignatureMetadata<String, String>. Since the props of implementations of SignatureMetadata could be different types then I thought the <?, ?> shall be here.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/SignatureMetadataType.java
romanstrobl: Alright, I think we can keep it for now.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
romanstrobl: How do you mean it?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/SignatureMetadataType.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Represents PowerAuth signature metadata.
+     * This value is used for identifying the type of metadata in JSON serialization/deserialization.
+     */
+    public static final String POWERAUTH = "PowerAuthSignatureMetadata";
```
petrdvorak, line 37: Since this will get serialized into the JSON payload, we should use just `POWERAUTH`. Also, are we sure this will get into the serialized JSON? We need to be able to deserialize it back...

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/AuditingServiceBehavior.java
petrdvorak: Yes, we need to store it in the audit record too. 👍

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureMetadata.java
```diff
@@ -0,0 +1,43 @@
...
+ * @param <T1> The type of the first metadata parameter.
+ * @param <T2> The type of the second metadata parameter.
+ * @author Jan Dusil
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = PowerAuthSignatureMetadata.class,
+                name = SignatureMetadataType.POWERAUTH)
+})
+public interface SignatureMetadata<T1, T2> extends Serializable {
```
petrdvorak, line 40: Where do we keep the `type` property? We need to have it somewhere so that we are able to deserialize the value back...

petrdvorak (changes requested):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/SignatureMetadataConverter.java
romanstrobl: In current code the usage of generics is really strange. See the following message from IntelliJ IDEA:
image

My point is that we specify types for a generified class `SignatureMetadata`, however these types are never used, so they just make the code more complicated for no good reason. If you remove the `<T1, T2>` and `<?, ?>` types everywhere from the code for the `SignatureMetadata`, the tests are still green, so why do we need this generification in the first place?

banterCZ: I guess that it's just a leftover from the previous version, where there used to be such code
```java
T1 getMetadataParam1();
```

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/SignatureMetadata.java
jandusil: The type is dynamically added by the annotation
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
@JsonSubTypes({
        @JsonSubTypes.Type(value = PowerAuthSignatureMetadata.class,
                name = SignatureMetadataType.POWERAUTH)
})

petrdvorak (approved):
Looks OK now

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/PowerAuthSignatureMetadata.java
```diff
@@ -0,0 +1,58 @@
...
+ * Concrete implementation of the {@link SignatureMetadata} interface for PowerAuth.
+ * Contains metadata parameters specific to PowerAuth Signature.
+ *
+ * @author Jan Dusil
+ */
+@AllArgsConstructor
+@NoArgsConstructor
+@Data
+public class PowerAuthSignatureMetadata implements SignatureMetadata {
+
```
banterCZ, line 39: Missing 
```java
@Serial
private static final long serialVersionUID = ...
```

romanstrobl (approved):
Much better now!

PR #987, opened 2023-08-25
Fix #986: Add logstash-logback-encoder dependency

### LINKED ISSUE #986: Add logstash-logback-encoder dependency

banterCZ:


romanstrobl (approved):
Looks OK.

Issue #762, opened 2022-09-08
Insufficient input validation in activation prepare endpoint


hvge:
By accident I found a wrong input validation in `/rest/v3/activation/prepare` endpoint if mobile client doesn't send valid JSON in encrypted data. For example, if `devicePublicKey` is missing, then the following crash is reported:
```
java.lang.NullPointerException: null
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:889) ~[guava-31.0.1-jre.jar:na]
	at com.google.common.io.BaseEncoding$StandardBaseEncoding.trimTrailingPadding(BaseEncoding.java:675) ~[guava-31.0.1-jre.jar:na]
	at com.google.common.io.BaseEncoding.decodeChecked(BaseEncoding.java:232) ~[guava-31.0.1-jre.jar:na]
	at com.google.common.io.BaseEncoding.decode(BaseEncoding.java:217) ~[guava-31.0.1-jre.jar:na]
	at io.getlime.security.powerauth.app.server.service.behavior.tasks.v3.ActivationServiceBehavior.prepareActivation(ActivationServiceBehavior.java:888) ~[classes/:1.2.5]
	at io.getlime.security.powerauth.app.server.service.behavior.tasks.v3.ActivationServiceBehavior$$FastClassBySpringCGLIB$$ed6eb169.invoke(<generated>) ~[classes/:1.2.5]
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218) ~[spring-core-5.3.18.jar:5.3.18]
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:689) ~[spring-aop-5.3.18.jar:5.3.18]
	at io.getlime.security.powerauth.app.server.service.behavior.tasks.v3.ActivationServiceBehavior$$EnhancerBySpringCGLIB$$35bd738c.prepareActivation(<generated>) ~[classes/:1.2.5]
	at io.getlime.security.powerauth.app.server.service.v3.PowerAuthServiceImpl.prepareActivation(PowerAuthServiceImpl.java:314) ~[classes/:1.2.5]
	at io.getlime.security.powerauth.app.server.service.v3.PowerAuthServiceImpl$$FastClassBySpringCGLIB$$9b67a966.invoke(<generated>) ~[classes/:1.2.5]
[...]

PR #1005, opened 2023-09-06
Fix ActivationServiceBehaviorTest

banterCZ:
Follow-up to #762

### DIFF
--- a/powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehaviorTest.java
+++ b/powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehaviorTest.java
@@ -17,8 +17,6 @@
  */
 package io.getlime.security.powerauth.app.server.service.behavior.tasks;
 
-// TODO Lubos resolve conflicts and rewrite to EncryptorFactory
-/*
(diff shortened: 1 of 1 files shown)


powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehaviorTest.java
```diff
@@ -71,7 +67,7 @@ class ActivationServiceBehaviorTest {
 
     private final KeyConvertor keyConvertor = new KeyConvertor();
     private final ObjectMapper objectMapper = new ObjectMapper();
-    private final String version = "3.2";
+    private final String version = "3.1";
```
romanstrobl, line 70: Why don't we use version `3.2`? It is the latest version of the protocol.

romanstrobl (approved):
Looks OK, just one question about protocol version.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehaviorTest.java
banterCZ: 🤔 Most probably I overcomplicated that, because it was reported earlier. Let's change it to **3.2**

PR #1006, opened 2023-09-06
Fix #980: Crypto 3.2: Adjust time window for timestamp validation

### LINKED ISSUE #980: Crypto 3.2: Adjust time window for timestamp validation
There's still 2 hour time window for the timestamp validation in ECIES and tokens in version 3.2. We should fix this and set it to 1 minute for protocol version 3.2. The value can be also altered in the server's configuration properties.

We expect that mobile device will use always the time synchronised with the server.

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
```diff
@@ -52,9 +52,15 @@ class DefaultReplayVerificationService implements ReplayVerificationService {
     private final PowerAuthServiceConfiguration config;
 
     @Override
-    public void checkAndPersistUniqueValue(UniqueValueType type, Date requestTimestamp, String ephemeralPublicKey, String nonce, String identifier) throws GenericServiceException {
+    public void checkAndPersistUniqueValue(UniqueValueType type, Date requestTimestamp, String ephemeralPublicKey, String nonce, String identifier, String version) throws GenericServiceException {
         logger.debug("Checking and persisting unique value, request type: {}, identifier: {}", type, identifier);
-        final Date expiration = Date.from(Instant.now().plus(config.getRequestExpirationInMilliseconds(), ChronoUnit.MILLIS));
+        final int requestExpiration;
+        if ("3.2".equals(version)) {
```
petrdvorak, line 58: I would prefer to revert the logic to use the new behavior for all new versions (we can easily list the prior versions). This is something we will for sure forget to edit, which will accidentally extend the validity for 3.2+ releases, and we won't even notice.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -162,12 +162,19 @@ public class PowerAuthServiceConfiguration {
...
+     * Expiration of timestamps for ECIES and MAC token requests for protocol version 3.2+.
      */
     @Value("${powerauth.service.crypto.requestExpirationInMilliseconds}")
     @Min(0)
     private int requestExpirationInMilliseconds;
 
+    /**
+     * Expiration of timestamps for ECIES and MAC token requests for protocol version 3.1 or older.
+     */
+    @Value("${powerauth.service.crypto.requestExpirationInMillisecondsBeforeV32}")
```
petrdvorak, line 174: I would generify the naming a bit, maybe using `*Extended` instead of `*BeforeV32`?

powerauth-java-server/src/main/resources/application.properties
```diff
@@ -55,7 +55,8 @@ powerauth.service.crypto.activationValidityInMilliseconds=300000
 powerauth.service.crypto.signatureMaxFailedAttempts=5
 powerauth.service.crypto.signatureValidationLookahead=20
 powerauth.service.crypto.offlineSignatureComponentLength=8
-powerauth.service.crypto.requestExpirationInMilliseconds=7200000
+powerauth.service.crypto.requestExpirationInMilliseconds=60000
+powerauth.service.crypto.requestExpirationInMillisecondsBeforeV32=7200000
```
petrdvorak, line 59: dtto naming

powerauth-java-server/src/test/resources/application.properties
```diff
@@ -42,7 +42,8 @@ powerauth.service.crypto.activationValidityInMilliseconds=120000
 powerauth.service.crypto.signatureMaxFailedAttempts=5
 powerauth.service.crypto.signatureValidationLookahead=20
 powerauth.service.crypto.offlineSignatureComponentLength=8
-powerauth.service.crypto.requestExpirationInMilliseconds=7200000
+powerauth.service.crypto.requestExpirationInMilliseconds=60000
+powerauth.service.crypto.requestExpirationInMillisecondsBeforeV32=7200000
```
petrdvorak, line 46: dtto naming

petrdvorak (changes requested):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
romanstrobl: Ok, I can revert the logic.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
romanstrobl: Sure.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
```diff
@@ -52,9 +52,15 @@ class DefaultReplayVerificationService implements ReplayVerificationService {
     private final PowerAuthServiceConfiguration config;
 
     @Override
-    public void checkAndPersistUniqueValue(UniqueValueType type, Date requestTimestamp, String ephemeralPublicKey, String nonce, String identifier) throws GenericServiceException {
+    public void checkAndPersistUniqueValue(UniqueValueType type, Date requestTimestamp, String ephemeralPublicKey, String nonce, String identifier, String version) throws GenericServiceException {
         logger.debug("Checking and persisting unique value, request type: {}, identifier: {}", type, identifier);
-        final Date expiration = Date.from(Instant.now().plus(config.getRequestExpirationInMilliseconds(), ChronoUnit.MILLIS));
+        final int requestExpiration;
+        if (!"3.2".equals(version)) {
```
petrdvorak, line 58: ... this does not fix the issue raised in my original comment, we will still need to adjust the code here in case version changes which is tricky.

petrdvorak (changes requested):
Still, that one thing is not fixed in a defensive enough manner.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
romanstrobl: So, how do you suggest to fix it?

romanstrobl: I think I see now what you mean.

petrdvorak (approved):
Looks OK

PR #1015, opened 2023-09-07
Fix #1014: Add protocol version into token verification endpoint, update documentation

### LINKED ISSUE #1014: Add protocol version into token verification endpoint
Add the protocol version to token verification endpoint as a parameter and use the version for calculation of token digest.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1024, opened 2023-09-07
Fix #1023: Make TelemetryReportException extend Exception instead of Throwable

### LINKED ISSUE #1023: Make TelemetryReportException extend Exception instead of Throwable

petrdvorak:


banterCZ (approved):
Yes, please!

PR #1026, opened 2023-09-07
Fix #1025: Missing private modifier for offset

### LINKED ISSUE #1025: Missing private modifier for offset
In class `SdkDataReader` the field `offset` should be private.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1032, opened 2023-09-08
Fix #1022: Signature metadata is not generic enough

### LINKED ISSUE #1022: Signature metadata is not generic enough
In class `SignatureEntity` the signature data is hardcoded to `PowerAuthSignatureMetadata`:

```
@Column(name = "signature_metadata")
@Convert(converter = SignatureMetadataConverter.class)
private PowerAuthSignatureMetadata signatureMetadata;
```

See:
https://github.com/wultra/powerauth-server/blob/625808d7d1f9a3cbbc32cf93a39985c937429689/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/SignatureEntity.java#L80

The goal of introducing the `signature_metadata` column was to make the signature metadata generic, which is not achieved by fixing the signature metadata type to `PowerAuthSignatureMetadata`. We should try to make the functionality more generic, so that multiple types of signature data can be written into the database.

jandusil:
- Add generic type to SignatureEntity

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/SignatureEntity.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/SignatureEntity.java
@@ -17,7 +17,7 @@
  */
 package io.getlime.security.powerauth.app.server.database.model.entity;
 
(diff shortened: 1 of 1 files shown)


banterCZ (approved):
Looks good.

PR #1038, opened 2023-09-11
Fix #1034: Prefer Duration over int and long

### LINKED ISSUE #1034: Prefer Duration over int and long
In Java code, replace `int` and `long` by `Duration`, e.g. in `PowerAuthServiceConfiguration#activationsCleanupLookBackInMilliseconds`
Keep current properties for backward compatibility. Spring Boot binds value without unit to milliseconds by default.

banterCZ:


powerauth-java-server/src/main/resources/application.properties
```diff
@@ -55,8 +55,8 @@ powerauth.service.crypto.activationValidityInMilliseconds=300000
 powerauth.service.crypto.signatureMaxFailedAttempts=5
 powerauth.service.crypto.signatureValidationLookahead=20
 powerauth.service.crypto.offlineSignatureComponentLength=8
-powerauth.service.crypto.requestExpirationInMilliseconds=60000
-powerauth.service.crypto.requestExpirationInMillisecondsExtended=7200000
+powerauth.service.crypto.requestExpiration=60s
```
banterCZ, line 58: This is a safe change because the property has been newly introduced.

banterCZ:
I have reverted changes in the property files and just kept the Java type improvements.

romanstrobl (approved):
No problem now.

PR #1047, opened 2023-09-11
Fix #1046: Update documentation for error codes

### LINKED ISSUE #1046: Update documentation for error codes
The list of error codes needs to be updated.

romanstrobl:


banterCZ (approved):
I have also added an i18n item.

PR #1057, opened 2023-09-18
Fix #1041: Add information about token verification change to migration guide

### LINKED ISSUE #1041: Add information about token verification change to migration guide
Document change of REST API and PowerAuth client due to versioning of the token verification endpoint.

romanstrobl:


banterCZ (approved):
Looks good.

PR #1066, opened 2023-10-04
Fix #1065: Filter jboss descriptor with bouncycastle version from pom

### LINKED ISSUE #1065: Filter jboss descriptor with bouncycastle version from pom

banterCZ:


docs/Installing-Bouncy-Castle.md
```diff
@@ -1,169 +1,3 @@
 # Installing Bouncy Castle
 
 Since PowerAuth Server in version 1.1.x, installing the Bouncy Castle in OS or Java container is no longer required. The latest version is bundled with the app and cryptographic primitives should work out of the box.
-
```
banterCZ, line 4: I believe that we may remove it. It is available in the documentation for older versions.

romanstrobl (approved):
Good idea.

PR #1072, opened 2023-10-16
Fix #1071: Exclude lombok from war

### LINKED ISSUE #1071: Exclude lombok from war
`spring-boot-maven-plugin` does not respect `provided` scope on purpose, which is strange, but let's get rid of lombok and servlet-api from the final artifact. https://docs.spring.io/spring-boot/docs/2.3.x/maven-plugin/reference/html/#repackage-example-exclude-dependency

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1079, opened 2023-10-17
Fix #1078: Overload GetActivationListForUserRequest constructor

### LINKED ISSUE #1078: Overload GetActivationListForUserRequest constructor

jandusil:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/GetActivationListForUserRequest.java
```diff
@@ -19,14 +19,18 @@
...
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 /**
  * Model class representing request for activation list for a given user.
  *
  * @author Petr Dvorak, petr@wultra.com
  */
 @Data
+@AllArgsConstructor
```
banterCZ, line 32: What about rather introducing a builder?

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/GetActivationListForUserRequest.java
```diff
@@ -19,18 +19,16 @@
...
 
 /**
  * Model class representing request for activation list for a given user.
  *
  * @author Petr Dvorak, petr@wultra.com
  */
 @Data
-@AllArgsConstructor
-@NoArgsConstructor
+@Builder
```
banterCZ, line 31: One more thing, this annotation creates an all-arg package private constructor and you are not able to call `new GetActivationListForUserRequest()` anymore. Is it on purpose?

jandusil:
Not needed

PR #1103, opened 2023-10-31
Fix #1100: Application check in update callback api

### LINKED ISSUE #1100: Application check in update callback api
`/application/callback/update` expects `callbackId` of the callback url to edit and `applicationId` of the application to which the callback is assigned. There should be a check during the update to validate the passed `applicationId` corresponds to the right application owning the callback.

If the `callbackId` belongs to a different application, throw an exception.

jnpsk:
Fix #1100:
If `applicationId` from the request does not match the `applicationId` paired with the callback, behave like the callback does not exist at all.

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
@@ -154,8 +154,8 @@ public class CallbackUrlBehavior {
         }
 
         final Optional<CallbackUrlEntity> entityOptional = callbackUrlRepository.findById(request.getId());
-        if (entityOptional.isEmpty()) {
-            logger.warn("Invalid callback ID: "+request.getId());
+        if (entityOptional.isEmpty() || !request.getApplicationId().equals(entityOptional.get().getApplication().getId())) {
+            logger.warn("Invalid callback ID: {}", request.getId());
             // Rollback is not required, error occurs before writing to database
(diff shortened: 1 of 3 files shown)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -154,8 +154,8 @@ public UpdateCallbackUrlResponse updateCallbackUrl(UpdateCallbackUrlRequest requ
         }
 
         final Optional<CallbackUrlEntity> entityOptional = callbackUrlRepository.findById(request.getId());
-        if (entityOptional.isEmpty()) {
-            logger.warn("Invalid callback ID: "+request.getId());
+        if (entityOptional.isEmpty() || !request.getApplicationId().equals(entityOptional.get().getApplication().getId())) {
+            logger.warn("Invalid callback ID: {}", request.getId());
             // Rollback is not required, error occurs before writing to database
             throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
         }
```
banterCZ, line 161: First of all, I would like to see a unit test for that.
FYI, it may be rewritten to Java 8 style, for example
```suggestion
        final CallbackUrlEntity entity = entityOptional
                .filter(it -> it.getApplication().getId().equals(request.getApplicationId()))
                .orElseThrow(() -> {
                    // Rollback is not required, error occurs before writing to database
                    logger.warn("Invalid callback ID: {}", request.getId());
                    return localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
                });
```

PR #1105, opened 2023-11-01
Fix #1104: Increase default pagination value

### LINKED ISSUE #1104: Increase default pagination value

jandusil:


jandusil:
After consulting the topic, it is not needed

PR #1109, opened 2023-11-03
Fix #1107: Update activation name

### LINKED ISSUE #1107: Update activation name
Expose a new API for setting the activation name for activations that are not in the state CREATED, REMOVED, or BLOCKED.
The change should be recorded in the history and the audit.

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
```diff
@@ -216,54 +192,32 @@ public void setActivationVersion(Integer version) {
...
-        int hash = 7;
-        hash = 23 * hash + Objects.hashCode(this.activation);
-        hash = 23 * hash + Objects.hashCode(this.activationStatus);
-        hash = 23 * hash + Objects.hashCode(this.eventReason);
-        hash = 23 * hash + Objects.hashCode(this.externalUserId);
-        hash = 23 * hash + Objects.hashCode(this.timestampCreated);
-        hash = 23 * hash + Objects.hashCode(this.activationVersion);
-        return hash;
+        // hashCode should always return the same value even when autogenerated ID has been set; performance penalty is negligible
+        return getClass().hashCode();
```
banterCZ, line 206: Let's not add the new field `applicationName` and return constant as suggested https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/
If agreed, I will update our guidelines.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -0,0 +1,126 @@
...
+        assertEquals(1, historyEntries.size());
+
+        final ActivationHistoryEntity historyEntry = historyEntries.iterator().next();
+        assertEquals("my iPhone", historyEntry.getActivationName());
+        assertEquals("ACTIVATION_UPDATED", historyEntry.getEventReason());
+        assertEquals("joe-1", historyEntry.getExternalUserId());
+
+        databaseAudit.flush();
+        @SuppressWarnings("unchecked")
+        final List<Tuple> auditEntries = entityManager.createNativeQuery("select * from audit_log", Tuple.class).getResultList();
```
banterCZ, line 96: - [ ] Improve select to get one record only.

romanstrobl (approved):
Looks OK, I have two minor suggestions.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/AdditionalInformation.java
```diff
@@ -80,6 +80,11 @@ public static class Reason {
          */
         public static final String ACTIVATION_VERSION_CHANGED = "ACTIVATION_VERSION_CHANGED";
 
+        /**
+         * Logged when the activation has been updated, e.g. updated activation name.
+         */
+        public static final String ACTIVATION_UPDATED = "ACTIVATION_UPDATED";
```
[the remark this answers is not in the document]
banterCZ: It was generic on purpose. I was open to a later extension, but I agree, let's make it more specific.

zcgandcomp (changes requested):
Overall it is fine. I have only one remark - feel free to discuss. 

Question:
Can we name the methods and objects `updateActivationName`? And the as url use  `/activation/name/update`.

Reason:
We are not updating activation we have a specialized method for updating activation name only. So this is an analogy to the naming of method `updateActivationOtp`

PR #1110, opened 2023-11-06
Set Java version for GHA to 21

banterCZ:


banterCZ:
Was just a test of GHA.

PR #1112, opened 2023-11-06
Fix #1108: Update database structure diagram

### LINKED ISSUE #1108: Update database structure diagram
Regenerate database structure diagram, this one is outdated https://github.com/wultra/powerauth-server/blob/releases/1.5.x/docs/images/arch_db_structure.png

Moreover, the cardinality of application <-> operation has changed but table `pa_operation_application` is missing foreign keys, which should be added.

Backport at least the diagram into 1.5.x (FK may be part only of 1.6.0)

jnpsk:


banterCZ (approved):
Thanks for the backport.

PR #1118, opened 2023-11-08
Fix #1107: Update activation name (backport)

### LINKED ISSUE #1107: Update activation name

banterCZ:
* Fix #1107: Update activation name

- add activation name into history

(cherry picked from commit a10c77f19d78c1e1b21aff4382eac3b11e7eb1e3)


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
```diff
@@ -216,54 +193,31 @@ public void setActivationVersion(Integer version) {
...
-            return false;
-        }
-        if (!Objects.equals(this.timestampCreated, other.timestampCreated)) {
+        } else if (this == obj) {
+            return true;
+        } else if (!this.getClass().equals(ProxyUtils.getUserClass(obj))) {
             return false;
+        } else {
+            final ActivationHistoryEntity that = (ActivationHistoryEntity) obj;
+            return null != this.getId() && this.getId().equals(that.getId());
```
romanstrobl, line 219: This change is incorrect, because an unpersisted and persisted entity from the same instance will not be equal. This is caused by autegenerated ID, which is initially null and then becomes non-null. This can bring unexpected and hard to discover issues when using collections.

romanstrobl (changes requested):
We should not use ID comparison in equals for entities with auto-generated IDs.

romanstrobl (approved):
Looks good now, thanks!

PR #1119, opened 2023-11-08
Fix equals and hashCode of ActivationHistoryEntity

banterCZ:
A follow-up to #1107

### DIFF
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
@@ -212,13 +212,13 @@ public class ActivationHistoryEntity implements Serializable {
             return false;
         } else {
             final ActivationHistoryEntity that = (ActivationHistoryEntity) o;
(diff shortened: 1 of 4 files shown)


romanstrobl (changes requested):
Better, but still incorrect. We should include fields which identify the entity into the equals and hashCode contract, not various parameters such as reason why activation was blocked, that does not correspond to the identity of the entity.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationHistoryEntity.java
```diff
@@ -202,22 +202,27 @@ public void setActivationName(final String activationName) {
...
         }
     }
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(nullSaveActivationId(activation), timestampCreated);
+    }
+
+    // TODO (racansky, 2023-11-08) remove when activation equals and hashCode implemented correctly
+    private static String nullSaveActivationId(final ActivationRecordEntity activation) {
```
romanstrobl, line 224: Nullsafe?

romanstrobl (changes requested):
Incorrect method name.

romanstrobl (approved):
Excellent, thanks for your patience.

PR #1123, opened 2023-11-09
Fix #1122: Add error handling of db changes to migration guide

### LINKED ISSUE #1122: Add error handling of db changes to migration guide
Migration guide `PowerAuth-Server-1.6.0.md` does not contain recommended steps to take in case of failure.

jnpsk:


jnpsk:
> Can you add SQL which will list inconsitences in DB? (even how to fix it)

@zcgandcomp Do we want to add sql statements to list or remove problematic rows? None of the previous guides contains it.

zcgandcomp:
> > Can you add SQL which will list inconsitences in DB? (even how to fix it)
> 
> @zcgandcomp Do we want to add sql statements to list or remove problematic rows? None of the previous guides contains it.

As we can run into the issues with different databases (probably not, but in theory) I would just state that if foreign keys cannot be automatically created relevant lines in the pa_operation_application should be removed (after backup).

zcgandcomp (approved):
I think the description looks fine and clear for potential troubleshooting.

PR #1125, opened 2023-11-12
Fix #1124: Adjust properties of Operation entity/dtos for additional …

### LINKED ISSUE #1124: Adjust properties of Operation entity/dtos for additional auditing

jandusil:
…auditing

### DIFF
--- a/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
+++ b/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
@@ -33,6 +33,6 @@ import java.util.Map;
 public class OperationFailApprovalRequest {
 
     private String operationId;
-    private final Map<String, Serializable> additionalData = new LinkedHashMap<>();
+    private Map<String, Object> additionalData = new LinkedHashMap<>();
 
 }
--- a/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
+++ b/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
@@ -20,7 +20,6 @@ package com.wultra.security.powerauth.client.model.request;
 
 import lombok.Data;
 
-import java.io.Serializable;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
--- a/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
+++ b/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
@@ -32,6 +32,6 @@ import java.util.Map;
 public class OperationFailApprovalRequest {
 
     private String operationId;
-    private Map<String, Object> additionalData = new LinkedHashMap<>();
+    private final Map<String, Object> additionalData = new LinkedHashMap<>();
 
 }
--- a/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
+++ b/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
@@ -20,6 +20,7 @@ package com.wultra.security.powerauth.client.model.request;
 
 import lombok.Data;
 
+import java.io.Serializable;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -32,6 +33,6 @@ import java.util.Map;
 public class OperationFailApprovalRequest {
 
     private String operationId;
-    private final Map<String, String> additionalData = new LinkedHashMap<>();
+    private final Map<String, Serializable> additionalData = new LinkedHashMap<>();
 
 }
--- a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverter.java
+++ b/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverter.java
@@ -27,7 +27,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
-import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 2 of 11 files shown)


jandusil:
Connected with https://github.com/wultra/enrollment-server/issues/904

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -797,7 +798,7 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
             return ProximityCheckResult.DISABLED;
         }
 
-        final String otp = request.getAdditionalData().get(PROXIMITY_OTP);
+        final String otp = (String) request.getAdditionalData().get(PROXIMITY_OTP);
```
banterCZ, line 801: How is `ClassCastException` avoided?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -730,8 +731,8 @@ private boolean isDataEqual(@NotNull OperationEntity operation, String providedD
     }
 
     // Merge two maps into new one, replacing values in the first map when collision occurs
-    private Map<String, String> mapMerge(Map<String, String> m1, Map<String, String> m2) {
-        final Map<String, String> m3 = new HashMap<>();
+    private Map<String, Serializable> mapMerge(Map<String, Serializable> m1, Map<String, Serializable> m2) {
+        final Map<String, Serializable> m3 = new HashMap<>();
```
banterCZ, line 735: Do we have a test for that? What will it look like? I mean, will the user-agent be a string or a structured map?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverterTest.java
```diff
@@ -0,0 +1,144 @@
...
+        Map<String, Serializable> testMap = new HashMap<>();
+        testMap.put("key1", "value1");
+        final String exampleRequestUserAgent = "PowerAuthNetworking/1.1.7 (en; cellular) com.wultra.app.MobileToken.wtest/2.0.0 (Apple; iOS/16.6.1; iphone12,3)";
+        final UserAgent.Device device = UserAgent.parse(exampleRequestUserAgent);
+        testMap.put("key2", device);
+
+        final String jsonResult = converter.convertToDatabaseColumn(testMap);
+        assertNotNull(jsonResult);
+        assertFalse(jsonResult.isEmpty());
+        assertEquals("{\"key1\":\"value1\"," +
```
banterCZ, line 64: You may use multiline string if you want to.

banterCZ (approved):
Looks good.

romanstrobl (changes requested):
This pull request looks strange to me.

We either want to have:
- Map<String, String>
- Map<String, Object>

I do not understand Map<String, Serializable> so I will reject this request until it is explained, maybe it is valid, but I just do not understand what is going on.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/OperationEntity.java
```diff
@@ -77,7 +77,7 @@ public class OperationEntity implements Serializable {
 
     @Column(name = "additional_data")
     @Convert(converter = MapToJsonConverter.class)
-    private Map<String, String> additionalData = new HashMap<>();
+        private Map<String, Serializable> additionalData = new HashMap<>();
```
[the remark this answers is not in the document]
petrdvorak: WTF?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverterTest.java
jandusil: I don't think I can, then it adds newline chars

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationFailApprovalRequest.java
```diff
@@ -20,6 +20,7 @@
 
 import lombok.Data;
 
+import java.io.Serializable;
```
banterCZ, line 23: Really needed?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverter.java
```diff
@@ -27,6 +27,7 @@
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import java.io.Serializable;
```
banterCZ, line 30: dtto

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/database/model/converter/MapToJsonConverterTest.java
banterCZ: 🤔 Yes, you are right, that in that case you have to use `.replace("\n", "")`, which is sad. My idea was to avoid escaping. But it is just a minor issue to consider.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -813,6 +821,16 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
         }
     }
 
+    private static void parseDeviceFromUserAgent(Map<String, Object> additionalData) {
```
banterCZ, line 824: I have objections to the method name, even if it is _private_. `parse` evokes a return value to me, which is missing. In general, I do not like side effects, mainly hidden. This method is _parse and add_.

banterCZ (commented):
👍 The main PR comment is fixed, and now `Map<String, Object>` is used.

romanstrobl (dismissed):
Approved now.

banterCZ (approved):
Looks good.

PR #1133, opened 2023-11-16
Fix #1132: Add paging to operation list

### LINKED ISSUE #1132: Add paging to operation list
The operation list endpoint should support paging. For big response, the client fails because of `org.springframework.core.io.buffer.DataBufferLimitException: Exceeded limit on max bytes to buffer : 1048576`

API should be implemented similarly to the application list.

Has to be backported to 1.5.x

banterCZ:


romanstrobl (approved):
Great!

PR #1139, opened 2023-11-16
Fix #1132: Add paging to operation list (backport)

### LINKED ISSUE #1132: Add paging to operation list

banterCZ:
(cherry picked from commit 36e2fb9a09b54b0c3a21b77fff4928bba9d9262b)


romanstrobl (approved):
Looks OK.

PR #1153, opened 2023-11-23
Fix #1152: FIDO2: Invalid timestamp check for operation claim

### LINKED ISSUE #1152: FIDO2: Invalid timestamp check for operation claim
The operation claim code does not work because the timestamp check logic is reverted.

romanstrobl:


petrdvorak (approved):
Looks OK and good catch... Could we maybe add a comment here and to `expireOperation` method?

- "Only pending, non-expired operations should be claimable."
- "Pending operation with expired timestamp should be expired."

romanstrobl:
I added the comments after minor rewording.

PR #1190, opened 2023-12-12
Fix #1184: Memory leak in CallbackUrlBehavior for RestClient instances (backport)

### LINKED ISSUE #1184: Memory leak in CallbackUrlBehavior for RestClient instances
The `restClientCache` uses an unstable hashmap key which causes memory leaks. Use the simplest key possible for this cache to avoid leaking `RestClient` instances.

banterCZ:
(cherry picked from commit 33de1e0eebf42913ebc76cbbb2fbc2c681802c05)


romanstrobl (approved):
Perfect.

PR #1200, opened 2023-12-12
Fix #1199: Synchronize get access to REST client cache

### LINKED ISSUE #1199: Synchronize get access to REST client cache
Let's improve the REST client cache map access synchorization for better concurrency handling.

romanstrobl:


banterCZ (approved):
Looks good.

PR #1205, opened 2023-12-12
Fix #1204: Version 1.4.2 does not run in JDK 11

### LINKED ISSUE #1204: Version 1.4.2 does not run in JDK 11
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project powerauth-java-server: Compilation failure
[ERROR] /powerauth-server/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java:[663,84] cannot find symbol
[ERROR]   symbol:   method formatted(int,int)
[ERROR]   location: class java.lang.String

```

banterCZ:


zcgandcomp (approved):
Looks OK

PR #1215, opened 2023-12-13
Fix #1213: Specify maven.compiler.release property

### LINKED ISSUE #1213: Specify maven.compiler.release property
When built by JDK 11, it failed to run JRE 11, see #1204 
By default, the -source and -target options don’t guarantee a cross-compilation.

banterCZ:


romanstrobl (approved):
Good idea.

PR #1233, opened 2023-12-27
Fix #1113: Throwable and Error should not be caught

### LINKED ISSUE #1113: Throwable and Error should not be caught
`Throwable` is the superclass of all errors and exceptions in Java. `Error` is the superclass of all errors, which are not meant to be caught by applications.

Catching either `Throwable` or `Error` will also catch `OutOfMemoryError` and `InternalError`, from which an application should not attempt to recover.

https://rules.sonarsource.com/java/RSPEC-1181/

jandusil:


banterCZ (changes requested):
Looks good but I suggest to merge it after release 1.6.0

banterCZ (approved):

PR #1235, opened 2024-01-03
Fix #1099: Move tests from powerauth-tests to this repo

### LINKED ISSUE #1099: Move tests from powerauth-tests to this repo
Move tests from `powerauth-tests` to this repository. Look at the package `com.wultra.security.powerauth.test.v3x` and search test which do not need cryptography running.

jandusil:
- Move operation tests
- Add javadoc
- Edit sql for tests


powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -125,4 +459,740 @@ void testUpdateActivation_badRequest() throws Exception {
...
+     */
+    @Test
+    void testCallbackUpdate() throws Exception {
+        final CreateCallbackUrlRequest callbackUrlRequest = createCallbackUrlRequest();
+        final CreateCallbackUrlResponse callbackUrlResponse = callbackUrlBehavior.createCallbackUrl(callbackUrlRequest);
+        final GetCallbackUrlListResponse callbacks = callbackUrlBehavior.
+                getCallbackUrlList(createCallbackUrlListRequest());
+
+        /* Verify first created callback */
+        boolean callbackFound = false;
```
banterCZ, line 632: May we refactor into using `stream.filter(...)findAny()`?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -125,4 +459,740 @@ void testUpdateActivation_badRequest() throws Exception {
...
+     * Creates a request object for creating an operation.
+     * <p>
+     * This helper method constructs and returns an {@link OperationCreateRequest} with
+     * predefined application ID, template name, user ID, and proximity OTP settings.
+     *
+     * @param proximityOtp a boolean indicating whether proximity OTP is enabled
+     * @return a configured {@link OperationCreateRequest} instance
+     * @throws Exception if the operation creation request setup fails
+     */
+    private OperationCreateRequest createOperationCreateRequest(final boolean proximityOtp) throws Exception {
```
banterCZ, line 1121: Can by `static`? Moreover, let's rename the parameter, `proximityOtp` looks like OTP value to me.
```suggestion
    private static OperationCreateRequest createOperationCreateRequest(final boolean proximityOtpEnabled) throws Exception {
```

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -125,4 +459,740 @@ void testUpdateActivation_badRequest() throws Exception {
...
+     * Wraps an object in a 'requestObject' JSON structure.
+     * <p>
+     * This helper method wraps a given object within a 'requestObject' JSON structure, useful
+     * for preparing request bodies for MockMvc calls.
+     *
+     * @param obj the object to be wrapped in the request JSON
+     * @return a JSON string representation of the object wrapped in 'requestObject'
+     * @throws Exception if JSON processing fails
+     */
+    private static String wrapInRequestObjectJson(final Object obj) throws Exception {
```
banterCZ, line 1173: Personally, I do not like this method at all for several reasons. Mainly, we bypass testing of object mapper. For example, when the property name changes, the test should fail (in my opinion), but in this case it does not. May we use just string payload instead?

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -125,4 +459,740 @@ void testUpdateActivation_badRequest() throws Exception {
...
+
+        MvcResult result = mockMvc.perform(post("/rest/v3/integration/create")
+                        .content(wrapInRequestObjectJson(createIntegrationRequest))
+                        .contentType(MediaType.APPLICATION_JSON)
+                        .accept(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("$.responseObject.name").value(integrationName))
+                .andReturn();
+
+        final String integrationId = convertMvcResultToObject(result,
```
banterCZ, line 888: 🤔 Isn't it complicated to convert json to POJO? What about using something like `JsonPath.read(result.getResponse().getContentAsString(), "$.id")`

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehaviorTest.java
```diff
@@ -43,6 +43,8 @@
 import java.util.*;
 import java.util.concurrent.TimeUnit;
 
+import static com.wultra.security.powerauth.client.model.enumeration.UserActionResult.APPROVAL_FAILED;
```
banterCZ, line 46: I prefer to isolate test constants from production constants. When production code changes, this test does not fail.

banterCZ (commented):
Let's discuss whether we want to use request/response POJO instead of JSON in tests.

romanstrobl (changes requested):
I found two issues:
- The tests are using mocks for REST API calls using `mockMvc.perform(post("..."));`.
  - This changes the functionality of tests which used to test real REST API controllers but now have been downgraded to mocking.
  - We may not catch compatibility issues, object mapping issues, security configuration problems and other issues caused by mocking instead of real REST API tests.
  - I do not see any benefit to using mocks, we can simply test the real application.
- The tests use the same application `PA_Tests` as end-to-end tests
  - I suggest to avoid using same application `PA_Tests`, because the purpose of these tests is different (component vs. end-to-end tests).
  - Why are we setting up the keypairs using SQL inserts instead of creating everything using the REST API? We miss testing this functionality. When always using same static keys we do not test the underlying cryptography properly.

To summarize, we tried to write unit tests, while these tests were originally component tests, so we downgraded the coverage quite significantly. I suggest we stick to the original idea of preparing component tests which should replace end-to-end tests which are do not require end-to-end scenarios.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTestConfig.java
```diff
@@ -0,0 +1,493 @@
...
+    @Bean
+    public PowerAuthClient powerAuthClient() throws Exception {
+        final PowerAuthRestClientConfiguration config = new PowerAuthRestClientConfiguration();
+        config.setAcceptInvalidSslCertificate(true);
+        return new PowerAuthRestClient(POWERAUTH_REST_URL);
+    }
+
+
+    /*
+     *****************************************************************
```
romanstrobl, line 196: Can we avoid this style of comments with many `*` characters? It's not common in Java.

powerauth-java-server/pom.xml
```diff
@@ -178,6 +178,21 @@
             <version>${webauthn4j.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- For run at Apple M1 architecture -->
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-resolver-dns-native-macos</artifactId>
+            <scope>test</scope>
```
banterCZ, line 186: I believe that it could be even `runtime` scope.

romanstrobl (approved):
These changes are more inline with the original intention, thanks.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -17,112 +17,1047 @@
...
+        final LookupActivationsRequest lookupActivationsRequest = new LookupActivationsRequest();
+        /* We are looking for an activation created during initialization of the test suite. */
+        final Date timestampCreated = Date.from(LocalDateTime.now().minusSeconds(1).atZone(ZoneId.systemDefault()).toInstant());
+        lookupActivationsRequest.setUserIds(List.of(PowerAuthControllerTestConfig.USER_ID));
+        lookupActivationsRequest.setApplicationIds(List.of(config.getApplicationId()));
+        lookupActivationsRequest.setActivationStatus(ActivationStatus.CREATED);
+        lookupActivationsRequest.setTimestampLastUsedAfter(timestampCreated);
+
+        final LookupActivationsResponse lookupActivationsResponse = powerAuthClient.lookupActivations(lookupActivationsRequest);
+        assertEquals(1, lookupActivationsResponse.getActivations().size());
```
banterCZ, line 197: FYI: We do not use Hamcrest yet or not heavily, but there are more elegant ways, e.g. `assertThat(collection, hasSize(1));` or some similar syntax with AssertJ (introducing AssertJ is a bigger step, but Hamcrest should be ok, if you want to).

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -17,112 +17,1047 @@
...
+        baseRequest.setUserId(PowerAuthControllerTestConfig.USER_ID);
+        baseRequest.setApplicationId(config.getApplicationId());
+        final List<String> activationIds = new ArrayList<>();
+
+        for (int i = 0; i < 10; i++) {
+            final InitActivationRequest initActivationRequest = new InitActivationRequest();
+            initActivationRequest.setApplicationId(config.getApplicationId());
+            initActivationRequest.setUserId(PowerAuthControllerTestConfig.USER_ID);
+            final InitActivationResponse initResponse = powerAuthClient.initActivation(initActivationRequest);
+            activationIds.add(initResponse.getActivationId());
```
banterCZ, line 138: It could be nice to use `stream` -> `map`, but since we declare checked exceptions, it is not possible (or not easily at least) 😞

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTest.java
```diff
@@ -17,112 +17,1047 @@
...
+        final String expectedErrorMessage = "requestObject.activationId - must not be blank," +
+                " requestObject.activationName - must not be blank, requestObject.externalUserId - must not be blank";
+        final String expectedErrorCode = "ERR0024";
+        final PowerAuthClientException thrownException = assertThrows(
+                PowerAuthClientException.class,
+                () -> powerAuthClient.updateActivationName(new UpdateActivationNameRequest())
+        );
+        assertEquals(expectedErrorMessage, thrownException.getMessage());
+        assertEquals(expectedErrorMessage, thrownException.getLocalizedMessage());
+        assertEquals(expectedErrorMessage, thrownException.getLocalizedMessage());
```
banterCZ, line 381: Duplicated line.
```suggestion
        assertEquals(expectedErrorMessage, thrownException.getLocalizedMessage());
```

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTestConfig.java
```diff
@@ -0,0 +1,493 @@
...
+import java.security.PublicKey;
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Configuration class for PowerAuth Controller tests.
+ * <p>
+ * This class provides configuration settings and helper methods
```
banterCZ, line 53: 🤔 I do not like the mixture of concept configuration and helper methods in a single class.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTestConfig.java
```diff
@@ -0,0 +1,493 @@
...
+     * This method checks for the existence of an application and its version. If not present,
+     * it creates them and sets relevant fields in the test configuration. It also ensures the
+     * application version is supported and sets up activation recovery settings.
+     *
+     * @param powerAuthClient the PowerAuthClient to perform application-related operations
+     * @throws Exception if any error occurs during application creation or setup
+     */
+    protected void createApplication(final PowerAuthClient powerAuthClient) throws Exception {
+        final GetApplicationListResponse applicationsListResponse = powerAuthClient.getApplicationList();
+        var applicationOptional = applicationsListResponse.getApplications().stream()
```
banterCZ, line 285: May be `final`, more occurences in this class.

powerauth-java-server/src/test/java/io/getlime/security/powerauth/app/server/controller/api/PowerAuthControllerTestConfig.java
```diff
@@ -0,0 +1,493 @@
...
+
+    public void setLoginOperationTemplateName(final String loginOperationTemplateName) {
+        this.loginOperationTemplateName = loginOperationTemplateName;
+    }
+
+    public String getActivationId() {
+        return activationId;
+    }
+
+    public void setActivationId(final String activationId) {
```
banterCZ, line 157: Any reason why not to use Lombok?

banterCZ (commented):
Thanks for transforming the test from a mock into an integration one.

banterCZ (approved):
Thanks for cleaning up the configuration class.

PR #1254, opened 2024-01-17
Fix #1253: FIDO2: Improve error handling

### LINKED ISSUE #1253: FIDO2: Improve error handling
There are several cases when error handling can be improved for following code constructs:
- `catch (IOException e) { throw new RuntimeException(e);}`
- `} catch (JsonProcessingException e) { // }`
- `// Unknown version is converted to 0 in service`

romanstrobl:


powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/RegistrationConverter.java
```diff
@@ -64,7 +64,8 @@ public AuthenticatorDetail convert(RegistrationChallenge challenge, Registration
             authenticatorDetail.setMaxFailedAttempts(5L);
             return authenticatorDetail;
         } catch (JsonProcessingException e) {
-            throw new RuntimeException(e);
+            logger.warn(e.getMessage(), e);
```
banterCZ, line 67: General `RuntimeException` is not good, but either returning `null`, I would suggest to change it into `Optional`.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/AttestationObjectDeserializer.java
```diff
@@ -52,15 +52,16 @@ public AttestationObjectDeserializer(Class<?> vc) {
...
+    public AttestationObject deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
         try {
             final String originalTextValue = jsonParser.getText();
             final byte[] decodedAttestationObject = Base64.getDecoder().decode(originalTextValue);
             final AttestationObject attestationObject = cborMapper.readValue(decodedAttestationObject, AttestationObject.class);
             attestationObject.setEncoded(originalTextValue);
             return attestationObject;
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            logger.warn(e.getMessage(), e);
```
banterCZ, line 63: 👀 The declaration is `public abstract T deserialize(JsonParser p, DeserializationContext ctxt)
        throws IOException, JacksonException;` so why to catch the exception at all?

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/CollectedClientDataDeserializer.java
```diff
@@ -60,7 +60,8 @@ public CollectedClientData deserialize(JsonParser jsonParser, DeserializationCon
             collectedClientData.setEncoded(new String(decodedClientDataJSON));
             return collectedClientData;
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            logger.warn(e.getMessage(), e);
```
banterCZ, line 63: dtto

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthAuthenticatorProvider.java
```diff
@@ -265,7 +265,8 @@ private AuthenticatorDetail convert(Activation activation, ApplicationEntity app
         try {
             authenticatorDetail.setExtras(objectMapper.readValue(activation.getExtras(), new TypeReference<HashMap<String,Object>>() {}));
         } catch (JsonProcessingException e) {
-            //
+            logger.warn(e.getMessage(), e);
```
banterCZ, line 268: Indeed, improved, better than swallowing the exception but may we use `Optional`?

banterCZ (commented):
Let's think whether we could improve it a little bit.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/RegistrationConverter.java
romanstrobl: Sure, we can use `Optional`.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/AttestationObjectDeserializer.java
romanstrobl: We want to avoid a 500 error, but I will review this.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthAuthenticatorProvider.java
romanstrobl: Sure, we can use `Optional`.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/service/AssertionService.java
```diff
@@ -88,7 +90,11 @@ public AssertionVerificationResponse authenticate(AssertionVerificationRequest r
             final String applicationId = request.getApplicationId();
             final String authenticatorId = request.getId();
             final String challenge = request.getResponse().getClientDataJSON().getChallenge();
-            final AuthenticatorDetail authenticatorDetail = authenticatorProvider.findByCredentialId(applicationId, authenticatorId);
+            final Optional<AuthenticatorDetail> authenticatorOptional = authenticatorProvider.findByCredentialId(applicationId, authenticatorId);
+            if (authenticatorOptional.isEmpty()) {
```
banterCZ, line 94: You may use `#orElseThrow()`.

banterCZ (approved):

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/service/AssertionService.java
romanstrobl: Good idea.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/AttestationObjectDeserializer.java
romanstrobl: I will leave the implementation as is for now, because I do not think a generic error handler for `IOException` is a good idea and we want to avoid unhandled 500 errors. We can improve it later in case of any issues, however this is a corner case which should not occur in actual implementation because the attestation object is provided by the authenticator and shouldbe serialized correctly.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/CollectedClientDataDeserializer.java
romanstrobl: I will leave the implementation as is for now, because I do not think a generic error handler for `IOException` is a good idea and we want to avoid unhandled 500 errors. We can improve it later in case of any issues, however this is a corner case which should not occur in actual implementation because the collected client data object is provided by the authenticator and should be serialized correctly.

PR #1270, opened 2024-01-23
Fix #1269: Document offline anti-fraud check

### LINKED ISSUE #1269: Document offline anti-fraud check
A follow-up to #934

banterCZ:


zcgandcomp (commented):
Do we need to rename the model? This can bring issues with generated client side. It is considered as compatible change. And yes this can bring more clarity into a data model for us, but is this necessary?

banterCZ:
> Do we need to rename the model? This can bring issues with generated client side. It is considered as compatible change. And yes this can bring more clarity into a data model for us, but is this necessary?

Swagger-ui is confused by a class with the same name and generates the same schema for create and verify.

zcgandcomp (approved):
Looks OK. If we want to make nonce in a different lenght we will change it at that particular time.

PR #1276, opened 2024-01-25
Fix #1275: Improve logging of verifyOfflineSignature

### LINKED ISSUE #1275: Improve logging of verifyOfflineSignature

banterCZ:


romanstrobl (changes requested):
This change is not consistent with other PowerAuth server logging. I suggest we introduce a new logging message or update other endpoints (verify signature, verify token) to contain similar logging.

banterCZ:
> This change is not consistent with other PowerAuth server logging. I suggest we introduce a new logging message or update other endpoints (verify signature, verify token) to contain similar logging.

Thanks, good point. Let's improve it in those methods as well.

romanstrobl (approved):
Agree, logging these results will improve usefullness of logs.

PR #1278, opened 2024-01-26
Fix #1277: FIDO2: Add a protocol check into existing services

### LINKED ISSUE #1277: FIDO2: Add a protocol check into existing services
Check that an activation uses `POWERAUTH` protocol for existing PowerAuth services.

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -2001,6 +2001,13 @@ private ActivationRecovery createRecoveryCodeForActivation(ActivationRecordEntit
                 throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
             }
 
+            // Check if protocol is POWERAUTH
+            if (!Protocols.POWERAUTH.toString().equals(activationEntity.getProtocol())) {
+                logger.warn("Invalid protocol in method createRecoveryCodeForActivation");
```
banterCZ, line 2006: Nitpick: Personally, instead of the comment, I prefer a method. Moreover, `toString()` is fragile. I would call `POWERAUTH.getProtocol()` or even introduce a new method `POWERAUTH.isProtocol(String)` (or some better name).

banterCZ (approved):
Looks good, however, I suggest considering better naming.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Good point, I will introduce methods for these checks.

banterCZ (approved):

PR #1282, opened 2024-01-28
Fix #1281: FIDO2: Add protocol into activation status endpoint response

### LINKED ISSUE #1281: FIDO2: Add protocol into activation status endpoint response
The activation status endpoint should expose the `protocol` parameter which differenciates `powerauth` and `fido2` activations. Similarly, the callbacks should allow exposing the `protocol` parameter.

romanstrobl:


banterCZ (approved):
Looks good.

PR #1286, opened 2024-01-28
Fix #1285: FIDO2: Review provider API

### LINKED ISSUE #1285: FIDO2: Review provider API
Review the provider API and clean it up.

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthRegistrationProvider.java
```diff
@@ -75,7 +75,7 @@ public RegistrationChallenge provideChallengeForRegistration(String userId, Stri
 
     @Override
     @Transactional
-    public RegistrationChallenge provideChallengeForRegistrationChallengeValue(String applicationId, String challengeValue) throws Fido2AuthenticationFailedException {
+    public RegistrationChallenge findRegistrationChallengeByValue(String applicationId, String challengeValue) throws Fido2AuthenticationFailedException {
```
banterCZ, line 78: Out of the scope of this PR: May this method be `@Transacional(readonly = true)`?

banterCZ (dismissed):
Thanks, the naming is improved.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthRegistrationProvider.java
romanstrobl: I will review the `@Transactional` annotations, good point.

banterCZ (approved):
Thanks for marking transactions read-only.

PR #1295, opened 2024-02-05
Fix #1294: Extract zxing version to the maven property

### LINKED ISSUE #1294: Extract zxing version to the maven property
Extract the version of `com.google.zxing:core` to the maven property and make it sure, that it is the same for `com.google.zxing:javese` as well.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1301, opened 2024-02-08
Fix #1300: Handle explicit null JSON value in final fields

### LINKED ISSUE #1300: Handle explicit null JSON value in final fields
Use `@JsonSetter(nulls = Nulls.SKIP)` for final fields in request objects.

banterCZ:


romanstrobl (approved):
Good solution!

PR #1305, opened 2024-02-19
Fix #1304: FIDO2: Use templateName instead of operationType parameter for challenges

### LINKED ISSUE #1304: FIDO2: Use templateName instead of operationType parameter for challenges
The request for FIDO2 challenges uses `operationType` as a parameter. This is not completely correct, because the operation is created with a specific `templateName`, see:
https://github.com/wultra/powerauth-server/blob/95b3325559c46ccb26c2d398e365d753dcaf6b41/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthAssertionProvider.java#L79

We should rename the `operationType` parameter to `templateName` to avoid any confusion.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #1338, opened 2024-02-23
Fix #1337: Missing NOT NULL constraint in migration guide

### LINKED ISSUE #1337: Missing NOT NULL constraint in migration guide
The `config_key` should be `NOT NULL` in :
```
CREATE TABLE pa_application_config
(
    id                 INTEGER NOT NULL PRIMARY KEY,
    application_id     INTEGER NOT NULL,
    config_key         VARCHAR(255),
    config_values      TEXT
);
```

romanstrobl:


banterCZ (approved):
A nice catch!

PR #1342, opened 2024-02-26
Fix #1341: Coverity: Useless call

### LINKED ISSUE #1341: Coverity: Useless call
414578 [Useless call](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_USELESS_CALL)
A function call that seems to have an intended effect has no actual effect on the logic of the program.

In com.​wultra.​security.​powerauth.​rest.​client.​PowerAuthFido2RestClient.​PowerAuthFido2RestClient(java.​lang.​String, com.​wultra.​security.​powerauth.​rest.​client.​PowerAuthRestClientConfiguration): A function is called that is only useful for its return value, and this value is ignored.

banterCZ:


powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthFido2RestClient.java
```diff
@@ -84,7 +83,6 @@ public PowerAuthFido2RestClient(String baseUrl, PowerAuthRestClientConfiguration
             if (config.getProxyUsername() != null) {
                 proxyBuilder.username(config.getProxyUsername()).password(config.getProxyPassword());
             }
-            proxyBuilder.build();
```
banterCZ, line 87: Lime-core API is a little bit confusing. `proxyBuilder.build()` returns `DefaultRestClient.Builder`, which has no effect here.

romanstrobl (approved):
Good catch!

jandusil (approved):
Nice

PR #1344, opened 2024-02-26
Fix #1343: Coverity: Dereference null return value

### LINKED ISSUE #1343: Coverity: Dereference null return value

banterCZ:


romanstrobl (approved):
Great!

jandusil (approved):
Great, thanks

PR #1346, opened 2024-02-26
Fix #1345: Coverity: Reliance on default encoding

### LINKED ISSUE #1345: Coverity: Reliance on default encoding
414583 [Dm: Dubious method used](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_FB)
In com.​wultra.​powerauth.​fido2.​rest.​model.​converter.​serialization.​CollectedClientDataDeserializer.​deserialize(com.​fasterxml.​jackson.​core.​JsonParser, com.​fasterxml.​jackson.​databind.​DeserializationContext): Found a call to a method which will perform a byte to String (or String to byte) conversion, and will assume that the default platform encoding is suitable.

banterCZ:


jandusil (approved):
Nice

PR #1352, opened 2024-02-26
Fix #1351: Coverity: Useless call

### LINKED ISSUE #1351: Coverity: Useless call
414578 [Useless call](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_USELESS_CALL)
A function call that seems to have an intended effect has no actual effect on the logic of the program.

In com.​wultra.​security.​powerauth.​rest.​client.​PowerAuthFido2RestClient.​PowerAuthFido2RestClient(java.​lang.​String, com.​wultra.​security.​powerauth.​rest.​client.​PowerAuthRestClientConfiguration): A function is called that is only useful for its return value, and this value is ignored.

banterCZ:


jandusil (approved):
looks okay

Issue #1299, opened 2024-02-07
FIDO2: Check AAGUID during registration


romanstrobl:
Add an option to configure allowed AAGUIDs which can be used when registering FIDO2 tokens. This configuration should be bound to the application.

Additionally, allow to configure accepted attestation formats and check them during registration.

PR #1353, opened 2024-02-26
Fix column name pa_application_config.config_key in db scripts

banterCZ:
A follow-up to #1299


jnpsk (approved):
great, I just noticed this typo when applying changes

romanstrobl (approved):
Ah yes, thank you!

jandusil (approved):
Great catch, thanks

PR #1354, opened 2024-02-27
Fix #1310: FIDO2: Persist and check counter value

### LINKED ISSUE #1310: FIDO2: Persist and check counter value
Implement persistance and checking of counter value based on: https://www.w3.org/TR/webauthn-3/#sctn-sign-counter

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/fido2/PowerAuthCryptographyService.java
```diff
@@ -68,6 +81,25 @@ private boolean verifySignature(CollectedClientData clientDataJSON, Authenticato
         return signatureUtils.validateECDSASignature(clientDataJSONEncodedHash, signature, publicKey);
     }
 
+    private boolean checkAndPersistCounter(String applicationId, String authenticatorId, int signCount) {
+        final List<ActivationRecordEntity> activations = activationRepository.findByExternalId(applicationId, authenticatorId);
+        if (activations.size() != 1) {
+            logger.warn("Multiple activations with same external ID found, external ID: {}", authenticatorId);
```
banterCZ, line 87: The message is not correct when the size is zero.

romanstrobl: Good point, I will update the logic.

PR #1367, opened 2024-03-04
Fix #1366: testUpdateActivation_badRequest fails for a non-English locale

### LINKED ISSUE #1366: testUpdateActivation_badRequest fails for a non-English locale
When the operation system runs a non-English locale, `PowerAuthControllerTest#testUpdateActivation_badRequest` fails.

banterCZ:


jandusil (approved):
Okay

PR #1373, opened 2024-03-04
Fix #1372: Log a successfull callback call as an info

### LINKED ISSUE #1372: Log a successfull callback call as an info
Currently, we log unsuccessful calls as "WARN" but successful as "DEBUG" - as a result, we do not monitor the ratio of successful / unsuccessful callbacks in our monitoring. We should change log level to "INFO".

petrdvorak:


banterCZ (approved):
A good idea.

romanstrobl (approved):
Nice improvement.

PR #1389, opened 2024-03-07
Fix #1388: FIDO2: Reversed order of parameters in findByCredentialId

### LINKED ISSUE #1388: FIDO2: Reversed order of parameters in findByCredentialId

romanstrobl:


banterCZ (approved):
A nice catch!
In the future API, let's not use several method parameters of the same type in a row.

PR #1392, opened 2024-03-07
Fix #1374: Personalized FIDO2 assertion challenge

### LINKED ISSUE #1374: Personalized FIDO2 assertion challenge
Currently, we allow fetching an anonymous assertion challenge. We should allow personalizing the challenge immediately. As a result, we can return list of supported FIDO2 authenticators right away and ensure future check of the authenticator that was actually used against allowed authenticators.

petrdvorak:


romanstrobl (approved):
Looks OK.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/AssertionChallengeConverter.java
```diff
@@ -48,6 +55,65 @@ public AssertionChallengeResponse fromChallenge(AssertionChallenge source) {
...
+        destination.setApplications(source.getApplicationIds());
+        destination.setTemplateName(source.getTemplateName());
+        destination.getParameters().putAll(source.getParameters());
+
+        if (authenticatorDetails != null && !authenticatorDetails.isEmpty()) {
+            final Set<String> allowCredentials = new LinkedHashSet<>();
+            for (AuthenticatorDetail ad : authenticatorDetails) {
+                allowCredentials.add(ad.getExternalId());
+            }
+            destination.setAdditionalData(Map.of("allowCredentials", allowCredentials));
```
zcgandcomp, line 82: If possible do not put this logic - really on the `activationId`. But at least put proper externalized constant, i.e., ATTR_ALLOW_CREDENTIALS -  with modification of visibility.

zcgandcomp (changes requested):

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/service/AssertionService.java
```diff
@@ -75,12 +77,25 @@ public AssertionService(CryptographyService cryptographyService, AuthenticatorPr
...
     public AssertionChallengeResponse requestAssertionChallenge(AssertionChallengeRequest request) throws Exception {
-        final AssertionChallenge assertionChallenge = assertionProvider.provideChallengeForAssertion(
-                request.getApplicationIds(), request.getTemplateName(), request.getParameters(), request.getExternalId()
-        );
+
+        final List<AuthenticatorDetail> authenticatorDetails = new ArrayList<>();
+
+        // If user ID is specified, fetch the user authenticators that should be allowed to respond the challenge
+        final String userId = request.getUserId();
+        if (userId != null) {
```
zcgandcomp, line 85: We should move this logic to the provider. 
As of now:

- This service gets a list of possible authenticators from the provider. 
- Sends the list to the provider.
- The provider adds the list to his response.

Getting the list of possible authenticators and adding them to the response on the same level will increase readability.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/AssertionChallengeConverter.java
petrdvorak: The change to activation IDs needs to be done later, it seems too fragile to fix in the same change.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/service/AssertionService.java
petrdvorak: Didn't improve readability much, but OK

romanstrobl (dismissed):
Nice improvements!

zcgandcomp (dismissed):
Acceptable.

romanstrobl (approved):
Approved.

PR #1394, opened 2024-03-11
FIDO2: Basic Attestation

romanstrobl:


jnpsk (approved):
nice

zcgandcomp (approved):
Trust path verification works - even with intermediate certs.
Trust path verification works only with a correct CA.
The correct public key is used for the signature verification.
We can have more tests, but fine for now.

PR #1396, opened 2024-03-13
Fix #1393: FIDO2: externalize authenticator configuration

### LINKED ISSUE #1393: FIDO2: externalize authenticator configuration
Currently, Fido2Authenticators.java contains a hardcoded configuration for the FIDO authenticator.

Externalize the config so it can be changed by configuration. Suggestion DB - but be aware of the performance consequences.

The configuration changes are not expected to be often - in production for the first release and more frequent for setting in the test environments. So singleton with long eviction should be fine.

jnpsk:
To externalise authenticator configuration a new table `pa_fido2_authenticator` was created. It contains the same details about a fido2 authenticator as `Fido2Authenticators` did.   
The initial records are inserted to the table during liquibase update from `aaguid.csv` file converted from the `Fido2Authenticators` class.

To handle increased time complexity of db query, a spring caching mechanism is employed. The cache flush is scheduled task with fixed delay (defaults to 1hr).

This is just a draft of the initial approach. We could implement a REST API to manage the FIDO2 Authetnicators. Then the cache flushing could be handled on the table update. Does it worth it?


powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/config/CachingConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Jan Pesek, jan.pesek@wultra.com
+ */
+@Configuration
+@Slf4j
+public class CachingConfig {
+
+    @CacheEvict(value = "fido2-authenticators-cache", allEntries = true)
+    @Scheduled(fixedDelayString = "${powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction:3600000}")
+    public void evictFido2AuthenticatorCache() {
+        logger.info("Flush FIDO2 Authenticators cache.");
```
romanstrobl, line 38: Can we switch the cache flush to `debug` level?

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/database/entity/Fido2AuthenticatorEntity.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.Serializable;
+
+/**
+ * Entity representing a FIDO2 Authenticator details.
+ *
+ * @author Jan Pesek, jan.pesek@wultra.com
+ */
+@Entity
+@Table(name = "pa_fido2_authenticator")
+@Data
```
romanstrobl, line 35: The `@Data` annotation produces `equals` and `hashCode` implementations which are not suitable for JPA classes. Can we add our own implementations?

romanstrobl (approved):
Good job. I found several minor issues.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/database/entity/Fido2AuthenticatorEntity.java
jnpsk: Thanks. Now both `Fido2Authenticator` and `Fido2AuthenticatorEntity` has `equals` and `hashCode` based on `aaguid` only.

docs/db/changelog/changesets/powerauth-java-server/1.7.x/db.changelog-version.xml
```diff
@@ -6,5 +6,6 @@
     <include file="20240115-add-columns-fido2.xml" relativeToChangelogFile="true" />
     <include file="20240212-application-config.xml" relativeToChangelogFile="true" />
     <include file="20240222-add-tag-1.7.0.xml" relativeToChangelogFile="true" />
+    <include file="20240312-fido2-authenticator.xml" relativeToChangelogFile="true" />
```
romanstrobl, line 9: The order seems incorrect, release tag should be after last change.

romanstrobl:
We should update migration guide in `docs/PowerAuth-Server-1.7.0.md` to reflect DB changes.

romanstrobl:
We should also update file `docs/sql/mssql/migration_1.6.0_1.7.0.sql`.

jnpsk:
- Database new table was described in md files.
- New database structure png was generated.
- Sql scripts to manual migration were generated.

romanstrobl (approved):
Database documentation was updated.

docs/db/changelog/changesets/powerauth-java-server/1.7.x/20240312-fido2-authenticator.xml
```diff
@@ -0,0 +1,54 @@
...
+  ~ GNU Affero General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
```
banterCZ, line 23: 👀 Nice, but mind that in Maven, we have a transitive dependency on Liquibase 4.24.0 but in Docker, we still use 4.23.2 cc @zcgandcomp I do not expect any big discrepancy, just highlighting that.

docs/images/arch_db_structure.png
banterCZ, line 1: We should agree on whether to use dark mode or not.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/database/entity/Fido2AuthenticatorEntity.java
```diff
@@ -0,0 +1,68 @@
...
+    private String description;
+
+    @Enumerated(EnumType.STRING)
+    @Column(name = "signature_type", nullable = false)
+    private SignatureType signatureType;
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
banterCZ, line 58: Shouldn't we use `ProxyUtils`? https://github.com/wultra/development-internal/wiki/Equals-and-hashCode-in-JPA-entities

docs/images/arch_db_structure.png
jnpsk: I used the white one to match developer portal colours. We used that one before.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/database/entity/Fido2AuthenticatorEntity.java
jnpsk: I haven't found any rules of when a Proxy object is created on top of an Entity. To be safe and consistent with our recommendations I added the `ProxyUtils`.

banterCZ (approved):
Looks good.

PR #1398, opened 2024-03-15
Fix #1397: Return structured device info in the operation detail resp…

### LINKED ISSUE #1397: Return structured device info in the operation detail response
Currently, we only return the structured device info in the `audit_log` but not in operation detail. This should be improved (there is no need to store new data, just converting them in the response from the already present User-Agent will do the trick),

jandusil:
…onse


jandusil:
@petrdvorak 

Just a question - when the operation is created, the additionalData is set to `null.` Therefore, it is only after the operation is cancelled/approved/rejected when the userAgent is stored with the operation. Is that the **full** desired behaviour? Thanks
https://github.com/wultra/powerauth-server/blob/41e5750ba823dc423f7de5dd53bf9ab8d5e57b85/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java#L192

jandusil:
Tested that the change gets propagated to PAC:

Screenshot 2024-03-15 at 1 21 03

romanstrobl (approved):
Looks OK.

petrdvorak:
@jandusil Looks OK. The behavior you mention is correct. When the operation is created, there is no way to assign the user agent (and device info, consequently) to it immediately. At one point, I was thinking about enriching the additional data as the operation list is fetched which, in theory, is possible. But fetching can happen on multiple mobile devices, which is a problem. It would significantly complicate things and would not help with any investigations much.

petrdvorak (approved):

petrdvorak (changes requested):
How about operation callbacks?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -877,20 +877,23 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
...
+            final Map<String, Object> additionalDataExtended = new HashMap<>(additionalData);
+            parseDeviceFromUserAgent(additionalDataExtended).ifPresent(device ->
+                    additionalDataExtended.put(ATTR_DEVICE, device));
+            return additionalDataExtended;
+        }
+        return Collections.emptyMap();
     }
 
-    private static Optional <UserAgent.Device> parseDeviceFromUserAgent(Map<String, Object> additionalData) {
+    public static Optional <UserAgent.Device> parseDeviceFromUserAgent(Map<String, Object> additionalData) {
```
banterCZ, line 890:
```suggestion
    public static Optional<UserAgent.Device> parseDeviceFromUserAgent(Map<String, Object> additionalData) {
```

zcgandcomp (changes requested):
I have few remarks
- the method `convertFromEntity` enhance the  additional data with `device` object. But the data for audited are again parsed using `extendAdditionalDataWithDevice` - this is redundant.
- the method `convertFromEntity` is called frequently, but only for very few places we need to show `device` object (actually only on PAC getOperationDetail) - depends on the performance of the method - as we are playing with regexp - to do the conversion only when necessary and not all the time. (All the time is a good approach if the performance hit is small)
-

jandusil:
@zcgandcomp @petrdvorak, I've updated the solution. However, I've noticed that currently, when the methods `attemptApproveOperation`, `failApprovalOperation`, and `rejectOperation` encounter failures (i.e., they proceed to the else statement), we do not audit the UserAgent. Should we implement auditing for these scenarios as well? Thanks.

petrdvorak:
@jandusil If it is known in such a scenario, we could log it.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -595,11 +598,13 @@ public OperationDetailResponse cancelOperation(OperationCancelRequest request) t
                 .param("id", operationId)
                 .param("failureCount", operationEntity.getFailureCount())
                 .param("status", operationEntity.getStatus().name())
-                .param("additionalData", extendAuditedAdditionalData(operationEntity.getAdditionalData()))
+                .param("additionalData", extendAdditionalDataWithDevice(operationEntity.getAdditionalData()))
                 .build();
         audit.log(AuditLevel.INFO, "Operation canceled via explicit server call for operation ID: {}", auditDetail, operationId);
+        final OperationDetailResponse operationDetailResponse = convertFromEntity(savedEntity);
+        extendAdditionalDataWithDevice(operationDetailResponse);
```
zcgandcomp, line 605: But correctly we are returning the user agent when canceling the operation. This is good.

zcgandcomp (approved):
Ok from my POV it's OK.

Enriched detail is returned when relevant (detail, result of cancel).
Enriched detail is sent via callbacks.
All audit events are enriched (except create where it makes no sense)

Please if @banterCZ approves the java code itself we can merge.

banterCZ (approved):
Code style looks good.

petrdvorak (approved):
Approved

PR #1405, opened 2024-03-18
Fix #1404: FIDO2: REST API request / response class cleanup

### LINKED ISSUE #1404: FIDO2: REST API request / response class cleanup
Cleanup request and response class parameter names for first release.

romanstrobl:


petrdvorak (dismissed):
Looks OK. In next version, we could return a bit more of PowerAuth context (it might be helpful but requires more elaboration).

jnpsk (approved):
looks ok

PR #1409, opened 2024-03-18
Fix #1407: Reflect fido2 changes in client model

### LINKED ISSUE #1407: Reflect fido2 changes in client model
Issue https://github.com/wultra/powerauth-server/issues/1374 introduced changes in response / requests that are not reflected in `powerauth-client-model`.

jnpsk:
Issue https://github.com/wultra/powerauth-server/issues/1374 does not reflect changes made to requests/responses to client model.

There was also an error `ArrayList cannot be cast to class Set` while deserialising additional data to set, which was fixed by using list instead.


banterCZ (approved):
Looks good.

romanstrobl (approved):
Thanks for catching this!

PR #1418, opened 2024-03-19
Fix #1417: FIDO2: Invalid AAGUID format in configuration of fido2_aaguids_allowed

### LINKED ISSUE #1417: FIDO2: Invalid AAGUID format in configuration of fido2_aaguids_allowed
Conversion from `byte[]` to UUID is missing, the `byte[]` is interpreted as a String.

romanstrobl:
I switched the AAGUID configuration to UUID, it was incorrectly expected as raw bytes.


jnpsk (approved):

petrdvorak (approved):
Looks OK

PR #1426, opened 2024-03-25
Fix #1425: FIDO2: Return operation data in a separate allowCredential item

### LINKED ISSUE #1425: FIDO2: Return operation data in a separate allowCredential item

petrdvorak:


jnpsk (dismissed):
looks ok

PR #1427, opened 2024-03-26
Fix #1425: FIDO2: Return operation data in a separate allowCredential item

### LINKED ISSUE #1425: FIDO2: Return operation data in a separate allowCredential item

jnpsk:
Added Base64 encoding for the operation data.


banterCZ (approved):
Looks good, let's merge it, but we should cover this class with a unit test later on.

PR #1432, opened 2024-03-27
Fix #1430: Coverity: Unread field should be static

### LINKED ISSUE #1430: Coverity:  Unread field should be static
SpotBug and also Sonar issue https://rules.sonarsource.com/java/RSPEC-1170/

> Making a public constant just final as opposed to static final leads to duplicating its value for every instance of the class, uselessly increasing the amount of memory required to execute the application.
>
> Further, when a non-public, final field isn’t also static, it implies that different instances can have different values. However, initializing a non-static final field in its declaration forces every instance to have the same value. So such fields should either be made static or initialized in the constructor.

banterCZ:


romanstrobl (approved):
Looks good.

PR #1433, opened 2024-03-27
Fix #1431: Coverity: Bad casts of object references

### LINKED ISSUE #1431: Coverity: Bad casts of object references
In com.​wultra.​powerauth.​fido2.​database.​entity.​Fido2AuthenticatorEntity.​equals(java.​lang.​Object): The equals(Object o) method shouldn't make any assumptions about the type of object.

banterCZ:


banterCZ:
A False positive Coverity report.

PR #1439, opened 2024-03-28
Fix #1436: FIDO test app Opera browser - authenticatorAttachment error

### LINKED ISSUE #1436: FIDO test app Opera browser - authenticatorAttachment error
1. First I have registred a new user
2. Before logging in I have changed the "User Verification" to "preferred" -> see error message

jnpsk:
Although I found this in [level 3 spec](https://w3c.github.io/webauthn/#iface-pkcredential) only, the `authenticatorAttachment` can be null in both ceremonies. Which is also supported by the [taken test](https://github.com/wultra/powerauth-server/issues/1436).

This test also found out there was missing a cascade validation of parameters class in registration request and so the nested validation annotations were without effect.


banterCZ (approved):
Looks good. We will resolve the duplicity of model classes in another PR.

PR #1463, opened 2024-04-16
Fix #1437: Change ActivationRecordEntity#protocol into enum

### LINKED ISSUE #1437: Change ActivationRecordEntity#protocol into enum
Right now, it is a plain string.

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -303,6 +303,26 @@ public GetActivationListForUserResponse getActivationList(String applicationId,
...
+        if (source == null) {
+            return null;
+        }
+        return switch(source) {
+            case FIDO2 -> ActivationProtocol.FIDO2;
+            case POWERAUTH -> ActivationProtocol.POWERAUTH;
+        };
+    }
+
+    private static ActivationProtocol covert(final io.getlime.security.powerauth.app.server.database.model.enumeration.ActivationProtocol source) {
```
banterCZ, line 316: typo

romanstrobl (approved):
Nice improvement!

PR #1467, opened 2024-04-18
Fix #1399: Claiming of non-personalized operation scoped to activationId doesn't check the userId - activationId relationship

### LINKED ISSUE #1399: Validate activation ID when creating operation
Originally reported as a problem when claiming non-personalized operation. It is possible to specify a user ID different from the activation one.

The root cause is already in the operation creation. It has to be validated that if the activation ID is specified, the user ID is also specified and matches the value in the activation (avoid usage of the activation of a different user).

This issue does not affect security, just only data consistency.

banterCZ:


banterCZ:
Fixed in #1470

PR #1469, opened 2024-04-18
Fix #1468: Improve performance of deleting values in table pa_unique_values

### LINKED ISSUE #1468: Improve performance of deleting values in table pa_unique_value
The method `deleteAllByTimestampExpiresBefore` uses a `SELECT` and deletes records in database individually. We should use a `@Query` with `DELETE` statement using JPQL.

romanstrobl:


banterCZ (approved):
It was really tricky!

PR #1470, opened 2024-04-18
Fix #1399: Validate activation ID when creating operation

### LINKED ISSUE #1399: Validate activation ID when creating operation

banterCZ:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationCreateRequest.java
```diff
@@ -62,7 +62,7 @@ public class OperationCreateRequest {
     @Schema(description = "Whether proximity check should be used, overrides configuration from operation template", requiredMode = Schema.RequiredMode.NOT_REQUIRED)
     private Boolean proximityCheckEnabled;
 
-    @Schema(description = "Activation Id of a specific device", requiredMode = Schema.RequiredMode.NOT_REQUIRED, maxLength = 37)
+    @Schema(description = "Optional activation ID. It is possible to specify a single device (otherwise all user's activations are taken into account).", requiredMode = Schema.RequiredMode.NOT_REQUIRED, maxLength = 37)
```
zcgandcomp, line 65: Optionality is declared in the Schema

zcgandcomp (approved):
Looks OK.

jnpsk (approved):
looks ok

PR #1504, opened 2024-04-30
Fix #1502: FIDO2: Return excludeCredentials in RegistrationChallenge

### LINKED ISSUE #1502: FIDO2: Return excludeCredentials in RegistrationChallenge

jnpsk:
- `RegistrationChallengeResponse` now contains `excludeCredentials`, which is a list of existing credentials.
- `AllowCredentials` data class was renamed to `Credential` to serve both - allowCredential and excludeCredential lists. This has to be reflected on the client side.

### DIFF
--- a/powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/entity/Credential.java
+++ b/powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/entity/Credential.java
@@ -25,7 +25,7 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * Representation of an allowed authenticator instance.
+ * Representation of a FIDO2 Credential.
  *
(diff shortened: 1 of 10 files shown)


powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/entity/Credential.java
```diff
@@ -34,7 +34,7 @@
 @ToString
 @Builder
 @Jacksonized
-public class AllowCredentials {
+public class Credential {
```
banterCZ, line 37: I am ok with renaming, but it should be reflected in JavaDoc.

banterCZ (approved):
Looks good.

romanstrobl (approved):
I agree with the change and refactoring.

PR #1510, opened 2024-05-06
Fix #1482: Optimization: Operation expiration optimization

### LINKED ISSUE #1482: Optimization: Operation expiration optimization
The PowerAuth Server expires operation based on the attribute `timestampExpires`.
When a large number of operations is created the expiration can take very long time (days)
When selecting expired operations by status and `timestampExpires`. Index `pa_operation_status_exp` is used. However, the usage of the index provides worse performance.

Optimize the task:

- Limit the number of expired operations (e.g., 100) to prevent long-running sessions and unplanned resource exhaustion on the server.
- Drop unused index - check if the field is not used for selection somewhere else - in this case we cannot drop it.

jnpsk:
Changes:

1. Reordered columns in `pa_operation_status_exp` index: `pa_operation(timestamp_expires, status) -> pa_operation(status, timestamp_expires)`. Now the index is used during the query for expired pending operations.
2. Scheduled job to expire pending operation now sets as expired 100 operations only. This number is configurable via a new property.
3. CallbackUrl entities are no longer fetched via explicit repository call, but via getter on the entity.


banterCZ (approved):

romanstrobl (approved):
Useful performance improvement.

PR #1511, opened 2024-05-07
Fix #1508: Validate ValidateTokenRequest#protocolVersion parameter

### LINKED ISSUE #1508: Validate `ValidateTokenRequest#protocolVersion` parameter
The following property, `protocolVersion`, was added in 1.5.x, but not validated
https://github.com/wultra/powerauth-server/blob/30bfa16bbfda9f9b4e4ea02e498d75aa1becbeda/powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/ValidateTokenRequest.java#L37

As a result, implementors coming from older versions will have missing protocol versions in their requests and get `NullPointerException`.

```log
java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "<local7>" is null
	at io.getlime.security.powerauth.crypto.lib.util.TokenUtils.computeTokenDigest(TokenUtils.java:104)
	at io.getlime.security.powerauth.crypto.lib.util.TokenUtils.validateTokenDigest(TokenUtils.java:124)
	at io.getlime.security.powerauth.crypto.server.token.ServerTokenVerifier.validateTokenDigest(ServerTokenVerifier.java:55)
	at io.getlime.security.powerauth.app.server.service.behavior.tasks.TokenBehavior.validateToken(TokenBehavior.java:316)
```

banterCZ:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/ValidateTokenRequest.java
```diff
@@ -29,12 +30,21 @@
...
+
     @ToString.Exclude
+    @NotNull
     private String tokenDigest;
+
+    @NotNull
     @ToString.Exclude
     private String nonce;
+
+    @NotNull
```
banterCZ, line 44: First of all, it is propagated into swagger. Last but not least, it will be helpful if we switch to bean validation.

jnpsk (approved):
I agree with using bean validation

banterCZ:
> I agree with using bean validation

Bean validation is not activated yet (to keep current error handling).

romanstrobl (approved):
Good, thanks!

PR #1512, opened 2024-05-09
Fix #1472: Investigate @Scheduled methods with @Transactional annotation

### LINKED ISSUE #1472: Investigate @Scheduled methods with @Transactional annotation
We should review the behavior of the combination of scheduling and transactional behavior, just to be sure there could be no strange side effects.

banterCZ:
In general, the usage of `@Scheduled` and `@Transactional` together over a method should work. But let's improve the design, split them into separate layers, and isolate responsibilities.


powerauth-java-server/src/main/resources/application-dev.properties
```diff
@@ -3,3 +3,6 @@ spring.liquibase.enabled=true
 spring.liquibase.change-log=classpath:db/changelog/db.changelog-module.xml
 
 powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction=60000
+
+logging.level.com.wultra=DEBUG
+logging.level.io.getlime=DEBUG
```
romanstrobl, line 8: Is this change intentional?

romanstrobl (approved):
Looks OK.

jnpsk (approved):
nice!

PR #1515, opened 2024-05-10
Fix #1514: Superfluous combination of Date and Instance

### LINKED ISSUE #1514: Superfluous combination of Date and Instance
https://github.com/wultra/powerauth-server/blob/70d8ff0de22c4a8b662559542f703a034aff6e0c/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/replay/ReplayPersistenceService.java#L92

banterCZ:


romanstrobl (approved):
Looks OK.

PR #1521, opened 2024-05-20
Fix #1517: FIDO2: Enforce USB transport hint for WAU authenticators

### LINKED ISSUE #1517: FIDO2: Enforce USB transport hint for WAU authenticators
Currently, some browsers / OS combinations omit the sending of transport hints on authenticator registration. As a result, the customers who only use our WAU authenticators may provide an experience with the hybrid authenticator, which is not optimal (i.e., shows the QR code).

We should enforce `"transport": [ "usb" ]` hint in `allowCredentials` items for WAU authenticators.

jnpsk:
A new column `transports` has been added to the `pa_fido2_authenticator` table, which values is used during the allow/exclude credential list build as a fallback in case a browser failed to provide transport hints during the registration ceremony.

Effect on **non-Wultra authenticators**:
- If the DB column is not set or contains empty list, there is no behaviour changes compared to the current version.
- If the DB column contains non-empty list, the list will be used in case a browser did not provide transport hints during the registration ceremony.

Effect on **Wultra authenticator**:
- If the DB column for WAU record is not set or contains empty list, `["usb"]` will be used in case a browser did not provide transport hints during the registration ceremony.
- If the DB column for WAU record is non-empty list, the assigned list will be used in case a browser did not provide transport hints during the registration ceremony.

**The column `transports` is ignored in case a browser provided a relevant transport hints.**

### DIFF
--- a/docs/PowerAuth-Server-1.8.0.md
+++ b/docs/PowerAuth-Server-1.8.0.md
@@ -16,3 +16,11 @@ For manual changes use SQL scripts:
 
 Existing database index `pa_operation_status_exp` on the `pa_operation` table was modified to improve performance of the
 process of expiration of pending operations.
+
+### Add transports Column to pa_fido2_authenticator
+
+A new column `transports` has been added to the `pa_fido2_authenticator` table. The column allows you to assign
+transport hints to a FIDO2 Authenticator registered in the table. These transport hints will be used to build allow
+credential list or exclude credential list during WebAuthn ceremonies, serving as a fallback if the client fails to
+provide transport hints when registering a new credential. The format of the column is a list of authenticator transport
+values supported by the WebAuthn protocol, serialized as a JSON array.
--- a/docs/PowerAuth-Server-1.8.0.md
+++ b/docs/PowerAuth-Server-1.8.0.md
@@ -17,7 +17,7 @@ For manual changes use SQL scripts:
 Existing database index `pa_operation_status_exp` on the `pa_operation` table was modified to improve performance of the
 process of expiration of pending operations.
 
-### Add transports Column to pa_fido2_authenticator
+### Add Transports Column to pa_fido2_authenticator
 
 A new column `transports` has been added to the `pa_fido2_authenticator` table. The column allows you to assign
 transport hints to a FIDO2 Authenticator registered in the table. These transport hints will be used to build allow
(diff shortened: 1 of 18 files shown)


docs/db/changelog/changesets/powerauth-java-server/1.8.x/db.changelog-version.xml
```diff
@@ -5,5 +5,6 @@
                    https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
 
     <include file="20240424-index-expire-status.xml" relativeToChangelogFile="true" />
+    <include file="20240517-fido2-authenticator-transports.xml" relativeToChangelogFile="true" />
```
banterCZ, line 8: Do we want to backport this? I am asking because of the changelog version.

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/database/entity/converter/ListToStringConverter.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public List<String> convertToEntityAttribute(final String dbValue) {
+        if (!StringUtils.hasText(dbValue)) {
+            return Collections.emptyList();
+        }
+
+        try {
+            return objectMapper.readValue(dbValue, new TypeReference<>() {});
```
banterCZ, line 73: Theoretically, it may return `null`, see the source code of `ObjectMapper`.

banterCZ (dismissed):

docs/PowerAuth-Server-1.8.0.md
```diff
@@ -16,3 +16,11 @@ For manual changes use SQL scripts:
 
 Existing database index `pa_operation_status_exp` on the `pa_operation` table was modified to improve performance of the
 process of expiration of pending operations.
+
+### Add transports Column to pa_fido2_authenticator
```
romanstrobl, line 20: Inconsistent capitalization for `Add transports Column`.

docs/db/changelog/changesets/powerauth-java-server/1.8.x/db.changelog-version.xml
romanstrobl: We most likely will not need to backport this feature.

romanstrobl (dismissed):
Great!

docs/PowerAuth-Server-1.8.0.md
banterCZ:
```suggestion
### Add Transports Column to pa_fido2_authenticator
```

banterCZ (approved):

PR #1540, opened 2024-05-30
Fix #1539: Invalid import for Arrays

### LINKED ISSUE #1539: Invalid import for Arrays
There seems to be an incorrect import for `org.bouncycastle.util.Arrays`, used for:

```
final String credentialIdTrimmed = Base64.getEncoder().encodeToString(Arrays.copyOfRange(credentialIdBytes, 0, 32));
```

In other classes we use `java.util.Arrays`, so we should unify the `Arrays` usage.

romanstrobl:


banterCZ (approved):
Looks good.

PR #1543, opened 2024-05-31
Fix #1538: DB migration issue (MS SQL, possible Oracle)

### LINKED ISSUE #1538: DB migration issue (MS SQL, possible Oracle)
Step to reproduce
Create activation
Migrate to 1.7.x

Expected behavior:

Actual behavior:

The null value leads to:

```

```
when calling API

banterCZ:
This is a fix into 1.7.x; for develop, we will make the column not-null, see #1544


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/ActivationRecordEntity.java
```diff
@@ -667,6 +669,16 @@ public List<ActivationHistoryEntity> getActivationHistory() {
         return activationHistory;
     }
 
+    @PostLoad
+    public void setDefaultProtocol() {
```
zcgandcomp, line 673: This is a fix for null values in the DB. But the null values are updated in the liquibase script.

zcgandcomp (changes requested):

banterCZ:
- [ ] Let's make the column _not-null_ even for backport

romanstrobl (approved):
Good fix.

PR #1544, opened 2024-05-31
Fix #1538: DB migration issue (MS SQL, possible Oracle)

### LINKED ISSUE #1538: DB migration issue (MS SQL, possible Oracle)

banterCZ:
- Make protocol column not-null

Inspired but not cherry-picked by fix in 1.7.x #1543


docs/db/changelog/changesets/powerauth-java-server/1.7.x/20240115-add-columns-fido2.xml
```diff
@@ -49,4 +49,12 @@
         <modifyDataType tableName="pa_activation" columnName="extras" newDataType="varchar(4000)"/>
     </changeSet>
 
+    <!--  MS SQL default value issue https://github.com/liquibase/liquibase/issues/4644  -->
+    <changeSet id="4" logicalFilePath="powerauth-java-server/1.7.x/20240115-add-columns-fido2" author="Lubos Racansky" dbms="mssql">
+        <update tableName="pa_activation">
```
banterCZ, line 54: I do not see any simple way how to write a precondition for an existing constraint, even if named.

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240530-protocol-not-null.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                   https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet id="1" logicalFilePath="powerauth-java-server/1.8.x/20240530-protocol-not-null.xml" author="Lubos Racansky">
+        <comment>Make column pa_activation.protocol not-null.</comment>
```
banterCZ, line 8: I do not see any simple way how to write a precondition for an existing constraint, even if named.

docs/sql/mssql/migration_1.6.0_1.7.0.sql
```diff
@@ -12,6 +12,10 @@ GO
 ALTER TABLE pa_activation ALTER COLUMN extras varchar (4000);
 GO
 
+-- Changeset powerauth-java-server/1.7.x/20240115-add-columns-fido2::4::Lubos Racansky
+UPDATE pa_activation SET protocol = 'powerauth' WHERE protocol is null;
```
zcgandcomp, line 16: I would recommend, as this is manual work anyway to use

`alter table pa_activation add protocol varchar(255) default 'powerauth' WITH VALUES;
`
It should prevent potential performance issues. (there is no way how to do it in Liquibase via DDL)

banterCZ: There is one workaround https://github.com/liquibase/liquibase/issues/4644#issuecomment-2000458786
Since the changeset has been applied, I am afraid that we cannot use it.
But we may use native SQL in the changeset. Or are you speaking only about this SQL script?

zcgandcomp: In this SQL we should use this SQL statement.

Yes, exactly. In LB we need to add SQL update statement - as you have it not. It is fine.

zcgandcomp (dismissed):
Looks OK

banterCZ:
Let's make it as a cherry-pick of https://github.com/wultra/powerauth-server/pull/1543

PR #1550, opened 2024-06-03
Fix #1538: DB migration issue (MS SQL, possible Oracle) (forwardport)

### LINKED ISSUE #1538: DB migration issue (MS SQL, possible Oracle)

banterCZ:
* Fix problem with MS SQL default value
* Make column non-null

(cherry picked from commit 0a685281621f9e2b2e35a9385c63d5e0b41eb073)


romanstrobl (approved):
As expected.

PR #1562, opened 2024-06-05
Fix #1560: Create Operation does not return proximity OTP (backport)

### LINKED ISSUE #1560: Create Operation does not return proximity OTP
Even if the create operation request contains `proximityCheckEnabled=true`, the response is missing the proximity OTP.

banterCZ:
* Fix #1560: Create Operation does not return proximity OTP

Co-authored-by: Zdeněk Černý <zdenek.cerny@wultra.com>
(cherry picked from commit e17f6f6139b7a08824279514bc98743dfc87e7d7)


romanstrobl (approved):
Backport is OK.

PR #1565, opened 2024-06-05
Fix #1564: Missing signature version in vault unlock mapping

### LINKED ISSUE #1564: Missing signature version in vault unlock mapping

petrdvorak:


romanstrobl (approved):
Good catch!

PR #1567, opened 2024-06-06
Fix #1566: VerifyECDSASignature returns bad request for non existing activation

### LINKED ISSUE #1566: VerifyECDSASignature returns bad request for non existing activation
`PowerAuthVaultUnlockTest.vaultUnlockAndECDSASignatureNonExistentActivationTest` is failing https://github.com/wultra/powerauth-tests/blob/develop/powerauth-backend-tests/src/test/java/com/wultra/security/powerauth/test/v30/PowerAuthVaultUnlockTest.java#L136

It used to return invalid signature flag in the response.

Caused by #534

banterCZ:


romanstrobl (approved):
I agree the return value is better like this.

PR #1575, opened 2024-06-21
Fix #1552: Enable to lookup the userId if operation is created with activationId but without userId

### LINKED ISSUE #1552: Enable to lookup the userId if operation is created with activationId but without userId.
If `userId` is not present, but the operation is created with `activationId`, use the `userId` of the activation for the operation creation.

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -258,23 +257,33 @@ public OperationDetailResponse createOperation(OperationCreateRequest request) t
...
     }
 
+    private String fetchUserId(final OperationCreateRequest request) throws GenericServiceException {
+        if (request.getUserId() != null) {
+            return request.getUserId();
+        } else if (request.getActivationId() != null) {
+            final String activationId = request.getActivationId();
+            logger.debug("Filling missing user ID from the activation ID: {}", activationId);
+            return activationRepository.findById(activationId)
+                    .map(ActivationRecordEntity::getUserId)
```
zcgandcomp, line 267: This should not be null, like ever - if no activation exists, it will raise an exception later in the code.

banterCZ: Ok, let's throw an exception.

romanstrobl (approved):
Implemented as expected.

zcgandcomp (approved):
Looks OK

PR #1582, opened 2024-06-25
Remove superfluous transactional annotation

banterCZ:
A follow-up to #1472


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -64,7 +63,6 @@ public void expireOperations() {
 
     @Scheduled(fixedRateString = "${powerauth.service.scheduled.job.activationsCleanup:5000}")
     @SchedulerLock(name = "expireActivationsTask")
-    @Transactional
```
banterCZ, line 67: It is used here by accident. The service method is annotated.

PR #1595, opened 2024-07-15
Fix #341: Optimization: Callbacks invocation

### LINKED ISSUE #341: Add "removeOnBlock" switch to application
Currently, we always block the activation and it has to be removed manually. However, sometimes it would be useful to remove activation whenever it is blocked due to an invalid number of failed attempts.

- The settings should be per application.
- The settings should be context-aware (list of block reasons that lead to removal, or `*` for any)

jnpsk:
Fix [#341](https://github.com/wultra/tasklist/issues/341) by implementing the Outbox Pattern.

Instead of dealing with callback sending in the business logic, an `CallbackUrlEvent` is created and published for an event listener. The listener, implemented in `CallbackUrlEventDispatcher.class`, then processes the `CallbackUrlEvent` in a separate thread and transaction.


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -69,4 +72,20 @@ public void expireActivations() {
...
+        logger.debug("dispatchFailedCallbackUrlEvent");
+        callbackUrlEventDispatcher.dispatchFailedCallbackUrlEvent();
+    }
+
+    @Scheduled(cron = "${powerauth.service.scheduled.job.callbackUrlEventsCleanupCron:0 0 0 */3 * *}")
+    @SchedulerLock(name = "callbackUrlEventsCleanup")
+    public void cleanEvents() {
+        LockAssert.assertLocked();
+        logger.debug("deleteCallbackUrlEvents");
+        callbackUrlEventDispatcher.deleteCallbackUrlEventsAfterRetentionPeriod();
```
banterCZ, line 88: 🤔 When I see calling the _dispatcher_ from _scheduled_ task, I suggest splitting the listener and the logic into two classes (dispatcher and service).

banterCZ (commented):
Good job. Please update also the documentation. I have added several comments to consider.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
```diff
@@ -0,0 +1,74 @@
...
+@Repository
+public interface CallbackUrlEventRepository extends CrudRepository<CallbackUrlEventEntity, String> {
+
+    /**
+     * Lock timeout option to skip locked rows instead of waiting for exclusive lock.
+     */
+    String SKIP_LOCKED = "-2";
+
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints(value = {@QueryHint(name = JAKARTA_LOCK_TIMEOUT, value = SKIP_LOCKED)})
```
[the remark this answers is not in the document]
jnpsk: I've read about some problems, that these timeout hints are not working properly. However, I tried that locally on PSQL 14 and latest MSSQL and worked as expected on both platforms.
I did not find any other solution. If the hint is ignored, the query processing will be locked until an exclusive lock is acquired, which shouldn't cause any serious problems, just will make the processing slower.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventDispatcher.java
```diff
@@ -0,0 +1,329 @@
...
+        callbackUrlEventEntity.setStatus(CallbackUrlEventStatus.FAILED);
+        callbackUrlEventRepository.save(callbackUrlEventEntity);
+    }
+
+    /**
+     * Set a Callback URL Event as currently being processed.
+     * @param callbackUrlEventEntity The Callback URL Event to set as currently being processed.
+     */
+    private void markAsProcessing(final CallbackUrlEventEntity callbackUrlEventEntity) {
+        callbackUrlEventEntity.setStatus(CallbackUrlEventStatus.PROCESSING);
```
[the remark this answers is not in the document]
jnpsk: The callback is sent asynchronously by `RestClient#postNonBlocking` so the expected behavior is, that in a transaction a locked callback event record is marked as PROCESSING and dispatched. The transaction is committed right after that, meaning the callback event record will be in the PROCESSING state. The final states (FAILED or COMPLETED) are set after a response is available. I have observed this behavior during testing.
However, you made me curious about the transaction management during responses processing. I'll look at that.

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240704-callback-event-table.xml
```diff
@@ -0,0 +1,71 @@
...
+            <column name="id" type="varchar(36)">
+                <constraints primaryKey="true" nullable="false" />
+            </column>
+            <column name="application_callback_id" type="varchar(37)">
+                <constraints nullable="false" foreignKeyName="pa_application_callback_id_fk" referencedTableName="pa_application_callback" referencedColumnNames="id" />
+            </column>
+            <column name="callback_data" type="text">
+                <constraints nullable="false" />
+            </column>
+            <column name="status" type="varchar(255)" >
```
romanstrobl, line 24: Do we really need 255 chars? I think the values will be much shorter.

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240704-callback-event-table.xml
```diff
@@ -0,0 +1,71 @@
...
+
+    <changeSet id="4" logicalFilePath="powerauth-java-server/1.8.x/20240704-callback-event-table.xml" author="Jan Pesek">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="pa_application_callback" columnName="retention_period" />
+            </not>
+        </preConditions>
+        <comment>Add retention_period column to pa_application_callback table.</comment>
+        <addColumn tableName="pa_application_callback">
+            <column name="retention_period" type="varchar(255)" />
```
romanstrobl, line 67: dtto

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/AsyncConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+public class AsyncConfiguration {
+
+    @Bean(name = "callbackUrlEventsThreadPoolExecutor")
+    public Executor callbackUrlEventsThreadPoolExecutor(final PowerAuthCallbacksConfiguration powerAuthCallbacksConfiguration) {
+        final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(powerAuthCallbacksConfiguration.getThreadPoolCoreSize());
+        // Practically infinite queue size (default setting), so all Callback URL Events accepted via event listener are
+        // queued. When changing to significantly smaller queue size, an event may be rejected resulting in a PENDING
+        // Callback URL Event in the outbox table never being dispatched.
+        executor.setQueueCapacity(Integer.MAX_VALUE);
```
romanstrobl, line 44: Is this a good idea? Does not the server may run out of memory in case callbacks are called extensively (it may even be an attack). I would consider limiting the queue to a high value, but not so high.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/PowerAuthCallbacksConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Default maximum number of attempts in case the corresponding Callback URL does not define any.
+     */
+    private int defaultMaxAttempts = 1;
+
+    /**
+     * Default retention period of the event in case the corresponding Callback URL does not define any.
+     */
+    private Duration defaultRetentionPeriod = Duration.ofDays(3);
```
romanstrobl, line 46: I would suggest to keep the default longer, e.g. 30 days.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlEntity.java
```diff
@@ -90,6 +92,25 @@ public class CallbackUrlEntity implements Serializable {
     @Convert(converter = CallbackAuthenticationConverter.class)
     private CallbackUrlAuthenticationEntity authentication;
 
+    /**
+     * Maximum number of attempts to send the callback.
+     */
+    @Column(name = "max_attempts")
+    private Integer maxAttempts;
```
romanstrobl, line 99: We added new columns, do we handle correctly the migration of rows which already exist? I assume null values would be used, is this OK.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
```diff
@@ -0,0 +1,74 @@
...
+    @Query("""
+            SELECT c FROM CallbackUrlEventEntity c
+            WHERE c.id = :id
+            AND c.status = io.getlime.security.powerauth.app.server.database.model.enumeration.CallbackUrlEventStatus.PENDING
```
romanstrobl, line 56: Do we need to add a database index for this query on `status` column?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
```diff
@@ -0,0 +1,74 @@
...
+            """)
+    Optional<CallbackUrlEventEntity> findPendingByIdWithTryLock(String id);
+
+    @Query("""
+            SELECT c FROM CallbackUrlEventEntity c
+            WHERE c.status = io.getlime.security.powerauth.app.server.database.model.enumeration.CallbackUrlEventStatus.FAILED
+            AND c.timestampNextCall < :timestamp
+            """)
+    Stream<CallbackUrlEventEntity> findScheduledForRetry(LocalDateTime timestamp, Pageable pageable);
```
romanstrobl, line 65: Do we need to add a database index for this query on `status` and timestamp_next_call columns?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Modifying
+    @Query("""
+            DELETE FROM CallbackUrlEventEntity c
+            WHERE c.timestampDeleteAfter < :timestamp
+            """)
+    void deleteAllAfterRetentionPeriod(LocalDateTime timestamp);
```
romanstrobl, line 72: Do we need to add a database index for this query on `timestamp_delete_after` column?

romanstrobl (changes requested):
Good job. I have several concerns, especially related to performance and data migration which should be handled or answered.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlEventEntity.java
```diff
@@ -0,0 +1,92 @@
...
+    @Column(name = "attempts")
+    private int attempts;
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || !this.getClass().equals(ProxyUtils.getUserClass(o))) return false;
+
+        final CallbackUrlEventEntity that = (CallbackUrlEventEntity) o;
+        return callbackUrlEntity.equals(that.callbackUrlEntity) && callbackData.equals(that.callbackData);
```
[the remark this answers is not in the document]
jnpsk: the logic was modified to use only the uuid created on the entity initialisation

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
jnpsk: for now I renamed dispatcher to service, as a separated service class would just contain two methods with a single call to dispatcher

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240704-callback-event-table.xml
jnpsk: reduced to 32

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240704-callback-event-table.xml
jnpsk: reduced to 64

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
jnpsk: Added index on `(id, status)`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
jnpsk: added index on `(status, timestamp_next_call)`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
jnpsk: added index  on `(timestamp_delete_after)`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventDispatcher.java
```diff
@@ -0,0 +1,329 @@
...
+        callbackUrlEventRepository.save(callbackUrlEventEntity);
+    }
+
+    /**
+     * Handle failure of callback attempt.
+     * @param callbackUrlEventEntity Failed Callback URL Event.
+     * @param error Exception describing the cause of failure.
+     */
+    private void handleCallbackFailure(final CallbackUrlEventEntity callbackUrlEventEntity, final Throwable error) {
+        logger.warn("Callback failed, URL={}, callbackEventId={}, error={}", callbackUrlEventEntity.getCallbackUrlEntity().getCallbackUrl(), callbackUrlEventEntity.getId(), error.getMessage());
```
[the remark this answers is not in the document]
jnpsk: that left from the previous handler of callback failure. You are right, as we can monitor the callback events from the table now, we can use info logging.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlEntity.java
jnpsk: Yes, null values are fine. For that reason there are configurable defaults on application properties level.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/AsyncConfiguration.java
jnpsk: thanks for your insight. In that case another scheduled job should be implemented to handle PENDING events that was not delivered to the listener. That brings another consumer requiring exclusive lock, but I think the critical section is short enough not to cause long blocking queries.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -69,4 +72,20 @@ public void expireActivations() {
         activationServiceBehavior.expireActivations();
     }
 
+    @Scheduled(fixedRateString = "${powerauth.service.scheduled.job.retryFailedCallbackUrlEvent:3000}")
+    @SchedulerLock(name = "retryFailedCallbackUrlEvent")
```
banterCZ, line 76: Please include `lockAtLeastFor = "#{T(java.lang.Math).round(${powerauth.service.scheduled.job.retryFailedCallbackUrlEvent:3000} * 0.8)}")`, see #1597

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -69,4 +72,20 @@ public void expireActivations() {
...
+    public void callbackEvents() {
+        LockAssert.assertLocked();
+        logger.debug("dispatchFailedCallbackUrlEvent");
+        callbackUrlEventService.dispatchFailedCallbackUrlEvent();
+    }
+
+    @Scheduled(cron = "${powerauth.service.scheduled.job.callbackUrlEventsCleanupCron:0 0 0 */3 * *}")
```
banterCZ, line 83: 🤔 Is it a reasonable default? I mean, shouldn't we run the cleanup daily?

banterCZ (commented):
Great, almost done. Let's polish it a little bit.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
banterCZ: The new service is fine. I do not insist, but still think that `TransactionalEventListener` could be in a separate class (some _listener_?). @romanstrobl Any opinion?

docs/Database-Structure.md
```diff
@@ -329,6 +332,9 @@ CREATE TABLE pa_application_callback
 | type           | VARCHAR(64) | - | Callback type: `ACTIVATION_STATUS_CHANGE` or `OPERATION_STATUS_CHANGE`.   |
 | attributes     | TEXT | - | Callback attributes as a key-value map, serialized into JSON.             |
 | authentication | TEXT | - | Callback HTTP request authentication configuration, serialized into JSON. |
+| max_attempts | INTEGER | - | Maximum number of attempts to dispatch a callback. |
```
zcgandcomp, line 335: is there any expected/recommended value for infinity repetion?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
zcgandcomp, line 72: We should delete only finished callbacks. 
Personally I would not delete the callbacks at all, but ok this is customizable and there could be a issue with storage maintenance - so job itself is OK.

I would delete only the status `COMPLETED`. This will also help you against possible deadlock

Like other statuses should actually require some administration intervention (including cleanup in our database)

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/repository/CallbackUrlEventRepository.java
```diff
@@ -0,0 +1,74 @@
...
+            SELECT c FROM CallbackUrlEventEntity c
+            WHERE c.status = io.getlime.security.powerauth.app.server.database.model.enumeration.CallbackUrlEventStatus.FAILED
+            AND c.timestampNextCall < :timestamp
+            """)
+    Stream<CallbackUrlEventEntity> findScheduledForRetry(LocalDateTime timestamp, Pageable pageable);
+
+    @Modifying
+    @Query("""
+            DELETE FROM CallbackUrlEventEntity c
+            WHERE c.timestampDeleteAfter < :timestamp
```
zcgandcomp, line 70: The deletion should be pageable - we need to limit (batch) the job otherwise we can run into the issues with long running transactions.

zcgandcomp (changes requested):

docs/Database-Structure.md
jnpsk: We could recommend to use a very large number. An alternative is to implement the logic based on some negative constant, such as `-1`.

docs/Database-Structure.md
```diff
@@ -639,3 +645,41 @@ CREATE TABLE pa_fido2_authenticator (
...
+
+<!-- begin database table pa_application_callback_event -->
+### Callback URL Events
+
+Table stores Callback URL Events to monitor processing of the callbacks.
+
+#### Schema
+
+```sql
+CREATE TABLE pa_application_callback_event (
```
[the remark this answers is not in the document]
jnpsk: We should probably delete all the others DDLs from the file in that case

docs/PowerAuth-Server-1.8.0.md
```diff
@@ -31,3 +31,38 @@ values supported by the WebAuthn protocol, serialized as a JSON array.
...
+- `timestamp_created` - Creation timestamp of the Callback URL Event.
+- `timestamp_last_call` - Timestamp of the last time the Callback URL Event was sent.
+- `timestamp_next_call` - Timestamp of the next scheduled time to send the Callback URL Event.
+- `timestamp_delete_after` - Timestamp after which the Callback URL Event record should be deleted from the table.
+- `attempts` - Number of dispatch attempts made for the Callback URL Event.
+
+The `pa_application_callback_event` table comes with following indices:
+- `pa_application_callback_event_status_timestamp_next_call_idx` on `(status, timestamp_next_call)`,
+- `pa_application_callback_event_timestamp_delete_after_idx` on `(timestamp_delete_after)`, and
+- `pa_application_callback_event_id_status_idx` on `(id, status)`.
```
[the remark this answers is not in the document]
jnpsk: You are right, that in case of having an index on `(status, timestamp_next_call)` it makes no benefit to have another on `(status)` only.
The index on `(id, status)` is for the query `SELECT * FROM pa_application_callback_event WHERE id = '<id>' AND status = 'PENDING';`. Thats done by the event listener to prevent unnecessary fetching and locking a record that is not waiting to be dispatched anymore.

docs/images/arch_db_structure.png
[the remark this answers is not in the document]
jnpsk: I will add that in a separate issue

docs/PowerAuth-Server-1.8.0.md
```diff
@@ -31,3 +31,38 @@ values supported by the WebAuthn protocol, serialized as a JSON array.
...
+- `callback_data` - Data payload of the Callback URL Event.
+- `status` - Current state of the Callback URL Event.
+- `timestamp_created` - Creation timestamp of the Callback URL Event.
+- `timestamp_last_call` - Timestamp of the last time the Callback URL Event was sent.
+- `timestamp_next_call` - Timestamp of the next scheduled time to send the Callback URL Event.
+- `timestamp_delete_after` - Timestamp after which the Callback URL Event record should be deleted from the table.
+- `attempts` - Number of dispatch attempts made for the Callback URL Event.
+
+The `pa_application_callback_event` table comes with following indices:
+- `pa_application_callback_event_status_timestamp_next_call_idx` on `(status, timestamp_next_call)`,
```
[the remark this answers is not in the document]
jnpsk: Doesn't time range filtering benefits from an index?

docs/PowerAuth-Server-1.8.0.md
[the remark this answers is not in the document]
zcgandcomp: id is primary key right? Then it is a unique value, so there is no need to have any additional index on PK

docs/db/changelog/changesets/powerauth-java-server/1.8.x/20240704-callback-event-table.xml
```diff
@@ -0,0 +1,109 @@
...
+
+    <changeSet id="7" logicalFilePath="powerauth-java-server/1.8.x/20240704-callback-event-table.xml" author="Jan Pesek">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="pa_application_callback_event_id_status_idx"/>
+            </not>
+        </preConditions>
+        <comment>Create a new index on pa_application_callback_event(id, status).</comment>
+        <createIndex tableName="pa_application_callback_event" indexName="pa_application_callback_event_id_status_idx">
+            <column name="id" />
```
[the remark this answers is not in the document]
jnpsk: this index was removed after further discussion

docs/PowerAuth-Server-1.8.0.md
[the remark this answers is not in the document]
zcgandcomp: Like, for this type of table and expected data  - I think the benefit of indexing are wipe out be need of additional index structure.

How many PENDING callback do you expect? with the granularity of the nextcall this will not drastically reduce the cost of the search. On the other hand DB will need to maintain index, the DB will need to select the proper algorithm,... When deleting records the index is also recalculated,...
Plus you are hitting larger part of index (because you are selecting older then something) - yes it could be faster, but again the maintenance....

So, If we expect hundreds of PENDING events (Or many very old PENDING records forcing to read far part of the table), yes, it will start paying off.  But from my POV it will not.



btw as far as I recall index should be constructed from the most distinctive column to the least - BTREE is still favourite selection algorithm not licenced by Oracle.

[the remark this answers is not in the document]
jnpsk: good point with the timestamp granularity, the index probably won't have a significant effect in this case then

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlEventEntity.java
```diff
@@ -0,0 +1,92 @@
...
+@Entity
+@Table(name = "pa_application_callback_event")
+@Getter @Setter
+public class CallbackUrlEventEntity implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 3438887028420848470L;
+
+    @Id
+    @Column(name = "id", updatable = false, nullable = false, length = 36)
```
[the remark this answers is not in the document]
jnpsk: you're right. Those parameters does not seem to have any effect.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -69,4 +72,32 @@ public void expireActivations() {
...
+    @Scheduled(fixedRateString = "${powerauth.service.scheduled.job.dispatchPendingCallbackUrlEvent:3000}")
+    @SchedulerLock(
+            name = "dispatchPendingCallbackUrlEvent",
+            lockAtLeastFor = "#{T(java.lang.Math).round(${powerauth.service.scheduled.job.dispatchPendingCallbackUrlEvent:3000} * 0.8)}")
+    public void dispatchPendingCallbackUrlEvents() {
+        LockAssert.assertLocked();
+        logger.debug("dispatchPendingCallbackUrlEvents");
+        callbackUrlEventService.dispatchPendingCallbackUrlEvents();
+    }
+
```
jnpsk, line 94: This is not an ideal solution as it brings another thread fighting for a PENDING events in the database. However, because of the skip locked queries, it should not cause an extensive blocking.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/AsyncConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+@Slf4j
+public class AsyncConfiguration {
+
+    @Bean
+    public Executor callbackUrlEventsThreadPoolExecutor(final PowerAuthCallbacksConfiguration powerAuthCallbacksConfiguration) {
+        final ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(powerAuthCallbacksConfiguration.getThreadPoolCoreSize());
+        executor.setQueueCapacity(powerAuthCallbacksConfiguration.getThreadPoolQueueCapacity());
+        executor.setMaxPoolSize(powerAuthCallbacksConfiguration.getThreadPoolMaxSize());
+        executor.setRejectedExecutionHandler((r, e) -> logger.debug("A Callback URL Event was rejected by the ThreadPoolTaskExecutor."));
```
banterCZ, line 45: 🤔 Is _debug_ level enough, do we need to know this information in the production environment too?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventListener.java
```diff
@@ -0,0 +1,54 @@
...
+public class CallbackUrlEventListener {
+
+    private CallbackUrlEventService callbackUrlEventService;
+
+    /**
+     * Listener to react immediately to a newly created Callback URL Event that should be instantly dispatched.
+     * @param callbackUrlEventEntity Callback URL Event to dispatch.
+     */
+    @Async("callbackUrlEventsThreadPoolExecutor")
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
banterCZ, line 48: From the architectural point of view, I would rather see this configuration of transactions at the service level.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventService.java
```diff
@@ -0,0 +1,336 @@
...
+        synchronized (restClientCacheLock) {
+            restClient = restClientCache.get(cacheKey);
+            if (restClient == null) {
+                logger.debug("REST client not found in cache, initializing new REST client, callback cache key: {}", cacheKey);
+                restClient = createRestClientAndStoreInCache(callbackUrlEntity);
+            } else {
+                logger.debug("REST client found in cache, callback cache key: {}", cacheKey);
+            }
+        }
+        return restClient;
```
banterCZ, line 255: Let's use _final_ modifier as much as possible.
```suggestion
        synchronized (restClientCacheLock) {
            final RestClient restClient = restClientCache.get(cacheKey);
            if (restClient == null) {
                logger.debug("REST client not found in cache, initializing new REST client, callback cache key: {}", cacheKey);
                return createRestClientAndStoreInCache(callbackUrlEntity);
            } else {
                logger.debug("REST client found in cache, callback cache key: {}", cacheKey);
                return restClient;
            }
        }
```

banterCZ (dismissed):
Looks good to me. @korbelm @zcgandcomp May we do performance testing from this branch before merging?

zcgandcomp (dismissed):
Looks OK

docs/Database-Structure.md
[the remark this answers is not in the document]
banterCZ: I have opened an issue to clean it up. #1610

romanstrobl (dismissed):
I suggest we run load tests on this branch before merging. Otherwise I approve the changes.

jnpsk:
Coming with a significant changes due to inability to use the used locking mechanism on MSSQL.

Now the approach works as follows:

- States of a callback event are: 
    - `INSTANT` -> {`PENDING` (on full Executor’s queue), `PROCESSING` (when dispatched), `COMPLETED` (on Callback’s maxAttempts <= 0)}
    - `PENDING` -> {`PROCESSING` (when dispatched), `COMPLETED` (on Callback’s maxAttempts <= 0)}
    - `PROCESSING` -> {`FAILED`, `COMPLETED`}
    - `FAILED` -> {`PROCESSING` (on retry)}
    - `COMPLETED` -> {}

On an activation or operation change a new `CallbackUrlEventEntity` **(callback event)** is created and initialised to state `INSTANT` during 
the main business transaction `TrB`. Only after successful commit of the `TrB`, the callback event is enqueued to be dispatched by via a ThreadPoolExecutor.  If the ThreadPoolExecutor is busy, it rejects to dispatch the callback event and so the callback event will be transitioned to the state `PENDING` in database. And so, after the queueing, the callback event  is either stored in database in state `PENDING`, or is stored in database in state `INSTANT` and is, at the same time submitted, to ThreadPoolExecutor.

- ThreadPoolExecutor and its queue is thread safe and so the callback event is processed once. There is no other consumer for `INSTANT` events.
- `PENDING` events are processed by a scheduled job with distributed locks, and so only one job processes `PENDING` events in the same time. Similarly for `FAILED` events.

banterCZ (approved):
In general, looks good. I have added several ideas to consider. Did it pass the performance tests? Let's wait for other reviews.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/CallbackUrlEventStatus.java
```diff
@@ -0,0 +1,55 @@
...
+ * Possible states of a Callback URL Event.
+ *
+ * @author Jan Pesek, jan.pesek@wultra.com
+ */
+public enum CallbackUrlEventStatus {
+
+    /**
+     * Status of a Callback URL Event that is currently being dispatched.
+     */
+    PROCESSING,
```
zcgandcomp, line 31: I do not find any value in this status. And the locking mechanism is missing entirely.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/CallbackUrlEventStatus.java
```diff
@@ -0,0 +1,55 @@
...
+     * Status of a newly created Callback URL Event that is waiting to be
+     * processed by the Executor.
+     */
+    INSTANT,
+
+    /**
+     * Status of a newly created Callback URL Event that is waiting to be
+     * processed by the scheduled task.
+     */
+    PENDING,
```
zcgandcomp, line 43: PENDING is in used logic the same as FAILED. So this seems to me as a redundant state

zcgandcomp (changes requested):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventService.java
```diff
@@ -0,0 +1,302 @@
...
+        if (!shouldBeProcessed(callbackUrlEventEntity)) {
+            logger.debug("Callback URL Event {} should not be processed, state={}", callbackUrlEventEntity.getId(), callbackUrlEventEntity.getStatus());
+            return;
+        }
+
+        callbackUrlEventEntity.setStatus(CallbackUrlEventStatus.PROCESSING);
+        callbackUrlEventEntity.setTimestampNextCall(null);
+        callbackUrlEventEntity.setTimestampLastCall(LocalDateTime.now());
+        callbackUrlEventEntity.setAttempts(callbackUrlEventEntity.getAttempts() + 1);
+        callbackUrlEventRepository.save(callbackUrlEventEntity);
```
romanstrobl, line 144: Can we use `callbackUrlEventEntity = callbackUrlEventRepository.save(callbackUrlEventEntity)` to avoid issues in the future?

docs/Database-Structure.md
romanstrobl: Why would we allow infinity repetition? It does not seem to be a good idea.

romanstrobl (approved):
I found several minor additional issues.

docs/db/changelog/changesets/powerauth-java-server/1.9.x/20240704-callback-event-table.xml
```diff
@@ -0,0 +1,95 @@
...
+        <comment>Create a new index on pa_application_callback_event(status).</comment>
+        <createIndex tableName="pa_application_callback_event" indexName="pa_application_callback_event_status_idx">
+            <column name="status" />
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="6" logicalFilePath="powerauth-java-server/1.9.x/20240704-callback-event-table.xml" author="Jan Pesek">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <indexExists indexName="pa_application_callback_event_timestamp_delete_after_idx"/>
```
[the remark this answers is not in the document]
zcgandcomp: I believe this was prior to 12.2 19c should have limit 128.  We need to test liquibase in any case it could still honor old 30 char limit.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventService.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+
+    /**
+     * Get default force rerun period, after which is a Callback URL Event in PROCESSING state considered stale.
+     * @return Default force rerun period.
+     */
+    private Duration defaultForceRerunPeriod() {
+        // This is an arbitrary value, representing allowed delay before trying establishing remote connection.
+        final Duration allowedProcessingDelay = Duration.ofSeconds(10);
```
zcgandcomp, line 177: We should use max (arbitrary value, backoff period). We need to be sure the legitimate time of the request processing passed, but we also should not process the request sooner than his backoff period in case of processed failure.

scenario:
- we execute the callback
- somebody reset the server before the standard answer
- reset stale job will update all events that have last call timestamp older than now - forcedRerun.
  - if the call was successful we do not mind to wait longer period - as the receiver was already informed.
  - if the call was not successful we should wait the same period i.e. add backoff interval.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -458,110 +459,60 @@ private Map<String, Object> prepareCallbackDataOperation(CallbackUrlEntity callb
...
+        final LocalDateTime timestampNow = LocalDateTime.now();
+
+        final CallbackUrlEventEntity callbackUrlEventEntity = new CallbackUrlEventEntity();
+        callbackUrlEventEntity.setCallbackUrlEntity(callbackUrlEntity);
+        callbackUrlEventEntity.setCallbackData(callbackData);
+        callbackUrlEventEntity.setIdempotencyKey(UUID.randomUUID().toString());
+        callbackUrlEventEntity.setTimestampCreated(timestampNow);
+        callbackUrlEventEntity.setTimestampLastCall(timestampNow);
+        callbackUrlEventEntity.setAttempts(0);
+        callbackUrlEventEntity.setStatus(CallbackUrlEventStatus.PROCESSING);
```
zcgandcomp, line 514: After some thought.
To protect the situation when callback receiver doesn't have idempotency - for whatever reasons) We should set the status as INSTANT (and change the respective part in the postCallback and cleaning jobs. 

In this particular case ,we do not need to store the event at all (but it might be handy as log supplement) but this might bring too big complexity

zcgandcomp (changes requested):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
jnpsk: Instead of introducing another state and increasing complexity, the rerun logic is now based on a new column `timestamp_rerun_after` which is set to `now + rerun period` when an event gets into PROCESSING state (i.e. when first created or when taken from DB as PEDNING).
Events, which are configured with max attempts = 1, does not get `timestamp_rerun_after` and therefore are not part of the reset job.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/callbacks/CallbackUrlEventService.java
jnpsk: I consider server restart / crash as a rare event, which should affect the callback processing as little as possible and  not consider it as a callback processing failure.
Do you suggest to behave to stale events in the same way as to failed events? i.e. incrementing attempts and set backoff?

zcgandcomp (approved):
Ok, this should do a trick. Thank you looks good to me.

petrdvorak (commented):
Hello @zcgandcomp @jnpsk, I like the change but would consider introducing jitter to the webhook invocation and limiting the queue size...

Before this change, failed callbacks did not trigger any follow-up behavior - they just failed. This changes with this PR. In case the webhook processing service fails (I assume this could be the main reason for a webhook failure?), we will start adding callbacks to the queue and the queue will quickly fill up. Then, we will start retrying at the same time - while we allow limiting the size of the scheduled call batch, we could still overwhelm the processing callback.

Limiting the queue size and adding a bit of randomness could help with this.

zcgandcomp (approved):
Looks OK

zcgandcomp (dismissed):
Looks OK

docs/sql/mssql/migration_1.8.0_1.9.0.sql
```diff
@@ -17,3 +17,49 @@ GO
...
 GO
+
+-- Changeset powerauth-java-server/1.9.x/20240704-callback-event-table.xml::1::Jan Pesek
+-- Create a new table pa_callback_event
+CREATE TABLE pa_application_callback_event (id bigint NOT NULL, application_callback_id varchar(37) NOT NULL, callback_data varchar (max) NOT NULL, status varchar(32) NOT NULL, timestamp_created datetime2(6) CONSTRAINT DF_pa_application_callback_event_timestamp_created DEFAULT GETDATE() NOT NULL, timestamp_last_call datetime2(6), timestamp_next_call datetime2(6), timestamp_delete_after datetime2(6), timestamp_rerun_after datetime2(6), attempts int CONSTRAINT DF_pa_application_callback_event_attempts DEFAULT 0 NOT NULL, idempotency_key varchar(36) NOT NULL, CONSTRAINT PK_PA_APPLICATION_CALLBACK_EVENT PRIMARY KEY (id), CONSTRAINT pa_application_callback_id_fk FOREIGN KEY (application_callback_id) REFERENCES pa_application_callback(id));
+GO
+
+-- Changeset powerauth-java-server/1.9.x/20240704-callback-event-table.xml::2::Jan Pesek
+-- Add max_attempts column to pa_application_callback table.
+ALTER TABLE pa_application_callback ADD max_attempts int;
```
romanstrobl, line 28: Why do we alter a table that was just created? We can just add these columns in the `CREATE TABLE`. The PR was not merged yet to `develop`, so it should be safe even I understand the table went through changes.

docs/sql/mssql/migration_1.8.0_1.9.0.sql
```diff
@@ -17,3 +17,49 @@ GO
...
+GO
+
+-- Changeset powerauth-java-server/1.9.x/20240704-callback-event-table.xml::3::Jan Pesek
+-- Add initial_backoff column to pa_application_callback table.
+ALTER TABLE pa_application_callback ADD initial_backoff varchar(64);
```
romanstrobl, line 33: dtto

docs/sql/mssql/migration_1.8.0_1.9.0.sql
```diff
@@ -17,3 +17,49 @@ GO
...
+GO
+
+-- Changeset powerauth-java-server/1.9.x/20240704-callback-event-table.xml::4::Jan Pesek
+-- Add retention_period column to pa_application_callback table.
+ALTER TABLE pa_application_callback ADD retention_period varchar(64);
```
romanstrobl, line 38: dtto

romanstrobl (dismissed):
Approved now. I have one minor comment related to unnecessary ALTER TABLE statements.

docs/sql/mssql/migration_1.8.0_1.9.0.sql
romanstrobl: Ah, I see ok.

romanstrobl (approved):

PR #1597, opened 2024-07-16
Fix #1532: Add lockAtLeastFor to the scheduler configuration

### LINKED ISSUE #1532: Add lockAtLeastFor to the scheduler configuration
As of now high number of instances caused high repetition of the job.

Set `lockAtLeastFor` attribute to prevent too often calls. 

TBD global config vs per scheduled operation - as the scheduled operations can have different interval the `lockAtLeastFor` should be parameterizable per operation rather than on global level.

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -46,23 +46,29 @@ public class CleaningTask {
     private final ActivationServiceBehavior activationServiceBehavior;
 
     @Scheduled(fixedRateString = "${powerauth.service.scheduled.job.uniqueValueCleanup:60000}")
-    @SchedulerLock(name = "expireUniqueValuesTask")
+    @SchedulerLock(
+            name = "expireUniqueValuesTask",
+            lockAtLeastFor = "#{T(java.lang.Math).round(${powerauth.service.scheduled.job.uniqueValueCleanup:60000} * 0.8)}")
```
banterCZ, line 51: We cannot introduce a global value. My idea was to avoid adding other properties and set the value to 80 percent of the particular property.

zcgandcomp (approved):
Looks Good. The idea of 80 pct for lock least is cool for initial setting.

jnpsk (approved):
looks ok!

PR #1599, opened 2024-07-18
Increase defaultLockAtMostFor to 60m

banterCZ:
It fixes error: lockAtLeastFor is longer than lockAtMostFor for lock 'evictFido2AuthenticatorCacheTask'

A follow-up to #1532


romanstrobl (approved):
Looks OK.

zcgandcomp (commented):
And we have jobs that we are expecting to run over 10 minutes?

banterCZ:
Fixed by #1601

PR #1601, opened 2024-07-18
Remove scheduler lock from Fido2CleaningTask

banterCZ:
It fixes the error: `lockAtLeastFor is longer than lockAtMostFor for lock 'evictFido2AuthenticatorCacheTask'`
Cleaning local cache, no need to lock the job across the whole cluster.

A follow-up to #1532


zcgandcomp (approved):
Local cache, needs to run on all instances

jnpsk (approved):
nice catch

petrdvorak (approved):
Looks OK

PR #1606, opened 2024-07-22
Fix #1523: Data escaping in operation template

### LINKED ISSUE #1523: Data escaping in operation template
In the demo app when I enter into any field following character * it will simply cut off the rest of the value

f.e. I send 55464*5464!@#$% -> but on the token only this is displayed 55464

banterCZ:


romanstrobl (approved):
Perfect.

PR #1608, opened 2024-07-23
Fix #1607: Extract common code into EncryptionService

### LINKED ISSUE #1607: Extract common code into EncryptionService
There is duplicated code for encryption in `RecoveryPrivateKeyConverter`, `RecoveryPukConverter`, and `ServerPrivateKeyConverter`.

This relates to #1588

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/Encryptable.java
```diff
@@ -0,0 +1,46 @@
...
+package io.getlime.security.powerauth.app.server.database.model;
+
+import io.getlime.security.powerauth.app.server.database.model.enumeration.EncryptionMode;
+
+/**
+ * A generic wrapper for encryption, keeping both the mode and the data.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public interface Encryptable {
```
jnpsk, line 27: Just thinking if a base class, from which the other 3 would be extended, isn't more suitable approach in this case, instead of the "base record" in the service.

jnpsk (dismissed):
Looks ok, nice refactoring.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/Encryptable.java
banterCZ: Thanks for the opinion. Inheritance is very limiting. Moreover, I will need this interface for `ApplicationConfigEntity`.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/converter/RecoveryPukConverter.java
```diff
@@ -75,77 +51,8 @@ public RecoveryPukConverter(PowerAuthServiceConfiguration powerAuthServiceConfig
...
-                }
-            }
-            default -> {
-                logger.error("Unknown key encryption mode: {}", encryptionMode.getValue());
-                // Rollback is not required, error occurs before writing to database
-                throw localizationProvider.buildExceptionForCode(ServiceError.UNSUPPORTED_ENCRYPTION_MODE);
-            }
-        }
+    public String fromDBValue(final RecoveryPuk pukHash, final long applicationRid, final String userId, final String recoveryCode, final long pukIndex) throws GenericServiceException {
+        return encryptionService.fromDBValue(pukHash, createSecretKeyDerivations(applicationRid, userId, recoveryCode, pukIndex));
```
romanstrobl, line 55: Why is the method name plural in `createSecretKeyDerivations`?

I suggest to use `deriveSecretKey`.

romanstrobl (dismissed):
Looks OK, just one comment related to method naming.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/converter/RecoveryPukConverter.java
banterCZ: 🤔 Since the result is a type of `List`, let's rename that to `createSecretKeyDerivationInputs`, because deviation itself is happening in a different place.

jnpsk (approved):

PR #1617, opened 2024-07-25
Fix #1592: Implement procedure sp_getapplock for locking activations on MSSQL (backport)

### LINKED ISSUE #1592: Implement procedure sp_getapplock for locking activations on MSSQL
In order to prevent deadlocks due to pesimmistic locking of activations and operations on MSSQL, we should implement the `sp_getapplock` to lock these database resources in a more reliable way on MSSQL.

banterCZ:
(cherry picked from commit cafed8cd37fb6503fa9e298cffe88e63f52afa7d)

### DIFF
--- a/docs/Deploying-PowerAuth-Server.md
+++ b/docs/Deploying-PowerAuth-Server.md
@@ -8,8 +8,7 @@ Following databases are supported:
 
 - Oracle Database 11g, 12c, 19c, or 21c or
 - PostgreSQL 9.5.4 or newer, or
-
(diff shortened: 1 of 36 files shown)


banterCZ:
This is quite risky, see #1620 for a quick fix.

romanstrobl (approved):
This fix is risky because of the high number of changes. I will approve it so it can be used, however the fix should be tested thoroughly before putting into a release.

banterCZ:
Not needed anymore, introduced in 1.8.x

PR #1621, opened 2024-07-25
Avoid deadlock during expiration of operations

romanstrobl:


petrdvorak (dismissed):

zcgandcomp (approved):
Eventual expire of operation should not break the business execution - this is OK .
saveAndFlush will increase communication with DB, but usually is's happening just before end of transaction so OK.
There could be potentially inconsistency when operation is efter expiration time, but we are not setting status in DB till the next expiration job.

banterCZ (approved):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1196,13 +1187,13 @@ public void expireOperations() {
...
         final PageRequest pageRequest = PageRequest.of(0, powerAuthServiceConfiguration.getExpireOperationsLimit());
         try (final Stream<OperationEntity> pendingOperations = operationQueryService.findExpiredPendingOperations(currentTimestamp, pageRequest)) {
-            pendingOperations.forEach(op -> {
-                try {
-                    expireOperation(op, currentTimestamp);
-                } catch (GenericServiceException e) {
-                    logger.error("Operation expiration failed, operation ID: {}", op.getId());
-                }
-            });
+            final List<OperationEntity> updatedEntities = pendingOperations.peek(operationEntity -> {
```
banterCZ, line 1190: 👀 `peek` exists mainly to support debugging

banterCZ (approved):

jnpsk (approved):

PR #1628, opened 2024-07-28
Fix #1627: Update deadlocks on MSSQL

### LINKED ISSUE #1627: Update deadlocks on MSSQL
Following deadlocks occur due to concurrent updates of activation data in table `pa_activation`.

The problem occurs due to conflict on the indexes related to the table.

[deadlocks_update.txt](https://github.com/user-attachments/files/16403133/deadlocks_update.txt)

romanstrobl:
This fix moves the snapshot isolation for MSSQL before each transaction. The original fix did was bound only to current database session.


petrdvorak (approved):
Looks OK... MSSQL still has some way to go...

zcgandcomp (approved):
Looks OK.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/MssqlIsolationConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+        return new BeanPostProcessor() {
+            @Override
+            public Object postProcessBeforeInitialization(@NonNull Object bean, @NonNull String beanName) throws BeansException {
+                return bean;
+            }
+
+            @Override
+            public Object postProcessAfterInitialization(@NonNull Object bean, @NonNull String beanName) throws BeansException {
+                if (bean instanceof HikariDataSource hikariDataSource) {
+                    logger.debug("Setting initialization SQL: SET TRANSACTION ISOLATION LEVEL SNAPSHOT");
```
banterCZ, line 55: 🤔 Could it be INFO level? I think that it is an important information.

banterCZ (approved):
Great!

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/MssqlIsolationConfiguration.java
romanstrobl: Yes, I can switch to INFO.

banterCZ (approved):
Thanks for the INFO level.

PR #1637, opened 2024-07-31
Extend tests for encryption

banterCZ:
An important step before the refactor

Related to #1607


jnpsk (approved):
looks ok

romanstrobl (approved):
Thanks for the tests!

Issue #1603, opened 2024-07-18
Schema diagram how to


jnpsk:
Add steps to generate schema diagram to `Developer-How-To-Start.md`

PR #1639, opened 2024-08-01
Generate initial db diagram with SchemaCrawler

banterCZ:
I would like to have an initial SchemaCrawler db diagram, for better tracking of future changes. 

A follow-up to #1603


romanstrobl (approved):
Great!

jnpsk (approved):
👍

PR #1643, opened 2024-08-05
OIDC: Allow configuration key oauth2_providers #

banterCZ:


romanstrobl (dismissed):
Approved, but we should update documentation.

PR #1645, opened 2024-08-08
Fix #1644: Investigate error on Base64 decoding in challenge

### LINKED ISSUE #1644: Investigate error on Base64 decoding in challenge
We noticed the following error:

```
Illegal base64 character 2d (through reference chain: com.wultra.powerauth.fido2.rest.model.entity.CollectedClientData["challenge"])
```

We only saw it once when using `CZFlotantă žluťoučký koníček` as IBAN value, and we are now unable to reproduce the issue with the data.

jnpsk:
Challenge is encoded to Base64Url on the browser level. The encoded challenge is then encapsulated to the [Collected Client Data structure](https://www.w3.org/TR/webauthn-3/#dictionary-client-data), which hash is later signed by the authenticator. Base64Url decoding should be applied on the challenge during the CollectedClientData deserialization.

### DIFF
--- a/powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/Base64UrlToStringDeserializer.java
+++ b/powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/Base64UrlToStringDeserializer.java
@@ -34,7 +34,7 @@ import java.util.Base64;
  * @author Petr Dvorak, petr@wultra.com
  */
 @Slf4j
-public class Base64ToStringDeserializer extends StdDeserializer<String> {
+public class Base64UrlToStringDeserializer extends StdDeserializer<String> {
 
     @Serial
     private static final long serialVersionUID = 2540966716709142276L;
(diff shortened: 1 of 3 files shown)


powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/Base64UrlToStringDeserializer.java
```diff
@@ -64,7 +64,7 @@ public Base64ToStringDeserializer(Class<?> vc) {
     @Override
     public String deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws Fido2DeserializationException {
         try {
-            return new String(Base64.getDecoder().decode(jsonParser.getText()), StandardCharsets.UTF_8);
+            return new String(Base64.getUrlDecoder().decode(jsonParser.getText()), StandardCharsets.UTF_8);
```
banterCZ, line 67: 👍

banterCZ (approved):
Looks good.

zcgandcomp (approved):
Looks OK.

PR #1666, opened 2024-09-06
Fix #1659: Simplify JOSE ECDSA signature verification

### LINKED ISSUE #1659: Simplify JOSE ECDSA signature verification
The mobile SDK is providing method `powerAuthSdk.signJwtWithDevicePricateKey()` that calculate JWT signed with ES256 with using device's private key. Unfortunately, there's no simple way how to verify such signature on PowerAuth Server. The problem is, that JWT is using signatures in JOSE, instead of more common DER format, supported on our server.

I think that we should simply enhance `/rest/v3/signature/ecdsa/verify` endpoint with optional "signatureFormat" parameter, that will specify the format of the provided signature.

petrdvorak:


hvge (approved):
Looks OK.

PR #1669, opened 2024-09-09
Fix #1533: Callback OAUTH2.x authorization

### LINKED ISSUE #1533: Callback OAUTH2.x authorization
Add capability to protect callbacks via OAUTH2.x. 

Support grant_type`client_credentials`.

Encrypt the authentication column with the general encryption key.

Derive the specific key (e.g. from application id).

Update the API to support the OAUTH authorizaton as the API will be only option how to setup callbacks correctly (with authentication).

Attributes, if the authentication column is encrypted, are required; otherwise, we will face migration issues.

banterCZ:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlAuthentication.java
```diff
@@ -0,0 +1,107 @@
...
+package io.getlime.security.powerauth.app.server.database.model.entity;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.io.Serial;
+import java.io.Serializable;
+
+/**
+ * Entity class for storing callback authentication credentials in database.
```
romanstrobl, line 28: Is this class really an entity class?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -555,13 +559,62 @@ private RestClient initializeRestClient(CallbackUrlEntity callbackUrlEntity) thr
...
+        final CallbackUrlAuthentication.HttpBasic httpBasicAuth = authentication.getHttpBasic();
         if (httpBasicAuth != null && httpBasicAuth.isEnabled()) {
             builder.httpBasicAuth()
                     .username(httpBasicAuth.getUsername())
                     .password(httpBasicAuth.getPassword());
         }
+
+        final CallbackUrlAuthentication.OAuth2 oAuth2Config = authentication.getOAuth2();
+        if (oAuth2Config != null && oAuth2Config.isEnabled()) {
+            builder.filter(configureO2AuthExchangeFilter(oAuth2Config, callbackUrlEntity.getId()));
```
romanstrobl, line 571: Typo: `O2Auth`

romanstrobl (approved):
Looks OK, just 2 minor issues.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/entity/CallbackUrlAuthentication.java
banterCZ: No, it is not. I was about to change it but forgot. Thanks!

romanstrobl (approved):
OK now.

PR #1671, opened 2024-09-11
Fix #1670: Add commit phase to allow activation commit on key exchange

### LINKED ISSUE #1670: Add commit phase to allow activation commit on key exchange
Add a `commitPhase` parameter into `InitActivationRequest` to allow commit on key exchange or in `PENDING_COMMIT` state (default). Deprecate the original `activationOtpValidation` parameter, whose value can be determined from request data. Keep compatibility for existing scenarios with OTP.

romanstrobl:


docs/db/changelog/changesets/powerauth-java-server/1.9.x/20240910-commit-phase.xml
```diff
@@ -0,0 +1,18 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
+
+    <changeSet id="1" logicalFilePath="powerauth-java-server/1.9.x/20240910-commit-phase.xml" author="Roman Strobl">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="pa_activation" columnName="commit_phase"/>
+            </not>
+        </preConditions>
+        <comment>Add commit_phase column to pa_activation table.</comment>
+        <addColumn tableName="pa_activation">
```
zcgandcomp, line 13: Joy of MSSQL. If we need to have this for old records (can be the case) we need to use either update after the column is added, or specific add of the column for MSSQL:

`ALTER TABLE pa_activation ADD protocol commit_phase integer CONSTRAINT DF_pa_activation_commit_phase DEFAULT 0 WITH VALUES;`

If we do not need to care about already created records we can omit this.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ActivationController.java
```diff
@@ -214,38 +214,29 @@ public String activationDetail(
...
+     * @param commitPhase             Specifies when activation is committed.
      * @param activationOtp           Activation OTP code.
      * @param model                   Model with passed parameters.
      * @param redirectAttributes      Redirect attributes.
      * @return Redirect the user to activation detail.
      */
     @GetMapping("/activation/create")
     public String activationCreate(@RequestParam("applicationId") String applicationId, @RequestParam("userId") String userId,
-                                   @RequestParam("activationOtpValidation") String activationOtpValidation,
+                                   @RequestParam("commitPhase") String commitPhase,
```
banterCZ, line 225: 🤔 Should it be `required=false`? Or are we sure it is safe because it is called by our admin GUI?

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
```diff
@@ -146,13 +146,13 @@ public interface PowerAuthClient {
...
      * @param timestampActivationExpire Timestamp until when the activation can be committed.
-     * @param otpValidation             Mode that determines in which stage of activation should be additional OTP validated.
+     * @param commitPhase               Specifies when the activation is committed.
      * @param otp                       Additional OTP value.
      * @return {@link InitActivationResponse}
      * @throws PowerAuthClientException In case REST API call fails.
      */
     InitActivationResponse initActivation(String userId, String applicationId, Long maxFailureCount, Date timestampActivationExpire,
-                                          ActivationOtpValidation otpValidation, String otp) throws PowerAuthClientException;
+                                          CommitPhase commitPhase, String otp) throws PowerAuthClientException;
```
banterCZ, line 155: Ok, but let's review this approach later on. From the binary point of view, this is a trap and we should prefer methods with parameter objects such as `#initActivation(InitActivationRequest)` and deprecate these multi-parameter ones.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/enumeration/CommitPhase.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Enum representing phase when activation is commited.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public enum CommitPhase {
+    /**
+     *
+     * Activation is commited in PENDING_COMMIT state (default).
```
banterCZ, line 30: An extra line.
```suggestion
     * Activation is commited in PENDING_COMMIT state (default).
```

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/InitActivationRequest.java
```diff
@@ -40,7 +41,9 @@ public class InitActivationRequest {
     private String applicationId;
     private Date timestampActivationExpire;
     private Long maxFailureCount;
+    @Deprecated
```
banterCZ, line 44: Maybe we may add Javadoc, something like
```suggestion
    /**
    * @deprecated use {@link #commitPhase} instead
    **/
    @Deprecated
```

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/converter/ActivationCommitPhaseConverter.java
```diff
@@ -0,0 +1,59 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package io.getlime.security.powerauth.app.server.converter;
+
+import com.wultra.security.powerauth.client.model.enumeration.CommitPhase;
+
+/**
+ * Converter class between {@link CommitPhase} and
+ * {@link io.getlime.security.powerauth.app.server.database.model.enumeration.CommitPhase}.
```
banterCZ, line 25: Missing author.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ActivationCommitPhaseConverter.java
```diff
@@ -0,0 +1,71 @@
...
+import jakarta.persistence.Converter;
+import org.springframework.stereotype.Component;
+
+/**
+ * Converter between {@link CommitPhase} enumeration and integer values.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Converter
+@Component
```
banterCZ, line 33: Nitpick, it does not have to be a Spring bean, does it?

docs/db/changelog/changesets/powerauth-java-server/1.9.x/20240910-commit-phase.xml
```diff
@@ -0,0 +1,18 @@
...
+            <column name="commit_phase" type="integer" defaultValue="0"/>
```
banterCZ, line 14: Ok, already discussed in TLDR that it will be an enum represented as a number and not a string.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/ActivationOtpValidation.java
```diff
@@ -22,8 +22,8 @@
  * Enum representing possible activation OTP validation stages. Following values are supported:
  * <p>
  * - NONE = 0
- * - KEYS_EXCHANGE = 1
- * - COMMIT = 2
+ * - ON_KEY_EXCHANGE = 1
+ * - ON_COMMIT = 2
```
banterCZ, line 26: 🤔 In `CommitPhase.java`, there is `ON_COMMIT=0` https://github.com/wultra/powerauth-server/pull/1671/files?short_path=22681d7#diff-2e2152b1c4d5a479a3e8582131e414cc563391841b61655c12473a169495aa81R24
I realize, that it is a different enum, but at first sight, it is confusing.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ActivationCommitPhaseConverter.java
```diff
@@ -0,0 +1,71 @@
...
+     * @return {@link CommitPhase} enumeration.
+     */
+    @Override
+    public CommitPhase convertToEntityAttribute(Integer value) {
+        if (value == null) {
+            // For compatibility with old data
+            return CommitPhase.ON_COMMIT;
+        }
+        return switch (value) {
+            case 0 -> CommitPhase.ON_COMMIT;
```
banterCZ, line 65: 🤔 Shouldn't be the conversion based on `CommitPhase#value`?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1534,7 +1540,15 @@ public CommitActivationResponse commitActivation(CommitActivationRequest request
             }
 
             // Validate activation OTP
-            validateActivationOtp(com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.ON_COMMIT, activationOtp, activation, externalUserId);
+            validateActivationOtp(activationOtp, activation, externalUserId);
+
+            // Check the commit phase
+            System.out.println("PHASE: " + activation.getCommitPhase());
```
banterCZ, line 1546: ⚠️ `System.out.println`

banterCZ (commented):
In general, looks good. Please use logger instead of `System.out` and update the database diagram. Other comments are minor.

docs/db/changelog/changesets/powerauth-java-server/1.9.x/20240910-commit-phase.xml
romanstrobl: We can keep the `null` values for existing rows, this case is supported.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/enumeration/ActivationOtpValidation.java
romanstrobl: Yes, it is different.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Removed, sorry.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ActivationController.java
romanstrobl: The controller is used only by admin GUI which specifies the parameter.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
zcgandcomp, line 1556: System.out.println - use logger.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1534,7 +1540,15 @@ public CommitActivationResponse commitActivation(CommitActivationRequest request
             }
 
             // Validate activation OTP
-            validateActivationOtp(com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.ON_COMMIT, activationOtp, activation, externalUserId);
+            validateActivationOtp(activationOtp, activation, externalUserId);
```
zcgandcomp, line 1553: empty activationOtp is validated as valid (similar to in the previous phase).

zcgandcomp (changes requested):
I think the validation requires more checks. 

Also compatibility of API should be addressed.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
```diff
@@ -120,12 +120,12 @@ public interface PowerAuthClient {
...
      * @param userId        User ID for which a new CREATED activation should be created.
      * @param applicationId Application ID for which a new CREATED activation should be created.
-     * @param otpValidation Mode that determines in which stage of activation should be additional OTP validated.
+     * @param commitPhase   Specifies when the activation is committed.
      * @param otp           Additional OTP value.
      * @return {@link InitActivationResponse}
      * @throws PowerAuthClientException In case REST API call fails.
      */
-    InitActivationResponse initActivation(String userId, String applicationId, ActivationOtpValidation otpValidation, String otp) throws PowerAuthClientException;
+    InitActivationResponse initActivation(String userId, String applicationId, CommitPhase commitPhase, String otp) throws PowerAuthClientException;
```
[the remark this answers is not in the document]
romanstrobl: Yes, is this a problem? We want to remove the activationOtpValidation parameter, so this is a gradual migration to the new parameter.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -982,6 +981,15 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
         }
     }
 
+    private void validateOtpValidationAndCommitPhase(com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation activationOtpValidation, com.wultra.security.powerauth.client.model.enumeration.CommitPhase commitPhase) throws GenericServiceException {
+        // Validate combination of activation OTP and OTP validation mode.
+        if (activationOtpValidation != com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.NONE && commitPhase != null) {
```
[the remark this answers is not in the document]
romanstrobl: I was planning to drop the logic in next release, after the parameter cannot be used.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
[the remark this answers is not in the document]
zcgandcomp: Well basically in this case it is very fast gradual. And if we remove the otpValidation from the API we should remove it from most of the processing logic.

Still there is an impact for anybody using powerauth client.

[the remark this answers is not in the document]
romanstrobl: Ok, we can introduce new methods for the new logic to make the change compatible.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1701,52 +1715,22 @@ public UpdateActivationOtpResponse updateActivationOtp(UpdateActivationOtpReques
...
-        // Check whether this is validation in the different step of activation. If yes, then the confirmation
-        // OTP must not be provided.
-        if (expectedStage != currentStage) {
-            if (confirmationOtp != null) {
-                logger.info("Activation OTP is not expected, but is provided: {}", activationId);
-                // Rollback is not required, database is not used for writing yet.
-                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_ACTIVATION_OTP);
-            }
+        // Check whether activation OTP is specified.
+        if (!StringUtils.hasText(confirmationOtp)) {
```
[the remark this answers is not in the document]
romanstrobl: Ok, I will change the logic.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ActivationCommitPhaseConverter.java
romanstrobl: Removed `@Component`

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ActivationCommitPhaseConverter.java
romanstrobl: I used similar logic to: https://github.com/wultra/powerauth-server/blob/6e8fe90fb38101d10ca2604e7889a21aa24f522a/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/database/model/converter/ActivationOtpValidationConverter.java#L53

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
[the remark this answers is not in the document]
romanstrobl: See: https://github.com/wultra/powerauth-server/pull/1671/commits/495f1302182cbd04bfe58ea043d92df0b8c8583d

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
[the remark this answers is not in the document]
romanstrobl: This issue is already addressed in a separate commit: See: https://github.com/wultra/powerauth-server/pull/1671/commits/495f1302182cbd04bfe58ea043d92df0b8c8583d

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1319,6 +1324,7 @@ public CreateActivationResponse createActivation(CreateActivationRequest request
             initRequest.setTimestampActivationExpire(activationExpireTimestamp);
             initRequest.setActivationOtp(activationOtp);
             initRequest.setActivationOtpValidation(activationOtpValidation);
+            initRequest.setCommitPhase(com.wultra.security.powerauth.client.model.enumeration.CommitPhase.ON_COMMIT);
```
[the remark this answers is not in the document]
romanstrobl: Indeed, this can happen and conflicting values are checked here:
https://github.com/wultra/powerauth-server/pull/1671/files#diff-a4ce67a9fd8bc9dfacab0302fa5f1cb1a44ee59e9ceafb746ef9e8df877d7492R987

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: See:
https://github.com/wultra/powerauth-server/pull/1671/commits/495f1302182cbd04bfe58ea043d92df0b8c8583d

docs/WebServices-Methods.md
```diff
@@ -443,14 +443,15 @@ REST endpoint: `POST /rest/v3/activation/init`
...
+| `DateTime`                | `timestampActivationExpire` | Timestamp after when the activation cannot be completed anymore                                                                                                                                                                                       |
+| `Long`                    | `maxFailureCount`           | How many failures are allowed for this activation                                                                                                                                                                                                     |
+| `ActivationOtpValidation` | `activationOtpValidation`   | *Deprecated* optional activation OTP validation mode, use the `activationOtp` parameter during activation init or activation commit to control the OTP check. Use the `commitPhase` parameter for specifying when the activation should be committed. |
```
banterCZ, line 452: You may even use :warning: `:warning:` if you want to.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ActivationController.java
```diff
@@ -214,38 +214,29 @@ public String activationDetail(
...
+            switch (commitPhase) {
                 case "ON_COMMIT" ->
-                        response = client.initActivation(userId, applicationId, ActivationOtpValidation.ON_COMMIT, activationOtp);
-                default -> {
-                    redirectAttributes.addFlashAttribute("error", "Invalid OTP validation mode.");
-                    return "redirect:/activation/list?userId=" + userId;
-                }
+                        response = client.initActivation(userId, applicationId, CommitPhase.ON_COMMIT, activationOtp);
+                case "ON_KEY_EXCHANGE" ->
+                        response = client.initActivation(userId, applicationId, CommitPhase.ON_KEY_EXCHANGE, activationOtp);
```
banterCZ, line 235: Nitpick, but you may write
```java
final InitActivationResponse response = switch (commitPhase) {
                case "ON_COMMIT" -> client.initActivation(userId, applicationId, CommitPhase.ON_COMMIT, activationOtp);
                case "ON_KEY_EXCHANGE" -> client.initActivation(userId, applicationId, CommitPhase.ON_KEY_EXCHANGE, activationOtp);
                default -> client.initActivation(userId, applicationId, CommitPhase.ON_COMMIT, activationOtp);
            };
```

banterCZ (approved):
Thanks for adding tests and improving Javadoc.

banterCZ (approved):

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1701,52 +1715,61 @@ public UpdateActivationOtpResponse updateActivationOtp(UpdateActivationOtpReques
...
+            }
+        } else {
+            // Validation of commit phase:
+            // - if no OTP is specified for an activation, no validation is required
+            // - check each stage to make sure OTP is available when required
+            // - in case the commit is done in the next stage, skip validation
+            if (expectedOtpHash == null) {
+                return;
+            }
+            if (currentStage == ActivationOtpValidation.ON_KEY_EXCHANGE && commitPhase == CommitPhase.ON_KEY_EXCHANGE && !StringUtils.hasText(confirmationOtp)) {
```
zcgandcomp, line 1756: Like the description. I suggest:
- Change currentStage from ActivationOtpValidation to CommitPhase enum (this is to prepare for the future when activationOtpValidation is going to be dropped).
- Change the name of variable `commitPhase` to `expectedCommitPhase` (to distinguish what is expected by activation setting and what is the phase of the control)
This will increase the clarity of the code.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1701,52 +1715,61 @@ public UpdateActivationOtpResponse updateActivationOtp(UpdateActivationOtpReques
...
-            logger.info("Activation OTP is expected, but is missing: {}", activationId);
-            // Rollback is not required, database is not used for writing yet.
-            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_ACTIVATION_OTP);
+            if (currentStage == ActivationOtpValidation.ON_COMMIT && commitPhase == CommitPhase.ON_COMMIT && !StringUtils.hasText(confirmationOtp)) {
+                logger.info("Activation OTP is missing for activation ID: {}, stage: {}", activationId, currentStage);
+                // Rollback is not required, database is not used for writing yet.
+                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_ACTIVATION_OTP);
+            }
+            if (currentStage == ActivationOtpValidation.ON_KEY_EXCHANGE && commitPhase == CommitPhase.ON_COMMIT) {
+                // Validation is done after key exchange, skip it for now
```
zcgandcomp, line 1767: We can actually do an analogic check if somebody provides `confirmationOtp` in the on key exchange and the expected phase is on commit. But it's not critical

zcgandcomp (dismissed):
Two minor comments. Can be approved as is - decide if we polish this or not.

romanstrobl:
I will improve it, I already considered switching to CommitPhase for the stage.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -841,18 +843,14 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
...
-            final boolean hasActivationOtp = activationOtp != null && !activationOtp.isEmpty();
             if (activationOtpValidation == null) {
                 activationOtpValidation = com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.NONE;
             }
-            if ((activationOtpValidation == com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.NONE && hasActivationOtp) ||
-                    (activationOtpValidation != com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.NONE && !hasActivationOtp)) {
-                logger.warn("Activation OTP doesn't match its validation mode.");
-                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
-            }
+
```
zcgandcomp, line 849: And we remove the validation when activationOtpValidation is not NONE but the activation OTP is not provided.
No impact on functionality, but we can add validation, but this can clash with the conditions validation OTP later on.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Yes, I can add this case.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Good idea.

zcgandcomp (approved):
Nice job.

PR #1673, opened 2024-09-11
Fix #1672: Prevent creating activation scoped temporary key pair for FIDO2 activations

### LINKED ISSUE #1672: Prevent creating activation scoped temporary key pair for FIDO2 activations
While creating a temporary encryption key pair for FIDO2 activations does not hurt anything, there is nothing such a key could be used...

petrdvorak:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/TemporaryKeyBehavior.java
```diff
@@ -278,21 +279,23 @@ private TemporaryKeyResult obtainTemporaryKeyResult(TemporaryPublicKeyRequestCla
                 return result;
             } else {
 
-                final String appId = applicationVersionEntity.getApplication().getId();
+                final Long appId = applicationVersionEntity.getApplication().getRid();
```
banterCZ, line 282: 🤔 Is this change needed (the `id` should be also unique)? If so, I suggest renaming the variable, because it is confusing now (`id` vs `rid`).

banterCZ (approved):
The branch for FIDO2 ActivationProtocol looks good.

romanstrobl (approved):
Approved.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/TemporaryKeyBehavior.java
petrdvorak: We can refactor it... I used `rid` here because the comparison could be slightly faster...

PR #1685, opened 2024-09-30
Fix #1684: Update Client and Admin Console

### LINKED ISSUE #1684: Update Client and Admin Console
Issue https://github.com/wultra/powerauth-server/issues/1596 was solved by updating Callback Management API. The new request attributes should be modifiable from within the PA Client and from PA Admin Console.

jnpsk:
Update PA Admin Console and PA Client to reflect changes in Callback Management API (3 new attributes).

### DIFF
--- a/powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
+++ b/powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
@@ -34,7 +34,10 @@ import org.springframework.web.servlet.mvc.support.RedirectAttributes;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.time.Duration;
+import java.time.format.DateTimeParseException;
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 /**
@@ -171,7 +174,7 @@ public class ApplicationController {
      * @param model Model with passed parameters.
      * @return "callbackUpdate" view.
      */
-    @PostMapping("/application/detail/{applicationId}/callback/update")
+    @GetMapping("/application/detail/{applicationId}/callback/update")
     public String applicationUpdateCallback(@PathVariable("applicationId") String applicationId,
                                             @RequestParam String callbackId,
                                             Map<String, Object> model) {
@@ -226,6 +229,10 @@ public class ApplicationController {
                             model.put("auth_oAuth2Scope", oAuth2.getScope());
                             model.put("auth_oAuth2TokenUri", oAuth2.getTokenUri());
                         }
+
+                        model.put("retentionPeriod", callback.getRetentionPeriod());
+                        model.put("initialBackoff", callback.getInitialBackoff());
+                        model.put("maxAttempts", callback.getMaxAttempts());
                     }
                     return "callbackUpdate";
                 }
@@ -353,6 +360,19 @@ public class ApplicationController {
             if (errorAuth != null) {
                 error = errorAuth;
             }
+
+            if (!isValidDurationOrNull(allParams.get("retentionPeriod"))) {
+                error = "Retention period must be in ISO 8601 Duration format.";
+            }
+
+            if (!isValidDurationOrNull(allParams.get("initialBackoff"))) {
+                error = "Initial backoff must be in ISO 8601 Duration format.";
+            }
+
+            if (!isPositiveIntegerOrNull(allParams.get("maxAttempts"))) {
+                error = "Max attempts must be a positive integer.";
+            }
+
             if (error != null) {
                 for (String attribute: CALLBACK_ATTRIBUTES_OPTIONAL) {
                     if (allParams.get(attribute) != null) {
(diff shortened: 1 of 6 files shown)


powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
```diff
@@ -171,7 +174,7 @@ public String applicationCreateCallback(@PathVariable("applicationId") String id
      * @param model Model with passed parameters.
      * @return "callbackUpdate" view.
      */
-    @PostMapping("/application/detail/{applicationId}/callback/update")
+    @GetMapping("/application/detail/{applicationId}/callback/update")
```
romanstrobl, line 177: Why are we changing a `POST` method to `GET`? In REST, `GET` endpoints should not be used to modify data. Please clarify.

romanstrobl (changes requested):
In general looks OK, but in my opinion we should not use a `GET` REST API method for modifying data.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
banterCZ: Ok, I would like to hear the reason too. But my opinion is, that the JavaDoc says
> Show application callback update form

So the the path is misleading from the very beginning.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
```diff
@@ -890,16 +894,20 @@ public UpdateCallbackUrlResponse updateCallbackUrl(UpdateCallbackUrlRequest requ
     }
 
     @Override
-    public UpdateCallbackUrlResponse updateCallbackUrl(String id, String applicationId, String name, String callbackUrl, List<String> attributes, HttpAuthenticationPrivate authentication) throws PowerAuthClientException {
+    public UpdateCallbackUrlResponse updateCallbackUrl(String id, String applicationId, String name, CallbackUrlType type, String callbackUrl, List<String> attributes, HttpAuthenticationPrivate authentication, Duration retentionPeriod, Duration initialBackoff, Integer maxAttempts) throws PowerAuthClientException {
         final UpdateCallbackUrlRequest request = new UpdateCallbackUrlRequest();
         request.setId(id);
         request.setApplicationId(applicationId);
         request.setName(name);
+        request.setType(type.toString());
```
banterCZ, line 902: 🤔 Why is `type` a type of String instead of Enum? I know, we should solve this in a separate PR if agreed.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
jnpsk: If you take a look at the `ApplicationController`, there are few `GET` methods, whose API path suggest data modification, however those are only used to show forms in the browser. The other `POST` methods actually request changes on PowerAuth Server using Client. Their API path usually ends with `/do.submit`.

As @banterCZ said, it is misleading naming. Before this change it was not possible to show invalid input message on the callback update form, as the redirect `return "redirect:/application/detail/" + applicationId + "/callback/create";` expects the `GET` method on that endpoint.

romanstrobl: Ok, let's change the naming then.

jnpsk: Ok, I renamed the API paths to get update or create callback form. But misleading naming is used on many more endpoints in other controllers. I would not include it in this issue.

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
```diff
@@ -157,7 +160,7 @@ public String applicationVersionCreate(@PathVariable("applicationId") String id,
      * @param model Model with passed parameters.
      * @return "callbackCreate" view.
      */
-    @GetMapping("/application/detail/{applicationId}/callback/create")
+    @GetMapping("/application/detail/{applicationId}/callback/create/form")
```
banterCZ, line 163: Changing URL is not backward compatible but for Admin API we may afford that.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/PowerAuthClient.java
```diff
@@ -1154,16 +1155,19 @@ VaultUnlockResponse unlockVault(String activationId, String applicationKey, Stri
...
+     * @param attributes      Attributes to send in the callback data.
+     * @param authentication  Callback request authentication.
+     * @param retentionPeriod Duration after which a completed callback event is automatically removed.
+     * @param initialBackoff  Initial delay before retry attempt following a callback event failure.
+     * @param maxAttempts     Maximum number of attempts to send a callback event.
      * @return Information about new callback URL object.
      * @throws PowerAuthClientException In case REST API call fails.
      */
-    CreateCallbackUrlResponse createCallbackUrl(String applicationId, String name, CallbackUrlType type, String callbackUrl, List<String> attributes, HttpAuthenticationPrivate authentication) throws PowerAuthClientException;
+    CreateCallbackUrlResponse createCallbackUrl(String applicationId, String name, CallbackUrlType type, String callbackUrl, List<String> attributes, HttpAuthenticationPrivate authentication, Duration retentionPeriod, Duration initialBackoff, Integer maxAttempts) throws PowerAuthClientException;
```
banterCZ, line 1170: > Carthago delenda est

In long term, I prefer using methods with request object only and mark these as _deprecated_.

banterCZ (approved):

romanstrobl (approved):
Approved now.

powerauth-rest-client-spring/src/main/java/com/wultra/security/powerauth/rest/client/PowerAuthRestClient.java
jnpsk: https://github.com/wultra/powerauth-server/issues/1702

powerauth-admin/src/main/java/io/getlime/security/app/admin/controller/ApplicationController.java
jnpsk: Yes, I think we can assume those are just called internally.

PR #1710, opened 2024-10-03
Fix #1702: Use Enum Instead of String for Callback URL Type

### LINKED ISSUE #1702: Use Enum Instead of String for Callback URL Type
Callback URL Management API requests and responses use simple string for Callback URL Type. We may use the already defined enum for that purpose instead.

jnpsk:


romanstrobl (dismissed):
Looks OK.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/CreateCallbackUrlRequest.java
```diff
@@ -43,8 +45,8 @@ public class CreateCallbackUrlRequest {
     @NotBlank
     private String name;
 
-    @NotBlank
-    private String type;
+    @NotNull
+    private CallbackUrlType type;
```
banterCZ, line 49: 🤔 Shouldn't it be reflected in `WebServices-Methods.md`?

banterCZ (approved):
Looks good.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/CreateCallbackUrlRequest.java
jnpsk: sure, added now

banterCZ (approved):

PR #1714, opened 2024-10-03
Fix #1713: Callback Management API could return default values

### LINKED ISSUE #1713: Callback Management API could return default values
Callback Management API now allows to specify retention period or retry policy. When none is set on the callback level, application defaults are used. Those defaults could be returned in responses, when not overridden.

jnpsk:


banterCZ (approved):

romanstrobl (approved):
Good idea.

PR #1716, opened 2024-10-03
Fix #1715: Incorrect inverse condition when evaluating activation scoped temporary keys

### LINKED ISSUE #1715: Incorrect inverse condition when evaluating activation scoped temporary keys

petrdvorak:


romanstrobl (approved):
Looks OK.

banterCZ (approved):

PR #1724, opened 2024-10-04
Fix #1723: Do not include activation ID for the activation related temporary keys

### LINKED ISSUE #1723: Do not include activation ID for the activation related temporary keys

petrdvorak:


kober32 (approved):
Looks OK.

PR #1728, opened 2024-10-04
Fix #1722: Cannot remove callbacks due to foreign key constraint

### LINKED ISSUE #1722: Cannot remove callbacks due to foreign key constraint
When trying to remove a callback URL, I get the following error:

```
Caused by: org.postgresql.util.PSQLException: ERROR: update or delete on table "pa_application_callback" violates foreign key constraint "pa_application_callback_id_fk" on table "pa_application_callback_event"
  Detail: Key (id)=(1c3ba2b2-f279-4265-87c9-c7ca8d22fabb) is still referenced from table "pa_application_callback_event".
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2733) ~[postgresql-42.7.4.jar:42.7.4]
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2420) ~[postgresql-42.7.4.jar:42.7.4]
	... 157 common frames omitted
```

The REST API method is `/rest/v3/application/callback/remove`.

jnpsk:
Fix #1722 by adding `enabled` column to the `pa_application_callback` table for soft-delete purpose. When delete callback request is processed, a database record is not physically deleted anymore, instead `enabled` is set to false.

As @banterCZ suggested I tried `@SQLRestriction("enabled = true")` on the entity level, which applies filter to queries, basically appending "`where enabled = true`" to each query. So a disabled callback can't be listed, updated nor deleted repeatedly. The delete method on the repository is overridden to do an update instead. I added some tests for that.


romanstrobl (approved):
Nice solution!

banterCZ (approved):
A great soft-delete functionality!

PR #1742, opened 2024-10-14
Add deprecations for client methods without all crypto 3.3 parameters

romanstrobl:


banterCZ (approved):
Perfect! But shouldn't we mention that in the migration guide?

banterCZ (approved):
Thanks for updating the migration guides.

PR #1750, opened 2024-10-22
Fix #1747: Unnecessary FK constraint in callback event table

### LINKED ISSUE #1747: Unnecessary FK constraint in callback event table
Callback event table `pa_application_callback_event` has a foreign key (`application_callback_id`) that references `pa_application_callback(id)`. It causes unnecessary key validation during inserts to `pa_application_callback_event` table.

jnpsk:
1. Remove FK constraint on `pa_application_callback_event` on database level.
2. It is not possible to combine `@NotFound` annotation for entity reference mapping with `FetchType.LAZY` on Many to One relationship, therefore there is no entity reference anymore, instead there is just a String storing the id from `pa_application_callback`.
3. The DTO object `CallbackUrlEvent` now includes the callback configuration so it can be used on response handling without querying the database. This is to prevent situation when someone removes the associated `pa_application_callback` when waiting for the callback response.
4. The cache now includes the CallbackUrlEntity to save on database queries and to use the cache consistently both with events and rest client.


banterCZ (approved):
Looks good.

romanstrobl (approved):
Looks OK, let's test this change.

PR #1754, opened 2024-10-23
Fix #1753: Do not store user ID for failed operation approval attempts

### LINKED ISSUE #1753: Do not store user ID for failed operation approval attempts
Currently, when operation approval fails, we store the user ID at the operation to keep the record of the user who attempted to approve the operation. However, this may lead to incorrect operation personalization. Introduced in https://github.com/wultra/powerauth-server/pull/1151.

petrdvorak:


romanstrobl (approved):
Thank you for the fix!

zcgandcomp (commented):
Looks OK.

PR #1767, opened 2024-10-25
Fix #1766: JavaDoc warning when referencing Lombok generated methods

### LINKED ISSUE #1766: JavaDoc warning when referencing Lombok generated methods
There is following JavaDoc warning during compilation
```
[WARNING] Javadoc Warnings
[WARNING] /.../powerauth-server/powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/configuration/CacheConfiguration.java:44: warning: reference not found: CallbackUrlEntity#getId()
[WARNING] * {@link CallbackUrlEntity#getId()} is used as a cache key.
[WARNING] ^
[WARNING] 1 warning
```
That is probably caused by missing Lombok generated methods during JavaDoc generation. Similar issue were already solved here https://github.com/wultra/powerauth-push-server/pull/867

jnpsk:
The solution used in https://github.com/wultra/powerauth-push-server/pull/867 was applied. An alternative is to avoid using references to Lombok generated methods. Currently, the only problem seem so be `{@link CallbackUrlEntity#getId()}` in `CacheConfiguration` class.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,11 +227,39 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok-maven-plugin</artifactId>
+                <version>1.18.20.0</version>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.projectlombok</groupId>
+                        <artifactId>lombok</artifactId>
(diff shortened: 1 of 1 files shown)


romanstrobl (approved):
Thanks for the fix.

banterCZ (approved):
Nice!

PR #1772, opened 2024-11-14
Fix #766: Failing test testEncryptionAndDecryptionDifferentActivationFailServerPrivateKeyConverter

### LINKED ISSUE #766: Failing test testEncryptionAndDecryptionDifferentActivationFailServerPrivateKeyConverter
Log message: 

```
ServerPrivateKeyConverterTest.testEncryptionAndDecryptionDifferentActivationFailServerPrivateKeyConverter:78 Expected io.getlime.security.powerauth.app.server.service.exceptions.GenericServiceException to be thrown, but nothing was thrown.
```

romanstrobl:
The test was failing randomly because sometimes the decryption worked, depending on generated IV value. I fixed the test so it handles the state when decryption succeeds, however the decrypted private key is different.


petrdvorak (approved):
Looks OK

PR #1773, opened 2024-11-15
Fix #1771: Enforce presence of user ID before operation approval

### LINKED ISSUE #1771: Enforce presence of user ID before operation approval
Operations which do not have `userId` specified when created should be always claimed before approval. Operation approval should reject operations without `userId`. Implement operation claim for FIDO2 as part of this request to allow this check.

romanstrobl:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationClaimRequest.java
```diff
@@ -0,0 +1,41 @@
...
+@Data
+public class OperationClaimRequest  {
+
+    /**
+     * Operation identifier.
+     */
+    private String operationId;
+
+    /**
+     * Optional user identifier of the user who is claiming the operation.
```
banterCZ: 🤔 Here it is written _optional_, but required in `OperationClaimRequestValidator`. Could we use the bean validator or do you prefer to be consistent with other use cases?

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -369,13 +369,18 @@ public OperationUserActionResponse attemptApproveOperation(OperationApproveReque
                 throw localizationProvider.buildExceptionForCode(ServiceError.OPERATION_APPROVE_FAILURE);
             }
 
+            final String expectedUserId = operationEntity.getUserId();
+            if (expectedUserId == null) {
+                logger.warn("Operation cannot be approved, because user ID is not set: {}.", operationId);
```
banterCZ: 🤔 What about
```suggestion
                logger.warn("Operation ID: {} cannot be approved, because user ID is not set.", operationId);
```

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/OperationClaimRequest.java
romanstrobl: I will fix the wording. We can move to the bean validator, but lets do it as another issue for all validations.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/validator/OperationClaimRequestValidator.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class OperationClaimRequestValidator {
+
+    public static String validate(OperationClaimRequest source) {
+        if (source == null) {
+            return "Operation claim request must not be null";
+        }
+        if (source.getOperationId() == null) {
```
zcgandcomp:
```suggestion
        if (source.getOperationId() == null || source.getOperationId().isEmpty()) {
```

use the same style as for userId validation

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/controller/api/OperationsController.java
```diff
@@ -81,6 +81,21 @@ public ObjectResponse<OperationDetailResponse> operationDetail(@RequestBody Obje
...
+    /**
+     * Claim operation for a user.
+     *
+     * @param request Claim operation for a user.
+     * @return Get operation response.
+     * @throws Exception In case the service throws exception.
+     */
+    @PostMapping("/claim")
+    public ObjectResponse<OperationDetailResponse> operationClaim(@RequestBody ObjectRequest<OperationClaimRequest> request) throws Exception {
+        logger.info("OperationClaimRequest received: {}", request);
```
zcgandcomp, line 93: Excessive logging. The OperationClaimRequest is not quite small,  but logging of the whole object on info level is an antipattern.

Yes you can change as in the logging guideline.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -369,13 +369,18 @@ public OperationUserActionResponse attemptApproveOperation(OperationApproveReque
...
+            }
+
             // Check the operation properties match the request
             final PowerAuthSignatureTypes factorEnum = PowerAuthSignatureTypes.getEnumFromString(signatureType.toString());
             final ProximityCheckResult proximityCheckResult = fetchProximityCheckResult(operationEntity, request, currentInstant);
-            final String expectedUserId = operationEntity.getUserId();
             final boolean activationIdMatches = activationIdMatches(request, operationEntity.getActivationId());
             final boolean operationShouldFail = operationApprovalCustomizer.operationShouldFail(operationEntity, request);
-            if ((expectedUserId == null || expectedUserId.equals(userId)) // correct user approved the operation
+            if (expectedUserId.equals(userId) // correct user approved the operation
```
zcgandcomp, line 383: The line no. 388:

` operationEntity.setUserId(userId);
`
should be removed. We assert the equality. And there is no need to set the userId - it could lead to mistake in the future.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -763,6 +768,38 @@ public OperationDetailResponse operationDetail(OperationDetailRequest request) t
...
+            final Date currentTimestamp = new Date();
+            final String operationId = request.getOperationId();
+
+            final OperationEntity operation = operationQueryService.findOperationForUpdate(operationId).orElseThrow(() -> {
+                logger.warn("Operation was not found for ID: {}", operationId);
+                return localizationProvider.buildExceptionForCode(ServiceError.OPERATION_NOT_FOUND);
+            });
+
             final String userId = request.getUserId();
             final OperationEntity operationEntity = expireOperation(
```
zcgandcomp: This is IMHO hard to read, and the check on the operation expiration is doubled. No biggie the impact is minimal, but it is not pretty - up to you.

zcgandcomp (changes requested):
In general fine. Few smaller comments. The important part will be to do the proper facade in the enrollment-server so we do not impact mobile applications.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/controller/api/OperationsController.java
romanstrobl: Logging corresponds to the rest of PowerAuth server logging, I also noticed it. I will change it in all controllers in this ticket:

https://github.com/wultra/powerauth-server/issues/1775

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
romanstrobl: I will remove it.

zcgandcomp (approved):
Nice, like the tests.

PR #1777, opened 2024-11-20
fix #1776: Assertion of allowed credentials ID is always false

### LINKED ISSUE #1776: Assertion of allowed credentials ID is always false
Create the registration with FIDO2 device.
Use the registered FIDO2 device to approve the operation, i.e. to verify the assertion.
If the device does not send `attestedCredentailData` in the `authenticatorData` and the assertion is created for a specific user and his credentials the verification fails on the control.

zcgandcomp:


petrdvorak (approved):
Looks OK

jnpsk (approved):

PR #1778, opened 2024-11-21
Fix #1776: Assertion of allowed credentials ID is always false (backport)

### LINKED ISSUE #1776: Assertion of allowed credentials ID is always false

banterCZ:
(cherry picked from commit 12e24eab174ecab2050248962582908e31f15cd5)


romanstrobl (approved):
Looks OK.

PR #1796, opened 2024-11-29
Fix #1795: Validate user ID for operation approve and reject before updating operation

### LINKED ISSUE #1795: Validate user ID for operation approve and reject before updating operation
Move user ID validation before updating the operation failed attempts to avoid potential manipulation with operations of other users, which can lead to denial of service.

romanstrobl:


banterCZ (dismissed):
Looks good.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -375,13 +375,17 @@ public OperationUserActionResponse attemptApproveOperation(OperationApproveReque
                 throw localizationProvider.buildExceptionForCode(ServiceError.OPERATION_APPROVE_FAILURE);
             }
 
+            if (!expectedUserId.equals(userId)) {
+                logger.warn("Operation ID: {} cannot be approved, because user ID from the request '{}' does not match user ID from the operation '{}'.", operationId, userId, expectedUserId);
```
zcgandcomp: can we switch to pattern operationId (this is more easily searchable in logs) Thank you.

zcgandcomp (dismissed):
Over all good. 
Correct returned exception and we corrected previously incorrect exception from reject.

We should start using pattern `operationId` to have easier parsable logs.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
romanstrobl: Ok, I will update it.

banterCZ (approved):

PR #1804, opened 2024-12-13
Fix #1803: Ensure the getOperation doesn't change operation status

### LINKED ISSUE #1803: Ensure the getOperation doesn't change operation status in 1.8.x version

zcgandcomp:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -763,11 +763,10 @@ public OperationDetailResponse operationDetail(OperationDetailRequest request) t
...
 
             final String userId = request.getUserId();
-            final OperationEntity operationEntity = expireOperation(
-                    claimOperation(operation, userId, currentTimestamp),
-                    currentTimestamp
-            );
+            final OperationEntity operationEntity =
+                    claimOperation(operation, userId, currentTimestamp);
             final OperationDetailResponse operationDetailResponse = convertFromEntityAndFillOtp(operationEntity);
+            expireOperation(operationDetailResponse,currentTimestamp);
```
romanstrobl: Add space after `operationDetailResponse,`

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1013,6 +1012,17 @@ private OperationEntity expireOperation(OperationEntity operationEntity, Date cu
         return operationEntity;
     }
 
+    private void expireOperation(OperationDetailResponse operation, Date currentTimestamp) throws GenericServiceException {
```
romanstrobl: Let's not duplicate this logic, we can reuse the code below in new private method shared by both `expireOperation` methods and call it to obtain status.

romanstrobl (changes requested):
Good fix, but lets improve the code style.

powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: This method changes only the response class, not the entity. I suggest to rename it to avoid confusion.

banterCZ (approved):
It should work. Before, the status at the JPA entity level and the hibernate dirty checking caused the problem. Now, the response status has only been changed. The transaction cannot be converted to read-only because of the claim functionality.

jnpsk (approved):
Looks ok.

PR #1814, opened 2024-12-31
Fix #1813: [Feature]: Add (m)TLS configuration of Callbacks

### LINKED ISSUE #1813: [Feature]: Add (m)TLS configuration of Callbacks
### Description

Add new attributes when configuring callbacks via REST API:
 - `trustStoreContent` for the trust store
 - `keyStoreContent` for the key store

Validate that only location or content is used i.e. prevent accidental misconfiguration.

### Acceptance criteria

Callbacks authentication can be configured using REST API using the trust store and key store files content.

The new stored content is used when establishing the TLS. 

Old location setting is still valid and not deprecated.

Only location or content configuration is used at a time.

### Technical specification

The trust store and key store data should be stored as base64 in `CLOB` column data types.

### QA specification

Verify that only one configuration can be made i.e. when location is configured location is used, and vice versa, when content is configured content is used.

Verify that TLS and mTLS is working.

### JIRA issue code

_No response_

romanstrobl:


powerauth-java-server/src/main/java/io/getlime/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehavior.java
```diff
@@ -104,6 +104,23 @@ public CreateCallbackUrlResponse createCallbackUrl(CreateCallbackUrlRequest requ
...
+                    // Rollback is not required, error occurs before writing to database
+                    throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+                }
+                if (request.getAuthentication().getCertificate().isUseCustomTrustStore() &&
+                        request.getAuthentication().getCertificate().getTrustStoreLocation() != null &&
+                        request.getAuthentication().getCertificate().getTrustStoreContent() != null) {
+                    logger.warn("Invalid truststore configuration for callback URL: {}", request.getCallbackUrl());
+                    // Rollback is not required, error occurs before writing to database
+                    throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
+                }
```
banterCZ: What about extracting a `certificate` variable?
```suggestion
                final HttpAuthenticationPrivate.Certificate certificate = request.getAuthentication().getCertificate();
                if (certificate.isUseCustomKeyStore() &&
                        certificate.getKeyStoreLocation() != null &&
                        certificate.getKeyStoreContent() != null) {
                    logger.warn("Invalid keystore configuration for callback URL: {}", request.getCallbackUrl());
                    // Rollback is not required, error occurs before writing to database
                    throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
                }
                if (certificate.isUseCustomTrustStore() &&
                        certificate.getTrustStoreLocation() != null &&
                        certificate.getTrustStoreContent() != null) {
                    logger.warn("Invalid truststore configuration for callback URL: {}", request.getCallbackUrl());
                    // Rollback is not required, error occurs before writing to database
                    throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
                }
```

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/entity/HttpAuthenticationPublic.java
```diff
@@ -37,11 +38,15 @@ public static class Certificate {
         private boolean enabled;
         private boolean useCustomKeyStore;
         private String keyStoreLocation;
+        @ToString.Exclude
+        private String keyStoreContent;
```
banterCZ: Do we really want to expose `keyStoreContent` to the REST API response? Why not follow the same pattern with a flag only such as `keyStorePasswordSet`? Moreover, I do not see changes in the PowerAuth-Admin. Is it intentional?

banterCZ (commented):
Let's clarify REST API, otherwise, it looks good.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/entity/HttpAuthenticationPublic.java
romanstrobl: Good point, I will add the flag.

As for PowerAuth admin, I do not think we need to add it now.

powerauth-admin/src/main/webapp/WEB-INF/jsp/callbackCreate.jsp
```diff
@@ -129,6 +129,12 @@
                                     <input type="text" id="auth_keyStoreLocation" name="auth_keyStoreLocation" class="form-control" value="${auth_keyStoreLocation}"/>
                                 </div>
                             </div>
+                            <div class="form-group">
+                                <label for="attr_activationId" class="col-sm-4 control-label">Keystore Content</label>
```
banterCZ: Strictly speaking, the value of the `for` attribute should be the `id` of the related form field, so in this case `auth_keyStoreContent` (more occurrences in this file).

powerauth-admin/src/main/webapp/WEB-INF/jsp/callbackUpdate.jsp
```diff
@@ -126,10 +126,17 @@
                             </div>
                             <div class="form-group">
                                 <label for="attr_activationId" class="col-sm-4 control-label">Keystore Location</label>
-                                <div class="col-sm-7">
+                                <div class="col-sm-4">
                                     <input type="text" id="auth_keyStoreLocation" name="auth_keyStoreLocation" class="form-control" value="${auth_keyStoreLocation}"/>
                                 </div>
                             </div>
+                            <div class="form-group">
+                                <label for="attr_activationId" class="col-sm-4 control-label">Keystore Content</label>
```
banterCZ: dtto

powerauth-admin/src/main/webapp/WEB-INF/jsp/callbackUpdate.jsp
```diff
@@ -159,10 +166,17 @@
...
+                            <div class="col-sm-4">
                                 <input type="text" id="auth_trustStoreLocation" name="auth_trustStoreLocation" class="form-control" value="${auth_trustStoreLocation}"/>
                             </div>
                         </div>
+                        <div class="form-group">
+                            <label for="attr_activationId" class="col-sm-4 control-label">Truststore Content</label>
+                            <div class="col-sm-4">
+                                <input type="text" id="auth_trustStoreContent" name="auth_trustStoreContent" class="form-control"/>
+                            </div>
+                            <c:if test="${true eq auth_trustStoreContentSet}"><div class="col-sm-4" style="margin-top: 8px">Content is set.</div></c:if>
```
banterCZ, line 178: Using `style` and `class` looks strange, but ok, it has already been there.

banterCZ (dismissed):
Thanks for improvements of the REST API and implementation in the Admin.

powerauth-admin/src/main/webapp/WEB-INF/jsp/callbackCreate.jsp
romanstrobl: Good catch I will fix it.

banterCZ (approved):

zcgandcomp (approved):
Looks OK

PR #1845, opened 2025-02-06
Fix #1844: Migrate encryptors to generified interfaces for crypto4

### LINKED ISSUE #1844: Migrate encryptors to generified interfaces for crypto4
### Task specification

Migrate to the generified interfaces and classes introduced in crypto4 for existing end-to-end encryption.

romanstrobl:
Migration to generified interfaces and classes introduced in crypto4 for existing end-to-end encryption.


petrdvorak (approved):
Looks OK

PR #1862, opened 2025-02-27
Fix #1818: Remove recovery code functionality

### LINKED ISSUE #1818: Remove recovery code functionality
### Description

Recovery code functionality should be removed in alignment with goals of crypto4 project.

### Acceptance criteria

Recovery code functionality is removed from code, API including documentation.

### QA specification

Tests are updated to remove recovery code tests. Code is reviewed.

romanstrobl:
Recovery code functionality was removed from the server.


docs/PowerAuth-Server-2.0.0.md
```diff
@@ -9,3 +9,17 @@ Package names in Java code have been updated from historical `io.getlime` to `co
 We have unified validations in PowerAuth server REST API. The error code returned for failed request validations is always `ERR0024`. As a side effect, the error code `ERR0002` used for case when no application ID was set in request is no longer returned.
 
 The validation of requests is now stricter and more complete to ensure data integrity. In case you get the `ERR0024` error in your integration with PowerAuth server, please make sure the requests contain all parameters, as seen in REST API documentation available at `http[s]://[hostname]:[port]/powerauth-java-server/swagger-ui/index.html`.
+
+### Removed Recovery Code Functionality
```
jnpsk, line 13: We may also explicitly state that it affects some API endpoints in the migration guide
- `/rest/v3/recovery/*` is no longer available.
- `POST /rest/v3/activation/recovery/create` is no longer available.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/RecoveryCodeEntity.java
```diff
@@ -1,209 +0,0 @@
...
-
-/**
- * Database entity for recovery codes.
- *
- * @author Roman Strobl, roman.strobl@wultra.com
- */
-@Entity
-@Table(name = "pa_recovery_code")
-@Getter @Setter
-public class RecoveryCodeEntity implements Serializable {
```
jnpsk, line 41: Shouldn't we also handle deleting the corresponding tables?

jnpsk (commented):
Nice!
I noticed there is a dangling `recoveryCodeStatusSnippet.jsp` which was referenced from the deleted jsp files.   
Also the unused error codes are still in `Server-Error-Codes.md`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/RecoveryCodeEntity.java
romanstrobl: I will handle this change in a separate ticket. We should analyse the impact on rollbacks and auditing.

PR #1868, opened 2025-03-06
Fix #1863: Refactor cryptography in preparation for crypto4

### LINKED ISSUE #1863: Refactor cryptography in preparation for crypto4
### Task specification

Currently cryptography code is tightly coupled with PowerAuth services. We need to move the cryptography code out of the services, and create a generic API which will work both for crypto3 and crypto4 integration. We should use the `protocolVersion` parameter for choosing the right implementation of cryptography when processing requests.

romanstrobl:
This pull requests prepares the code base for crypto4:
- all cryptography code which depends on crypto algorithm was moved behind CryptographyService interface
- preparation of 3 different cryptography algorithms (EC-256, EC-384 and hybrid EC+PQC)
- lots of TODOs are present in the code for v4 implementation, this change targets v3 crypto only


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -553,11 +535,8 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                     // the real encryptedStatusBlob value.
                     if (devicePublicKeyBase64 != null) {
 
-                        final PrivateKey serverPrivateKey = keyConvertor.convertBytesToPrivateKey(EcCurve.P256, Base64.getDecoder().decode(serverPrivateKeyBase64));
-                        final PublicKey devicePublicKey = keyConvertor.convertBytesToPublicKey(EcCurve.P256, Base64.getDecoder().decode(devicePublicKeyBase64));
```
jnpsk, line 557: Both `serverPublicKeyBase64` and `serverPrivateKeyBase64` are now unused in the code above.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -683,10 +663,6 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                 response.setVersion(0L);
                 return response;
             }
-        } catch (InvalidKeySpecException | InvalidKeyException ex) {
-            logger.error(ex.getMessage(), ex);
-            // Rollback is not required, database is not used for writing
-            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_KEY_FORMAT);
```
jnpsk, line 689: The exception `InvalidKeySpecException` is now handled differently in the `CryptographyServiceEc256`.
Just pointing out that some specific errors are swallowed to a generic one in this PR on many places.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -838,10 +799,14 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
...
             activation.setFailedAttempts(0L);
-            activation.setApplication(masterKeyPair.getApplication());
-            activation.setMasterKeyPair(masterKeyPair);
+            activation.setApplication(applicationEntity);
+            activation.setMasterKeyPair(cryptographyService.getMasterKeyPair(applicationId).getMasterKeyPair());
+            // Server private and public keys are updated in the next step
+            // TODO - revise this
+            activation.setServerPrivateKeyEncryption(EncryptionMode.NO_ENCRYPTION);
+            activation.setServerPrivateKeyBase64("");
+            activation.setServerPublicKeyBase64("");
```
jnpsk: This should probably be deleted. It is later overriden by the `generateDeviceKeyPair` method.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -852,14 +817,13 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
...
-            final ServerPrivateKey serverPrivateKey = serverPrivateKeyConverter.toDBValue(serverKeyPrivateBytes, userId, activationId);
-            activation.setServerPrivateKeyEncryption(serverPrivateKey.encryptionMode());
-            activation.setServerPrivateKeyBase64(serverPrivateKey.serverPrivateKeyBase64());
-
             activationHistoryServiceBehavior.saveActivationAndLogChange(activation);
             callbackUrlBehavior.notifyCallbackListenersOnActivationChange(activation);
 
+            // Generate server key pair
+            // TODO - v4 support
+            cryptographyService.generateDeviceKeyPair(activationId);
```
jnpsk: The comment talks about server key pair, but then `generateDeviceKeyPair` is called.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -945,73 +901,19 @@ public PrepareActivationResponse prepareActivation(PrepareActivationRequest requ
                     request.getTimestamp()
             );
 
-            // Validate encrypted request
-            if (!encryptorFactory.getRequestResponseValidator(protocolVersion).validateEncryptedRequest(encryptedRequest)) {
```
jnpsk, line 949: This drops validation of `temporaryKeyId`, `nonce` and `timestamp` with respect to the `protocolVersion`.
See the [validation method](https://github.com/wultra/powerauth-crypto/blob/features/crypto4/powerauth-java-crypto/src/main/java/com/wultra/security/powerauth/crypto/lib/encryptor/ecies/EciesRequestResponseValidator.java#L74-L85).

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -1058,14 +960,8 @@ public PrepareActivationResponse prepareActivation(PrepareActivationRequest requ
 
             // Extract the device public key from request
             final byte[] devicePublicKeyBytes = Base64.getDecoder().decode(retrievedDevicePublicKey);
-            PublicKey devicePublicKey = null;
-            try {
-                devicePublicKey = keyConvertor.convertBytesToPublicKey(EcCurve.P256, devicePublicKeyBytes);
-            } catch (InvalidKeySpecException ex) {
-                logger.warn("Invalid public key, activation ID: {}", activation.getActivationId());
-                logger.debug("Invalid public key, activation ID: {}", activation.getActivationId(), ex);
-                handleInvalidPublicKey(activation);
```
jnpsk, line 1067: The `InvalidKeySpecException` is now handled differently and the `handleInvalidPublicKey` is never called, although it makes changes on the activation.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
```diff
@@ -0,0 +1,320 @@
...
+            return EcKeyPair.builder().masterKeyPair(masterKeyPairEntity).ecKeyPair(keyPair).build();
+        } catch (InvalidKeySpecException | CryptoProviderException | GenericCryptoException e) {
+            logger.error("Invalid keypair", e);
+            throw localizationProvider.buildExceptionForCode(ServiceError.NO_MASTER_SERVER_KEYPAIR);
+        }
+    }
+
+    @Override
+    public SecretKey generateSharedSecretKey(String activationId) throws GenericServiceException {
+        final ActivationRecordEntity activation = findActivation(activationId);
```
jnpsk: In many cases when passing activationId or applicationId to methods of this service, the corresponding entity is already fetched. This can be decided later with v4 implementations, but we could potentially save some DB  lookups.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
```diff
@@ -220,116 +131,48 @@ private GetEciesDecryptorResponse getEciesDecryptorParametersForApplication(GetE
...
-            // Lookup the application version and check that it is supported
-            final ApplicationVersionEntity applicationVersion = applicationVersionRepository.findByApplicationKey(applicationKey);
-            if (applicationVersion == null || !applicationVersion.getSupported()) {
-                logger.warn("Application version is incorrect, application key: {}", applicationKey);
-                // Rollback is not required, database is not used for writing
-                throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_APPLICATION);
-            }
-
-            // Check that application key from request belongs to same application as activation ID from request
-            if (!applicationVersion.getApplication().getRid().equals(activation.getApplication().getRid())) {
```
jnpsk, line 265: This check does not seem to happen now.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
```diff
@@ -220,116 +131,48 @@ private GetEciesDecryptorResponse getEciesDecryptorParametersForApplication(GetE
...
 
-            activationValidator.validatePowerAuthProtocol(activation.getProtocol(), localizationProvider);
+        activationValidator.validateActiveStatus(activation.getActivationStatus(), activation.getActivationId(), localizationProvider);
 
-            if (timestamp != null) {
-                // Check ECIES request for replay attacks and persist unique value from request
-                replayVerificationService.checkAndPersistUniqueValue(
-                        UniqueValueType.ECIES_APPLICATION_SCOPE,
-                        new Date(timestamp),
+        final EncryptionContext context = new EncryptionContext(request.getProtocolVersion(), request.getApplicationKey(), null, EncryptorId.ACTIVATION_SCOPE_GENERIC);
```
jnpsk: Shouldn't the `activationId` be passed so the `decryptInActivationScope` is used?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/TemporaryKeyBehavior.java
```diff
@@ -76,91 +40,21 @@
...
-
-            final TemporaryPublicKeyResponse response = new TemporaryPublicKeyResponse();
-            response.setJwt(signedJWT.serialize());
-            return response;
-        } catch (JOSEException ex){
-            logger.debug("JWT token verification failed: {}", ex.getMessage(), ex);
-            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
-        } catch (ParseException ex) {
-            logger.debug("JWT token parsing failed: {}", ex.getMessage(), ex);
-            throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
```
jnpsk, line 161: Both exceptions now throws `GENERIC_CRYPTOGRAPHY_ERROR`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/EncryptionServiceEc256.java
```diff
@@ -0,0 +1,206 @@
...
+    private final ActivationQueryService activationQueryService;
+
+    private final KeyConvertor KEY_CONVERTOR = new KeyConvertor();
+    private final EncryptorFactory ENCRYPTOR_FACTORY = new EncryptorFactory();
+    private final PowerAuthServerKeyFactory SERVER_KEY_FACTORY = new PowerAuthServerKeyFactory();
+
+    public DecryptionResult decrypt(EncryptedRequest encryptedRequest, String protocolVersion, String applicationKey, String activationId, EncryptorId encryptorId, boolean validateRequest) throws GenericServiceException {
+        try {
+            final ApplicationVersionEntity applicationVersion = findApplicationVersion(applicationKey);
+            final ActivationRecordEntity activation = findActivation(activationId);
```
jnpsk: This could be called after the validation of the request.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -482,19 +474,9 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
...
-                        throw localizationProvider.buildExceptionForCode(ServiceError.NO_MASTER_SERVER_KEYPAIR);
-                    }
-                    final String masterPrivateKeyBase64 = masterKeyPairEntity.getMasterKeyPrivateBase64();
-                    final byte[] masterPrivateKeyBytes = Base64.getDecoder().decode(masterPrivateKeyBase64);
-                    final byte[] activationSignature = powerAuthServerActivation.generateActivationSignature(
-                            activation.getActivationCode(),
-                            keyConvertor.convertBytesToPrivateKey(EcCurve.P256, masterPrivateKeyBytes)
-                    );
+                    // TODO - v4 support
+                    final CryptographyService cryptographyService = cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P256);
```
jnpsk: The `CryptographyService` is saved into a variable, yet later in this method, when the `CryptographyService` is needed, it is called "fluently" within the `cryptographyServiceFactory` chain. I guess this becomes cleaner with further implementation of the v4 support.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
```diff
@@ -0,0 +1,320 @@
...
+            keyPair.setName(applicationId + " Default Keypair");
+            masterKeyPairRepository.save(keyPair);
+        } catch (CryptoProviderException e) {
+            logger.error("Could not generate keypair", e);
+            throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
+        }
+    }
+
+    @Override
+    public BaseKeyPair getMasterKeyPair(String applicationId) throws GenericServiceException {
```
jnpsk: This method returns both `masterKeyPairEntity` and `keyPair` (derived from `masterKeyPairEntity`), but they don’t seem to be used together. Additionally, there’s a case where only the entity is used, making the key pair derivation unnecessary.
That said, this might just be a minor detail given the convenience of the abstraction. The v4 implementation will probably make it more clear.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
```diff
@@ -220,116 +131,48 @@ private GetEciesDecryptorResponse getEciesDecryptorParametersForApplication(GetE
...
+
+        activationValidator.validatePowerAuthProtocol(activation.getProtocol(), localizationProvider);
 
-            activationValidator.validatePowerAuthProtocol(activation.getProtocol(), localizationProvider);
+        activationValidator.validateActiveStatus(activation.getActivationStatus(), activation.getActivationId(), localizationProvider);
 
-            if (timestamp != null) {
-                // Check ECIES request for replay attacks and persist unique value from request
-                replayVerificationService.checkAndPersistUniqueValue(
-                        UniqueValueType.ECIES_APPLICATION_SCOPE,
```
jnpsk, line 246: It seems the `UniqueValueType.ECIES_ACTIVATION_SCOPE` will be used now instead, which seems correct to me.

jnpsk (commented):
Nice!
I am adding some comments, mostly minor details.
Please double-check the changes made in `EciesEncryptionBehavior#getEciesDecryptorParametersForActivation`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Yes, that is OK. These cryptography errors occur only when the cryptography fails, and we have sufficient logging for it and we do not need to propagate the error specifics to the client.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: It's there because the record will not be persisted with null values. I'll revise it in another PR.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Ok, I will remove the comment.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Yes, I believe this is correct, because the validator is always called during decryption:
https://github.com/wultra/powerauth-server/blob/e9d4fb618fe185e26e139f24b80b3e610b8a6009/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/EncryptionServiceEc256.java#L86

The validation should be done only once per request.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/TemporaryKeyBehavior.java
romanstrobl: Ok, we can keep these specific errors, I will update the code.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Ok, I will handle this case, as there is business logic attached to it.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: Yes, I am aware of this performance issue. I suggest we handle it later in a separate issue.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
romanstrobl: Yes, `activationId` should be passed, thank you for catching this.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/EncryptionServiceEc256.java
romanstrobl: Good idea.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: I was changing all usages to fluent calls, I will update it.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
romanstrobl: I think the original code was slightly incorrect - in this case the method is `getEciesDecryptorParametersForActivation`, so the replay attack should take into account the activation ID value (it is present in activation scope) and check the unique value including the activation ID.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/EciesEncryptionBehavior.java
romanstrobl: Thanks, I will reintroduce the check.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: At any case I will review the individual cases so we match the original error handling as much as possible, even I don't think we should be too specific for these errors.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: I will separate the use cases, there is a performance impact.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: I will handle the performance problem in this PR, it was not a good design.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
romanstrobl: Fixed now.

jnpsk (approved):

PR #1882, opened 2025-03-14
Update DB structure for crypto4

romanstrobl:
Updates of database structure for crypto4, allowing crypto agility in the future.


jnpsk (approved):
Looks OK. We should also keep the `Database-Structure.md` updated.

romanstrobl:
Yes, I will update `Database-Structure.md`, too.

PR #1891, opened 2025-03-27
Do not allow null values for encryption modes for new DB columns

### LINKED ISSUE #1890: Do not allow null values for encryption modes for new DB columns
### Task specification

The encryption mode columns are missing a `nullable=false` parameter.

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/ActivationRecordEntity.java
```diff
@@ -243,7 +243,7 @@ public class ActivationRecordEntity implements Serializable {
     /**
      * Mode of server private keys encryption {@code (0 = NO_ENCRYPTION, 1 = AES_HMAC)}.
      */
-    @Column(name = "server_private_keys_encryption")
+    @Column(name = "server_private_keys_encryption", nullable = false)
     @Enumerated
     private EncryptionMode serverPrivateKeysEncryption;
```
jnpsk, line 248: Wondering if it is correct to keep column nullable on database level, but set not null on JPA level. Furthermore, there is default value set on database level and not on JPA level.   
I am not sure if that combination works well. Would not the JPA save fail if the entity's value is set to NULL, as it marked as `nullable = false` without knowing there is a default value set?

romanstrobl: Thanks for pointing that out, we'll need to fix Liquibase as well.

We should use what was used on the original columns:
```
            <column name="server_private_key_encryption" type="integer" defaultValue="0">
                <constraints nullable="false" />
            </column>
```

jnpsk (approved):
looks ok

PR #1893, opened 2025-03-31
Fix #1819: Implement E2EE for crypto4

### LINKED ISSUE #1819: Implement E2EE for crypto4
### Description

Integrate the new end-to-end encryption algorithm into PowerAuth server endpoints for crypto4.

### Acceptance criteria

It is possible to use the new shared secret derivation, which depends on chosen algorithm.

#### Shared secret request

The `SharedSecretRequest` object contains the following properties:

| Property | Type | Required | Description |
|--|--|--|--|
| algorithm    | String          | always included         | Selected algorithm: `EC_P384`, `EC_P384_ML_L3`  |
| ecdhe        | String, Base64  | `EC_P384`, `EC_P384_ML_L3`  | Client's EC public key for ECDH calculation |
| mlkem        | String, Base64  | `EC_P384_ML_L3`              | ML-KEM encapsulation key                    |

#### Shared secret response

The `SharedSecretResponse` object contains the following properties:

| Property | Type | Required | Description |
|--|--|--|--|
| ecdhe        | String, Base64  | `EC_P384`, `EC_P384_ML_L3`  | Server's EC public key for ECDH calculation |
| mlkem        | String, Base64  | `EC_P384_ML_L3`              | ML-KEM ciphertext                           | 

### QA specification

Server returns temporary keys in v4 format using a chosen algorithm. The derived shared secret reflects the algorithm and can be used to exchange data with the client including the shared secret containing the PQC KEM output.

romanstrobl:
Implemenation of E2EE for crypto4.

Out of scope (see TODOs):
- using new keys for request validation and response signing for JWT in temporary key functionality, will be implemented separately
- using new master public keys via SDK configuration in temporary key tests, will be implemented separately
- any activation flow and application changes, will be reviewed separately (some tests fail temporarily because of this)

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
@@ -27,10 +27,12 @@ import com.wultra.security.powerauth.app.server.service.model.request.Encryption
 import com.wultra.security.powerauth.app.server.service.persistence.ActivationQueryService;
 import com.wultra.security.powerauth.app.server.service.validator.ActivationContextValidator;
 import com.wultra.security.powerauth.client.model.request.v4.ExtractEncryptorRequest;
+import com.wultra.security.powerauth.client.model.response.v3.GetEciesDecryptorResponse;
 import com.wultra.security.powerauth.client.model.response.v4.ExtractEncryptorResponse;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptedRequest;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptorId;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptorSecrets;
+import com.wultra.security.powerauth.crypto.lib.encryptor.model.v3.ServerEciesSecrets;
 import com.wultra.security.powerauth.crypto.lib.v4.encryptor.model.context.AeadSecrets;
 import com.wultra.security.powerauth.crypto.lib.v4.encryptor.model.request.AeadEncryptedRequest;
 import lombok.AllArgsConstructor;
@@ -101,16 +103,7 @@ public class EncryptionBehaviorAead {
         final EncryptionContext context = new EncryptionContext(request.getProtocolVersion(), request.getApplicationKey(), null, EncryptorId.APPLICATION_SCOPE_GENERIC);
 
         final EncryptorSecrets encryptorSecrets = encryptionService.deriveSecrets(encryptedRequest, context);
-        if (encryptorSecrets instanceof AeadSecrets aeadSecrets) {
-            // AEAD in V4
-            final ExtractEncryptorResponse response = new ExtractEncryptorResponse();
-            response.setSecretKey(Base64.getEncoder().encodeToString(aeadSecrets.getEnvelopeKey()));
-            response.setSharedInfo2(Base64.getEncoder().encodeToString(aeadSecrets.getSharedInfo2()));
-            return response;
-        }
-        logger.error("Unsupported EncryptorSecrets object");
-        // Rollback is not required, database is not used for writing
-        throw localizationProvider.buildExceptionForCode(ServiceError.DECRYPTION_FAILED);
+        return generateResponse(encryptorSecrets);
     }
 
     private ExtractEncryptorResponse extractEncryptorParametersForActivation(ExtractEncryptorRequest request) throws GenericServiceException {
@@ -140,6 +133,10 @@ public class EncryptionBehaviorAead {
                 ),
                 context
         );
+        return generateResponse(encryptorSecrets);
+    }
+
+    private ExtractEncryptorResponse generateResponse(EncryptorSecrets encryptorSecrets) throws GenericServiceException {
         if (encryptorSecrets instanceof AeadSecrets aeadSecrets) {
             // AEAD V4.0
             final ExtractEncryptorResponse response = new ExtractEncryptorResponse();
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
@@ -1,6 +1,6 @@
 /*
  * PowerAuth Server and related software components
- * Copyright (C) 2023 Wultra s.r.o.
+ * Copyright (C) 2025 Wultra s.r.o.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as published
@@ -14,8 +14,9 @@
  *
  * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
  */
-package com.wultra.security.powerauth.app.server.service.behavior.tasks;
+package com.wultra.security.powerauth.app.server.service.behavior.tasks.v4;
 
 import com.wultra.security.powerauth.app.server.database.model.entity.ActivationRecordEntity;
 import com.wultra.security.powerauth.app.server.service.crypto.CryptographyServiceFactory;
@@ -25,15 +26,13 @@ import com.wultra.security.powerauth.app.server.service.model.ServiceError;
 import com.wultra.security.powerauth.app.server.service.model.request.EncryptionContext;
 import com.wultra.security.powerauth.app.server.service.persistence.ActivationQueryService;
 import com.wultra.security.powerauth.app.server.service.validator.ActivationContextValidator;
-import com.wultra.security.powerauth.client.model.request.v3.GetEciesDecryptorRequest;
 import com.wultra.security.powerauth.client.model.request.v4.ExtractEncryptorRequest;
-import com.wultra.security.powerauth.client.model.response.v3.GetEciesDecryptorResponse;
 import com.wultra.security.powerauth.client.model.response.v4.ExtractEncryptorResponse;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptedRequest;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptorId;
 import com.wultra.security.powerauth.crypto.lib.encryptor.model.EncryptorSecrets;
-import com.wultra.security.powerauth.crypto.lib.encryptor.model.v3.EciesEncryptedRequest;
-import com.wultra.security.powerauth.crypto.lib.encryptor.model.v3.ServerEciesSecrets;
+import com.wultra.security.powerauth.crypto.lib.v4.encryptor.model.context.AeadSecrets;
+import com.wultra.security.powerauth.crypto.lib.v4.encryptor.model.request.AeadEncryptedRequest;
 import com.wultra.security.powerauth.crypto.lib.v4.model.context.SharedSecretAlgorithm;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -43,11 +42,11 @@ import org.springframework.transaction.annotation.Transactional;
 import java.util.Base64;
 
 /**
- * Behavior class implementing the ECIES service logic.
+ * Behavior class implementing the AEAD encryption service logic.
  *
  * <p><b>PowerAuth protocol versions:</b>
  * <ul>
- *     <li>3.0</li>
+ *     <li>4.0</li>
  * </ul>
  *
  * @author Roman Strobl, roman.strobl@wultra.com
@@ -55,7 +54,7 @@ import java.util.Base64;
 @Service
 @Slf4j
 @AllArgsConstructor
-public class EciesEncryptionBehavior {
+public class EncryptionBehaviorAead {
 
     private final LocalizationProvider localizationProvider;
     private final ActivationQueryService activationQueryService;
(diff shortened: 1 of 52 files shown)


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/TemporaryKeyServiceAead.java
```diff
@@ -0,0 +1,420 @@
...
+     * @throws GenericServiceException In case of a cryptography error.
+     */
+    @Override
+    public String requestTemporaryKey(String jwt) throws GenericServiceException {
+        try {
+            final SignedJWT decodedJWT = SignedJWT.parse(jwt);
+            final TemporaryPublicKeyRequestClaims requestClaims = buildTemporaryKeyClaims(decodedJWT);
+
+            // Validate claims
+            final String error = validateDecodedClaims(requestClaims);
```
jnpsk, line 139: The validation procedure is missing validation of the `SharedSecretRequest`. But I assume this will be added in https://github.com/wultra/powerauth-server/issues/1895.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/TemporaryKeyServiceAead.java
```diff
@@ -0,0 +1,420 @@
...
+            }
+
+            final Date currentTimestamp = new Date();
+
+            // Derive shared secret using SharedSecret algorithm
+            final SharedSecretAlgorithm algorithm = SharedSecretAlgorithm.valueOf(requestClaims.getSharedSecretRequest().getAlgorithm());
+            final ResponseCryptogram sharedSecretResponse = deriveSharedSecret(temporaryKeyResult.getSharedSecretRequest(), algorithm);
+
+            // Generate new key and store it
+            final KeyPair temporaryKeyPair = KEY_GENERATOR.generateKeyPair(EcCurve.P384);
```
jnpsk: Is this step still required for v4?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
```diff
@@ -91,12 +101,27 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
-            final MasterKeyPairEntity keyPair = new MasterKeyPairEntity();
+            // Key pairs for multiple algorithms are stored for the same entity
+            final PrivateKeyRegistry privateKeyRegistry;
+            final PublicKeyRegistry publicKeyRegistry;
+            MasterKeyPairEntity keyPair = masterKeyPairRepository.findFirstByApplicationIdOrderByTimestampCreatedDesc(application.getId());
+            if (keyPair == null) {
+                keyPair = new MasterKeyPairEntity();
+                privateKeyRegistry = new PrivateKeyRegistry();
+                publicKeyRegistry = new PublicKeyRegistry();
+                keyPair.setMasterPrivateKeysEncryption(EncryptionMode.NO_ENCRYPTION);
```
jnpsk: The setting to `EncryptionMode.NO_ENCRYPTION` is overriden in this code block later.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
```diff
@@ -91,12 +101,27 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
 
-            final MasterKeyPairEntity keyPair = new MasterKeyPairEntity();
+            // Key pairs for multiple algorithms are stored for the same entity
+            final PrivateKeyRegistry privateKeyRegistry;
+            final PublicKeyRegistry publicKeyRegistry;
+            MasterKeyPairEntity keyPair = masterKeyPairRepository.findFirstByApplicationIdOrderByTimestampCreatedDesc(application.getId());
+            if (keyPair == null) {
+                keyPair = new MasterKeyPairEntity();
+                privateKeyRegistry = new PrivateKeyRegistry();
+                publicKeyRegistry = new PublicKeyRegistry();
```
jnpsk, line 111: It is unnecessary to define those variables before the if-block. Moreover you can move them right before its actual usage, so it is after the _Store empty registries_ comment.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/CryptographyService.java
```diff
@@ -53,46 +64,47 @@ public interface CryptographyService {
...
+    public abstract SecretKey generateSharedSecretKey(ActivationRecordEntity activation) throws GenericServiceException;
 
     /**
      * Generate a key pair for an activation. The key pair can be composite in case of a hybrid algorithm.
      *
      * @param activation Activation.
      * @throws GenericServiceException In case of a cryptography error.
      */
-    void generateDeviceKeyPair(ActivationRecordEntity activation) throws GenericServiceException;
+    public abstract void generateDeviceKeyPair(ActivationRecordEntity activation) throws GenericServiceException;
```
jnpsk: I know the generated keys are attached to a device, however the naming can cause confusion, as it actually manipulates with something called _server_ keys under the hood.
Moreover there is the `storeDevicePublicKey` which operates with the actual key from a device.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
```diff
@@ -60,18 +138,55 @@ public SecretKey generateSharedSecretKey(ActivationRecordEntity activation) thro
...
     @Override
     public void generateDeviceKeyPair(ActivationRecordEntity activation) throws GenericServiceException {
-        // TODO
+        try {
+            final KeyPair serverKeyPair = KEY_GENERATOR_EC.generateKeyPair(EcCurve.P384);
+
+            // Store server public key in JSON format
+            final PublicKeyRegistry serverPublicKeys = new PublicKeyRegistry();
+            serverPublicKeys.storePublicKey(SharedSecretAlgorithm.EC_P384, KeyType.ECDSA, serverKeyPair.getPublic());
+            activation.setServerPublicKeys(publicKeysConverter.toDBValue(serverPublicKeys));
```
jnpsk, line 147: Is it safe to assume no other keys are present? What about defaults for the v3 columns?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -60,18 +153,80 @@ public SecretKey generateSharedSecretKey(ActivationRecordEntity activation) thro
...
+        }
         return null;
     }
 
     @Override
     public void storeDevicePublicKey(ActivationRecordEntity activation, BasePublicKey devicePublicKey) throws GenericServiceException {
-        // TODO
+        // The device public key is stored in JSON format in column device_public_keys
+        if (devicePublicKey instanceof EcPublicKey) {
+            final PublicKey ecPublicKey = ((EcPublicKey) devicePublicKey).getEcPublicKey();
```
jnpsk: You may use this feature instead
```suggestion
        if (devicePublicKey instanceof EcPublicKey ecDevicePublicKey) {
            final PublicKey ecPublicKey = ecDevicePublicKey.getEcPublicKey();
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
```diff
@@ -0,0 +1,156 @@
...
+    private ExtractEncryptorResponse extractEncryptorParametersForApplication(ExtractEncryptorRequest request) throws GenericServiceException {
+        final EncryptedRequest encryptedRequest = new AeadEncryptedRequest(
+                request.getTemporaryKeyId(),
+                null,
+                request.getNonce(),
+                request.getTimestamp()
+        );
+        final EncryptionContext context = new EncryptionContext(request.getProtocolVersion(), request.getApplicationKey(), null, EncryptorId.APPLICATION_SCOPE_GENERIC);
+
+        final EncryptorSecrets encryptorSecrets = cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P384).deriveSecrets(encryptedRequest, context);
```
jnpsk: How is it with `EC_P384_ML_L3` support?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
```diff
@@ -0,0 +1,156 @@
...
+            logger.info("Activation does not exist, activation ID: {}", activationId);
+            // Rollback is not required, database is not used for writing
+            return localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_NOT_FOUND);
+        });
+
+        activationValidator.validatePowerAuthProtocol(activation.getProtocol(), localizationProvider);
+
+        activationValidator.validateActiveStatus(activation.getActivationStatus(), activation.getActivationId(), localizationProvider);
+
+        final EncryptionContext context = new EncryptionContext(request.getProtocolVersion(), request.getApplicationKey(), activationId, EncryptorId.ACTIVATION_SCOPE_GENERIC);
```
jnpsk, line 126: It seems that the two methods differ only in building the context. The other processing may be extracted to a common method.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/EncryptionServiceAead.java
```diff
@@ -0,0 +1,186 @@
...
+import javax.crypto.SecretKey;
+import java.security.InvalidKeyException;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.util.Date;
+
+@Service
+@Slf4j
+public class EncryptionServiceAead extends EncryptionService {
```
jnpsk, line 71: Missing javadoc

jnpsk (commented):
Nice!
I am sharing some notes.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/TemporaryKeyServiceAead.java
romanstrobl: Good point, I added the validations.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/TemporaryKeyServiceAead.java
romanstrobl: You're right, this is a leftover from v3. I will remove key generation and change keypair columns to nullable, because we just need to store the encrypted shared secret for v4.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: Good catch, removed. However I changed the key generation process to deterministic and ordered in the next PR: https://github.com/wultra/powerauth-server/pull/1899 because we need to share the EC keypair between both crypto4 algorithms after simplification of JSON model for storing keys in database.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: This code is simplified in https://github.com/wultra/powerauth-server/pull/1899, so it is outdated.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/CryptographyService.java
romanstrobl: Yes, this does not sound right, and I already changed it in the activation process. I will change it here as well.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
romanstrobl: I changed the key generation procedure to ordered and deterministic in https://github.com/wultra/powerauth-server/pull/1899/files, let's review it there.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: Ah, yes, I used this pattern elsewhere, that's good.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
romanstrobl: Good question. The code does not depend on algorithm, because it is the same for both algorithms (it uses already derived shared secret). I will refactor the service, so encryption is handled separately since it no longer depends on algorithm used.

jnpsk (approved):
Ok, thanks for reflecting my comments!

PR #1899, opened 2025-04-02
Fix #1894: Update public master key serialization for crypto4

### LINKED ISSUE #1894: Update public master key serialization for crypto4
### Description

New keys are generated and stored in crypto4 and should be returned serialized for the mobile SDK in application detail.

### Acceptance criteria

New keys are serialized and returned in application detail endpoint.

### Key identifiers

| Key ID | Description |
|--------|-------------|
| `0x01` | KEY_MASTER_P256_PUBLIC on P-256 curve, required for protocol 3.2+       |
| `0x02` | KEY_MASTER_ECDSA_P384_PUBLIC on P-384 curve, required for protocol 4.0+ |
| `0x03` | KEY_MASTER_MLDSA65_PUBLIC, required for protocol 4.0+                   |

### QA specification

Change will be tested using integration tests.

romanstrobl:
Updated serialization of master public keys for new algorithms.

The SDK supports a flat list of keys, so we cannot diferrentiate keys by algorithm, hence the ECDSA key is the same for both v4 algorithms.


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -100,40 +98,32 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
+                logger.error("Key pair generation called in invalid order");
+                throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
             } else {
                 final PrivateKeys privateKeys = new PrivateKeys(keyPair.getMasterPrivateKeysEncryption(), keyPair.getMasterPrivateKeys());
                 privateKeyRegistry = masterPrivateKeysConverter.fromDBValue(privateKeys, application.getId());
                 publicKeyRegistry = publicKeysConverter.fromDBValue(keyPair.getMasterPublicKeys());
             }
 
             // Store private and public keys for EC curve P-384 in JSON format
-            privateKeyRegistry.storePrivateKey(SharedSecretAlgorithm.EC_P384_ML_L3, KeyType.ECDSA, privateKey);
```
jnpsk, line 124: The ECDSA key pair is not stored now.
Do you expect the `EC_P384` generation to be called before generation of this hybrid approach? If so, the key pair generation still happens in the beginning of this method.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -100,40 +98,32 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
-                keyPair.setTimestampCreated(new Date());
+                logger.error("Key pair generation called in invalid order");
+                throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
             } else {
                 final PrivateKeys privateKeys = new PrivateKeys(keyPair.getMasterPrivateKeysEncryption(), keyPair.getMasterPrivateKeys());
                 privateKeyRegistry = masterPrivateKeysConverter.fromDBValue(privateKeys, application.getId());
                 publicKeyRegistry = publicKeysConverter.fromDBValue(keyPair.getMasterPublicKeys());
             }
 
             // Store private and public keys for EC curve P-384 in JSON format
```
jnpsk: This comment does not reflect the actual procedure.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -180,11 +170,11 @@ public void generateServerKeyPair(ActivationRecordEntity activation) throws Gene
     public BasePublicKey convertDevicePublicKey(KeyType keyType, byte[] devicePublicKey) throws GenericServiceException {
```
jnpsk, line 162: Just noticed that on attempt of calling `convertDevicePublicKey` with unexpected `keyType`, hybrid approach returns null, but both `Ec*` throws an exception.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
```diff
@@ -101,23 +101,27 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
-                final String masterPublicKeys = publicKeysConverter.toDBValue(publicKeyRegistry);
-                keyPair.setMasterPublicKeys(masterPublicKeys);
+            if (keyPair != null) {
+                logger.error("Key pair generation called in invalid order");
+                throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
             }
+            keyPair = new MasterKeyPairEntity();
+            final PrivateKeyRegistry privateKeyRegistry = new PrivateKeyRegistry();
+            final PublicKeyRegistry publicKeyRegistry = new PublicKeyRegistry();
+            keyPair.setMasterPrivateKeysEncryption(EncryptionMode.NO_ENCRYPTION);
```
jnpsk: useless assignment of the `NO_ENCRYPTION`. probably some leftover

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
```diff
@@ -87,33 +88,29 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
-                privateKeyRegistry = new PrivateKeyRegistry();
-                publicKeyRegistry = new PublicKeyRegistry();
-                // Store empty values for v3
-                keyPair.setMasterKeyPrivateBase64("");
-                keyPair.setMasterKeyPublicBase64("");
-                keyPair.setApplication(application);
-                keyPair.setName(application.getId() + " Default Keypair");
-                keyPair.setTimestampCreated(new Date());
+                logger.error("Key pair generation called in invalid order");
+                throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
```
jnpsk: So it is required to have v3 master key pair first, even when running on v4 protocol?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
```diff
@@ -204,8 +240,10 @@ public CreateApplicationResponse createApplication(CreateApplicationRequest requ
             application.setId(applicationId);
             application = applicationRepository.save(application);
 
-            // TODO - v4 support
-            cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P256).generateMasterKeyPair(application);
+            for (SharedSecretAlgorithm algorithm: SharedSecretAlgorithm.values()) {
+                // Generate key pairs for all supported algorithms
+                cryptographyServiceFactory.getService(algorithm).generateMasterKeyPair(application);
+            }
```
jnpsk, line 246: Ahh, ok now I get the change in key generation ordering.
Still, I am not sure about the hardwired assumption a v3 key have to exists to run v4 protocol.

jnpsk (commented):
Good idea flattening the registry to reduce complexity of the structure.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
romanstrobl: Yes, it is required. When we initialize an activation we do not know whether it will be v3 or v4. The application detail returns both v3 and v4 public keys at the moment. It will change in the future once v3 is dropped.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: Removed, thanks. Yes, the order is expected to be P-256, P-384 and PQC.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: I will unify the error handling.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v3/CryptographyServiceEc256.java
romanstrobl: Removed.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
romanstrobl: I suggest we move the key generation to a separate service. I think it will simplify the logic which now depends on external calling in correct order. As of now, v3 is still supported, so we need to still generate v3 master keypairs. Without access to v3 keypairs part of the protocol would not work, e.g. temporary keys - it would not be possible to do an upgrade to v4 if we did not include the v3 master public key in application detail. We will drop v3 support in a future release.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -160,6 +160,9 @@ public void generateServerKeyPair(ActivationRecordEntity activation) throws Gene
 
     @Override
     public BasePublicKey convertDevicePublicKey(KeyType keyType, byte[] devicePublicKey) throws GenericServiceException {
+        if (keyType != KeyType.ECDSA_P384 && keyType != KeyType.MLDSA_65) {
+            throw new IllegalArgumentException("Unsupported key type: " + keyType);
+        }
```
jnpsk: Alternatively, this could be the default case of the switch bellow. Then the return null may be removed.

jnpsk (approved):
Looks ok

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: Sure, we can do that.

Issue #1910, opened 2025-04-14
Implement activation confirmation endpoint


romanstrobl:
### Description

In crypto protocol version 4 we add an activation confirmation endpoint which is used to confirm activation after setting up the biometry. We will update the state transition to `ACTIVE` stay due to this addition.

### Acceptance criteria

The activation confirmation endpoint is added and state transition for activation is updated.

#### New `/pa/v4/activation/confirm`

This new endpoint confirm activation from the client's perspective. The purpose of this endpoint is to set initial biometry factor configuration and to make sure that trust between the client and the server is established properly.

The new endpoint specification:

- Request payload is signed with PowerAuth authentication code, with combination of knowledge and possession factors
- Request payload:
  ```json
  {
    "requestObject": {
      "enableBiometry": true
    }
  }
  ```
  Properties in the request:
  - `enableBiometry` - boolean indicating that the mobile application enabled biometry factor for signing.
- Response payload (success):
  ```json
  {
    "status": "OK",
  }
  ```
- Mobile application may retry the operation in the following cases:
  - Reason of failure is due to a network error (e.g. response from the server was not received)
  - In case of network error, mobile SDK should get activation status and determine whether the activation was already confirmed. 

- Server may accept a repeated requests.

- Client must complete confirmation and persist the activation, only after the successful response.

#### Activation state

We won't add an additional state to the activation state enumeration. The reason for this is the compatibility with older clients and also the fact, waiting for commit on the server and waiting for client's confirmation may happen in parallel. The proposed solution how to solve this is to add an additional information about the pending confirmation into the activation record. Let's call it "pendingConfirmation" and apply the following changes to the state diagram:

- Transition from "PENDING_COMMIT" to "ACTIVE" is possible only when "pendingConfirmation" is `false`
- For protocol V3: "pendingConfirmation" is set to `false` automatically when the server receives device public keys
- For protocol V4: "pendingConfirmation" is set to `false` by receiving valid  `/pa/v4/activation/confirm` request

The information about pending confirmation should be added to the activation status blob.

### QA specification

Tested by end-to-end tests.

PR #1913, opened 2025-04-14
Fix #1821: Update activation flow for crypto4

### LINKED ISSUE #1821: Update activation flow for crypto4
### Description

Update the activation flow with new functionality for crypto4:

- updated requests and responses with algorithm choice
- multiple keys echanged for EC and PQC algorithms
- changes related to storing new keys in database
- shared secret is derived and stored in database in encrypted form

Cover the new activation functionality with tests.

Following tasks will be done separately to avoid too large pull request: 
- activation confirm endpoint with optional biometry configuration
- updated transition from `PENDING_COMMIT` state to `ACTIVE` state
- new activation signatures
- updated activation fingerprint

### Acceptance criteria

Activations can be created for both algorithms (EC and EC+PQC).

### Technical specification

The main change in activation is that we don't deduce activation's shared secret from the same keys we use for asymmetric signatures. The inner request payload contains two sets of keys - public keys for data signing (ECDSA or/and ML-DSA) and keys required for the shared secret deduction.

- The JSON request object before E2EE level 2 encryption:
  - There's new `sharedSecretRequest` property containing information required for `KEY_ACTIVATION_SECRET` computation. The selection of algorithm also determines the public keys exchanged for future data signing
  - There's new `devicePublicKeys` property replacing old `devicePublicKey`, containing public keys required for algorithm selected in `sharedSecret.algorithm`. The object contains the following properties:
    - `ecdsa` - contains `KEY_DEVICE_ECDSA_P384_PUBLIC` for `EC_P384` and `EC_P384_ML_L3`.
    - `mldsa` - contains `KEY_DEVICE_MLDSA65_PUBLIC` for `EC_P384_ML_L3`.
  - Example:
    ```json
    {
      "sharedSecretRequest": {
        "algorithm": "EC_P384_ML_L3",
        "ecdhe": "Base64",
        "mlkem": "Base64",
      },
      "devicePublicKeys": {
        "ecdsa": "Base64",
        "mldsa": "Base64"
      },
      "activationName": "My iPhone",
      "activationOtp": "12345678",
      "platform": "ios",
      "deviceInfo": "iPhone12,3",
      "extras": "Any data in string format"
    }
    ```

The similar changes are applied also in the response. 

- The JSON request object before E2EE level 2 encryption:
  - There's new `sharedSecretResponse` property containing information required for `KEY_ACTIVATION_SECRET` computation.
  - There's new `serverPublicKeys` property containing public keys required for algorithm selected in `sharedSecret.algorithm`.
    - `ecdsa` - contains `KEY_SERVER_ECDSA_P384_PUBLIC` for `EC_P384` and `EC_P384_ML_L3`.
    - `mldsa` - contains `KEY_SERVER_MLDSA65_PUBLIC` for `EC_P384_ML_L3`.
  - There's no `activationRecovery` property, because we're removing this feature from the protocol.
  - Example:
    ```json
    {
      "sharedSecretResponse": {
        "ecdhe": "Base64",
        "mlkem": "Base64",
      },
      "serverPublicKeys": {
        "ecdsa": "Base64",
        "mldsa": "Base64"
      },
      "ctrData": "Base64",
      "activationId": "UUID"
    }
    ```

### QA specification

Feature will be tested manually using command line tool, later we will cover all functionality by end-to-end tests.

romanstrobl:
Update the activation flow with new functionality for crypto4:

- updated requests and responses with algorithm choice
- multiple keys echanged for EC and PQC algorithms
- changes related to storing new keys in database
- shared secret is derived and stored in database in encrypted form

Following tasks will be done separately to avoid too large pull request:

- activation confirm endpoint with optional biometry configuration
- updated transition from `PENDING_COMMIT` state to `ACTIVE` state
- new activation signatures
- updated activation fingerprint


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/CryptographyService.java
```diff
@@ -54,22 +54,23 @@ public abstract class CryptographyService {
...
+     * @param keyType Key type.
      * @param application Application.
      * @return Key pair.
      * @throws GenericServiceException In case of a cryptography error.
      */
-    public abstract BaseKeyPair getMasterKeyPair(ApplicationEntity application) throws GenericServiceException;
+    public abstract BaseKeyPair getMasterKeyPair(KeyType keyType, ApplicationEntity application) throws GenericServiceException;
 
     /**
      * Generate shared secret key.
```
jnpsk:
```suggestion
     * Derive shared secret key.
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationCreateServiceBehavior.java
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Activation service specific to V3.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Service("activationCreateServiceBehaviorV3")
+@Slf4j
+@AllArgsConstructor
+public class ActivationCreateServiceBehavior {
```
jnpsk, line 62: There is naming inconsistency between `ActivationCreateServiceBehavior` and `ActivationServiceInitBehavior`.
Maybe `ActivationInitServiceBehavior` would be better.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationProcessServiceBehavior.java
```diff
@@ -0,0 +1,192 @@
...
+                // Activation failed due to invalid ECIES request, rollback transaction
+                throw localizationProvider.buildRollbackingExceptionForCode(ServiceError.INVALID_REQUEST);
+            }
+
+            // Validate activation OTP for stage ON_KEY_EXCHANGE
+            activationServiceValidationBehavior.validateActivationOtp(CommitPhase.ON_KEY_EXCHANGE, layer2Request.getActivationOtp(), activation, null);
+
+            // If activation OTP is provided and valid, or commit phase is ON_KEY_EXCHANGE, then the status is set directly to "ACTIVE".
+            final boolean isActive = StringUtils.hasText(layer2Request.getActivationOtp()) || activation.getCommitPhase() == CommitPhase.ON_KEY_EXCHANGE;
+            final ActivationStatus activationStatus = isActive ? ActivationStatus.ACTIVE : ActivationStatus.PENDING_COMMIT;
```
jnpsk, line 106: This code is shared for create and prepare. But it seems that during [the create procedure](https://github.com/wultra/powerauth-server/blob/features/crypto4/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java#L1155-L1156) the activation status was formerly always set to `PENDING_COMMIT`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationCreateServiceBehavior.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * Deactivate the activation in CREATED or PENDING_COMMIT if it's activation expiration timestamp
+     * is below the given timestamp.
+     *
+     * @param timestamp  Timestamp to check activations against.
+     * @param activation Activation to check.
+     */
+    private void deactivatePendingActivation(Date timestamp, ActivationRecordEntity activation, boolean isActivationLocked) throws GenericServiceException {
```
jnpsk: 🤔 this exact method is now in the shared `ActivationServiceBehavior` and then in both v3 and v4 version of `ActivationCreateServiceBehavior`. Maybe it would worth have this specific method deduplicated as it is version independent and also manages critical functionality.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationCreateServiceBehavior.java
```diff
@@ -0,0 +1,276 @@
...
+            // Get current timestamp
+            final Date timestamp = new Date();
+
+            // Decrypt activation data
+            final EncryptionContext context = new EncryptionContext(protocolVersion, applicationKey, null, EncryptorId.ACTIVATION_LAYER_2);
+            final DecryptionResult decryptionResult = encryptionService.decryptRequest(encryptedRequest, context);
+            final ApplicationEntity application = decryptionResult.getApplication();
+
+            // Prepare activation OTP mode
+            final com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation activationOtpValidation = activationOtp != null ? com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.ON_COMMIT : com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation.NONE;
```
jnpsk: The use of fully qualified name seems unnecessary

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationProcessServiceBehavior.java
```diff
@@ -0,0 +1,315 @@
...
+            // Decrypt activation data
+            final byte[] activationData = decryptionResult.getDecryptedData();
+
+            // Convert JSON data to activation layer 2 request object
+            ActivationLayer2Request layer2Request;
+            try {
+                layer2Request = objectMapper.readValue(activationData, ActivationLayer2Request.class);
+            } catch (IOException ex) {
+                logger.warn("Invalid activation request, activation ID: {}", activation.getActivationId());
+                // Activation failed due to invalid ECIES request, rollback transaction
```
jnpsk:
```suggestion
                // Activation failed due to invalid request, rollback transaction
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationProcessServiceBehavior.java
```diff
@@ -0,0 +1,315 @@
...
+            activation.setActivationName(layer2Request.getActivationName());
+            activation.setExternalId(layer2Request.getExternalId());
+            activation.setExtras(layer2Request.getExtras());
+            if (layer2Request.getPlatform() != null) {
+                activation.setPlatform(layer2Request.getPlatform().toLowerCase());
+            } else {
+                activation.setPlatform("unknown");
+            }
+            activation.setDeviceInfo(layer2Request.getDeviceInfo());
+            // PowerAuth protocol version 3.0 uses 0x3 as version in activation status
```
jnpsk:
```suggestion
            // PowerAuth protocol version 4.0 uses 0x4???? as version in activation status
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationProcessServiceBehavior.java
```diff
@@ -0,0 +1,315 @@
...
+        } catch (RuntimeException ex) {
+            logger.error("Runtime exception or error occurred, transaction will be rolled back", ex);
+            throw ex;
+        } catch (Exception ex) {
+            logger.error("Unknown error occurred", ex);
+            throw new GenericServiceException(ServiceError.UNKNOWN_ERROR, ex.getMessage());
+        }
+    }
+
+    private ActivationLayer2Response handleKeyAgreement(ActivationRecordEntity activation, ActivationLayer2Request layer2Request, String ctrDataBase64) throws GenericServiceException, GenericCryptoException, CryptoProviderException {
```
jnpsk: Just a nitpick - for naming consistency, the method could be named `computeSharedSecret` or something like that as we used the term shared secret more often than key agreement.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationCreateServiceBehavior.java
```diff
@@ -0,0 +1,279 @@
...
+     *     <li>3.2</li>
+     *     <li>3.3</li>
+     * </ul>
+     *
+     * @param request Request with prepared activation.
+     * @return Prepare activation response.
+     * @throws GenericServiceException If invalid values are provided.
+     */
+    @Transactional
+    public PrepareActivationResponse prepareActivation(PrepareActivationRequest request) throws GenericServiceException {
```
jnpsk, line 90: It might be little confusing that we have `Activation*Create*Service` with method `prepare`, which actually assumes having the activation already created.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
```diff
@@ -119,15 +125,48 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
+        }
+        final MasterKeyPairEntity masterKeyPairEntity = masterKeyPairRepository.findFirstByApplicationIdOrderByTimestampCreatedDesc(application.getId());
+        if (masterKeyPairEntity == null) {
+            logger.error("Missing master key pair for application ID: {}", application.getId());
+            // Rollback is not required, database is not used for writing
+            throw localizationProvider.buildExceptionForCode(ServiceError.NO_MASTER_SERVER_KEYPAIR);
+        }
+        try {
+            final String masterPrivateKeyBase64 = masterKeyPairEntity.getMasterPrivateKeys();
+            final EncryptionMode masterPrivateKeyEncryption = masterKeyPairEntity.getMasterPrivateKeysEncryption();
```
jnpsk: A big nitpick
```suggestion
            final String masterPrivateKeysBase64 = masterKeyPairEntity.getMasterPrivateKeys();
            final EncryptionMode masterPrivateKeysEncryption = masterKeyPairEntity.getMasterPrivateKeysEncryption();
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
```diff
@@ -119,15 +125,48 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
...
+            // Rollback is not required, database is not used for writing
+            throw localizationProvider.buildExceptionForCode(ServiceError.NO_MASTER_SERVER_KEYPAIR);
+        }
     }
 
     @Override
-    public SecretKey generateSharedSecretKey(ActivationRecordEntity activation) throws GenericServiceException {
-        // TODO
-        return null;
+    public SecretKey deriveSharedSecretKey(ActivationRecordEntity activation) throws GenericServiceException {
```
jnpsk, line 161: Nitpick - Maybe use different name for the method, as it actually just fetches a data from db, without any derivation. For example fetch or retrieve.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -208,31 +250,120 @@ public void storeDevicePublicKey(ActivationRecordEntity activation, BasePublicKe
...
+                    yield SIGNATURE_UTILS.validateECDSASignature(EcCurve.P384, data, signature, devicePublicKey);
+                } catch (CryptoProviderException | GenericCryptoException | InvalidKeyException e) {
+                    logger.error("Invalid keypair", e);
+                    throw localizationProvider.buildExceptionForCode(ServiceError.GENERIC_CRYPTOGRAPHY_ERROR);
+                }
+            }
+            case MLDSA_65 -> {
+                try {
+                    final PublicKeyRegistry publicKeyRegistry = publicKeysConverter.fromDBValue(activation.getDevicePublicKeys());
+                    final PublicKey devicePublicKey = publicKeyRegistry.getPublicKey(KeyType.ECDSA_P384).orElseThrow(() -> {
```
jnpsk: Wrong KeyType for the MLDSA signature
```suggestion
                    final PublicKey devicePublicKey = publicKeyRegistry.getPublicKey(KeyType.MLDSA_65).orElseThrow(() -> {
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/model/crypto/PqcKeyPair.java
```diff
@@ -27,19 +27,16 @@
...
- * Hybrid key pair wrapper.
+ * PQC key pair wrapper.
  *
  * @author Roman Strobl, roman.strobl@wultra.com
  */
 @Data
 @EqualsAndHashCode(callSuper = true)
 @SuperBuilder
-public class HybridKeyPair extends BaseKeyPair {
+public class PqcKeyPair extends BaseKeyPair {
```
jnpsk: With this change all the BaseKeyPair inheritors has exactly the same fields, just with different naming. Does it worth to keep it on this abstraction level in that case?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
```diff
@@ -122,15 +124,52 @@ public void generateMasterKeyPair(ApplicationEntity application) throws GenericS
     }
 
     @Override
-    public BaseKeyPair getMasterKeyPair(ApplicationEntity application) throws GenericServiceException {
-        // TODO
-        return null;
+    public BaseKeyPair getMasterKeyPair(KeyType keyType, ApplicationEntity application) throws GenericServiceException {
```
jnpsk: So for the hybrid approach usage, the method have to be actually called twice to get both keys?

jnpsk (commented):
Great job,
I still haven't seen modification of tests, but sharing some first note, mostly just nitpicks. I appreciate that you split up the large activation brain methods into smaller pieces. We may clean the code further later.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationProcessServiceBehavior.java
romanstrobl: Good catch. This is caused by unifying the logic for create and prepare methods.

I think the logic is still valid, because:
- For create method, the commit phase can be only `ON_COMMIT`, as set in `ActivationCreateServiceBehavior`. This eliminates the right condition for `isActive`.
- For create method, the OTP must not be present during `ON_KEY_EXCHANGE`, as checked during validation in `ActivationServiceValidationBehavior`. This eliminates the left condition for `isActive`.

Thus, during create method the activation status is always set to `PENDING_COMMIT`.

To make the intentions clear, I will revise the logic in https://github.com/wultra/powerauth-server/issues/1910 which will move this code into a separate service which will handle the updated state transition.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationCreateServiceBehavior.java
romanstrobl: Ok, I will move the logic into separate service to allow reuse.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationCreateServiceBehavior.java
romanstrobl: Do you have an idea for a better class name?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceEc384.java
romanstrobl: It depends on the version, for version 3 the shared key is derived in this method. For version 4 the shared key is fetched.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: Good catch!

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/model/crypto/PqcKeyPair.java
romanstrobl: Indeed, we can avoid this abstraction now.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/v4/CryptographyServiceHybrid.java
romanstrobl: Yes.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/util/TemporaryKeyTestService.java
```diff
@@ -0,0 +1,224 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.Base64;
+import java.util.Date;
+
+/**
+ * Utilities for handling temporary keys.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Service
```
jnpsk, line 68: 🤔 wouldn't that initiate the service also on regular run?

jnpsk (approved):
looks ok

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/util/TemporaryKeyTestService.java
romanstrobl: The service is in the test package, which should make it not available for regular runtime context.

PR #1946, opened 2025-05-22
Fix #1945: Allow additional activation statuses for temporary keys

### LINKED ISSUE #1945: Allow additional activation statuses for temporary keys
### Description

Following activation statuses should be allowed for extracting AEAD encryptor, due to activation status use cases:
- `ACTIVE`
- `BLOCKED`
- `PENDING_COMMIT`

### Acceptance criteria

It is possible to obtain activation status for all additional statusees.

### QA specification

Change will be tested by end-to-end tests.

romanstrobl:
Opening of additional statuses for end-to-end encryption, triggered by switching activation status encryption to standard encryption.

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
@@ -27,7 +27,6 @@ import com.wultra.security.powerauth.app.server.service.model.ServiceError;
 import com.wultra.security.powerauth.app.server.service.model.request.EncryptionContext;
 import com.wultra.security.powerauth.app.server.service.persistence.ActivationQueryService;
 import com.wultra.security.powerauth.app.server.service.validator.ActivationContextValidator;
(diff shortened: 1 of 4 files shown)


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/EncryptionBehaviorAead.java
```diff
@@ -111,6 +114,10 @@ private ExtractEncryptorResponse extractEncryptorParametersForActivation(Extract
         final String activationId = request.getActivationId();
         final Long timestamp = request.getTimestamp();
         final String nonce = request.getNonce();
+        final List<ActivationStatus> allowedStates = request.getAllowedStates();
+        final List<com.wultra.security.powerauth.app.server.database.model.enumeration.ActivationStatus> convertedStates = allowedStates.stream()
+                        .map(activationStatusConverter::convert)
+                        .toList();
```
jnpsk: Nitpick - you can, alternatively, create a stream directly from the `request.getAllowedStates()`, which would save one unnecessary list definition, and maybe helps to avoid full class name for the `database.model.enumeration.ActivationStatus`.

jnpsk (approved):
Looks good!

PR #1966, opened 2025-06-09
Fix #1963: Add warn level for failed operation assertion

### LINKED ISSUE #1963: Add warn level for failed operation assertion
### Description

During the operation approval multiple condition has to be met. Some already has warning if they fails, but some lacks this warning.

Add a warning-level log if operation approval fails in attemptApprove. Namely, a check of the allowed `activationId` and a proximity check.

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

Review by PR review.

banterCZ:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1114,12 +1114,22 @@ private static String generateTotpSeed(final OperationCreateRequest request, fin
         return null;
     }
 
-    private static boolean proximityCheckPassed(final ProximityCheckResult proximityCheckResult) {
-        return proximityCheckResult == ProximityCheckResult.SUCCESS || proximityCheckResult == ProximityCheckResult.DISABLED;
+    private static boolean proximityCheckPassed(final ProximityCheckResult proximityCheckResult, final OperationEntity operation) {
+        if (proximityCheckResult == ProximityCheckResult.SUCCESS || proximityCheckResult == ProximityCheckResult.DISABLED) {
+            return true;
+        } else {
+            logger.warn("action: approveOperation/proximityCheck, state: failed, result: {}, operationId: {}", proximityCheckResult, operation.getId());
```
banterCZ: 🤔 Of course, I started with just `action: proximityCheck`, but since it is part of operation approval, which is logged at the controller level as `action: approveOperation`, I suggest concatenating the subroutine to be clearer. If agreed, we may discuss the delimiter, whether to use `/`, `#`, `-`, or whatever.

romanstrobl: I would suggest something like:

```
logger.warn("action: approveOperation, step: proximityCheck, state: failed, result: {}, operationId: {}"
```

romanstrobl (commented):
See the suggestion how to improve the logging.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
zcgandcomp: We do not strictly need the `action`, this is when "Important Method Is Invoked" and should have start and stop status (so you can see the span).

In this particular case I can see in the log (for the trace and span)

```
INFO action: approveOperation, state: initiated, operationId:  uuid, userId: uuid, applicationId: string
WARN Proximity check enabled for operation ID: uuid but proximity OTP not sent
INFO action: approveOperation, state: succeeded, result:APPROVAL_FAILED
```

The log should be in the method `fetchProximityCheckResult`. And it is actually there since 1.10.x
We can actually track the method fetchProximityCheckResult in a standard pattern, but is excessive logging from my POV.
And as you can observe, we do not need to have `operationId` in the log, but this is acceptable for convenience.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1114,12 +1114,22 @@ private static String generateTotpSeed(final OperationCreateRequest request, fin
...
+        }
     }
 
-    private static boolean activationIdMatches(final OperationApproveRequest operationApproveRequest, String activationId) {
-        return activationId == null || activationId.equals(operationApproveRequest.getAdditionalData().get("activationId"));
+    private static boolean activationIdMatches(final OperationApproveRequest request, String activationId) {
+        if (activationId == null || activationId.equals(request.getAdditionalData().get("activationId"))) {
+            return true;
+        } else {
+            logger.warn("action: approveOperation/activationIdMatch, state: failed, operationId: {}", request.getOperationId());
```
zcgandcomp: suggestion what to log:
```
INFO action: approveOperation, state: initiated, operationId:  uuid, userId: uuid, applicationId: string
WARN activationIdMatches failed, activationId: uuid
INFO action: approveOperation, state: succeeded, result:APPROVAL_FAILED
```

It is interesting to keep the span logging, but this is excessive.

Do not use a construct `approveOperation/activationIdMatch`. But you are right the action should be reasonably unique.

zcgandcomp (changes requested):

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
banterCZ: Moved to `fetchProximityCheckResult`. I like the Roman's suggestion with the `step` attribute. Not all customers provide us logs with trace and span, I would like to be able to read them in that case as well.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1130,7 +1135,7 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
 
         final Object otpObject = request.getAdditionalData().get(PROXIMITY_OTP);
         if (otpObject == null) {
-            logger.warn("Proximity check enabled for operation ID: {} but proximity OTP not sent", operation.getId());
+            logger.warn("action: approveOperation, step:fetchProximityCheckResult, state: failed, operationId: {}, detail: proximity OTP not sent", operation.getId());
```
romanstrobl: Missing space in `step:fetchProximityCheckResult`

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1139,10 +1144,15 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
             final Duration otpStepDuration = powerAuthServiceConfiguration.getProximityCheckStepDuration();
             final int otpStepCount = powerAuthServiceConfiguration.getProximityCheckStepCount();
             final boolean result = Totp.validateTotpSha256(otp.getBytes(StandardCharsets.UTF_8), Base64.getDecoder().decode(seed), now, otpLength, otpStepCount, otpStepDuration);
-            logger.debug("OTP validation result: {} for operation ID: {}", result, operation.getId());
-            return result ? ProximityCheckResult.SUCCESS : ProximityCheckResult.FAILED;
+            logger.debug("OTP validation result: {} for operationId: {}", result, operation.getId());
+            if (result) {
+                return ProximityCheckResult.SUCCESS;
+            } else {
+                logger.warn("action: approveOperation, step:fetchProximityCheckResult, state: failed, operationId: {}", operation.getId());
```
romanstrobl: dtto

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/OperationServiceBehavior.java
```diff
@@ -1139,10 +1144,15 @@ private ProximityCheckResult fetchProximityCheckResult(final OperationEntity ope
...
+                return ProximityCheckResult.FAILED;
+            }
         } catch (CryptoProviderException | IllegalArgumentException e) {
-            logger.error("Unable to validate proximity OTP for operation ID: {}", operation.getId(), e);
+            logger.error("action: approveOperation, step:fetchProximityCheckResult, state: error, operationId: {}", operation.getId(), e);
```
romanstrobl: dtto

romanstrobl (changes requested):
The formatting seems a little off.

banterCZ:
Let's use it without `step` for the release and improve it later.

zcgandcomp (approved):
Looks OK.

romanstrobl (approved):
Thanks for the fixes.

PR #1979, opened 2025-06-18
Fix #1831: Add protection against replay attacks for nonce used in AEAD

### LINKED ISSUE #1831: Add protection against replay attacks for nonce used in AEAD
### Description

Implement protection against replay attacks for nonce parameter used in AEAD in crypto4.

### Acceptance criteria

Replay attack protection is updated based on specification for version 4, minor updates for versions 3.x may also be required.

### Technical specification

The PowerAuth Server will keep history of used unique values in new `pa_unique_values` table:

- Mapping for MAC Tokens:
  - `type` set to `1`
  - `identifier` set to `TOKEN_ID`
  - `unique_value` set to `NONCE`
  
- Mapping for Application Scoped ECIES: 
  - `type` set to `2`
  - `identifier` is `NULL`
  - `unique_value` set to `KEY_EPH_PUB` + `NONCE`

- Mapping for Activation Scoped ECIES:
  - `type` set to `3`
  - `identifier` is `ACTIVATION_ID`
  - `unique_value` set to `KEY_EPH_PUB` + `NONCE`

- Mapping for ECIES with temporary keys
  - `type` set to `4`
  - `identifier` is `TEMPORARY_KEY_ID`
  - `unique_value` set to `KEY_EPH_PUB` + `NONCE`

- Mapping for AEAD
  - `type` set to `5`
  - `identifier` is `TEMPORARY_KEY_ID`
  - `unique_value` set to `NONCE`

To keep the database table compact, the server should run a periodic cleanup job and remove expired values (e.g. `timestamp_expire` is less than current timestamp).

The PowerAuth Server must reject requests that contains duplicate unique values for each scenario. For example, if MAC token header is being validated, then it must look for type `1` with `APPLICATION_KEY` with `TOKEN_ID` and `NONCE` and if record exist, then the validation must fail.

#### Validations for TOKEN type

The timestamp is already available in 3.1 and older protocol versions, so the server can perform the full nonce validation:

Let's define `EXPIRATION` as

- 60 seconds for protocol 3.2+
- 120 minutes for protocol 3.1 and older

Then

1. Test whether token header is calculated correctly
   - If YES, then continue with step 2
   - If NO, then reject the request

2. Test whether TIMESTAMP is within range  `CURRENT_TIMESTAMP - EXPIRATION `< `TIMESTAMP` < `CURRENT_TIMESTAMP`.
   - If YES, then continue to 3.
   - If NO, then reject the request
   
3. Look for record containing APPLICATION_KEY, NONCE and TOKEN_ID in `pa_unique_values` table
   - If record doesn't exist, then continue with step 4.
   - If record exist and `timestamp_expire` is older than `CURRENT_TIMESTAMP`, then continue with step 4
   - If record exist and `timestamp_expire` is not older than `CURRENT_TIMESTAMP`, then reject the request
   
4. Create a new record in `pa_unique_values` (see Mapping for MAC Tokens above) and accept the request.

#### Validations for ECIES types (V3.2 and older)

The timestamp in ECIES encrypted request is available from the protocol version 3.2, so the full lookup is not available for the older protocol versions.

Let's define `EXPIRATION` as 60 seconds and define `IDENTIFIER` as

- `null` if ECIES is application scoped
- `ACTIVATION_ID` if ECIES is activation scoped

Then

1. Test whether ECIES payload contains TIMESTAMP
   - If YES, then continue with step 2
   - If NO, then accept this request

2. Test whether TIMESTAMP is within range  `CURRENT_TIMESTAMP - EXPIRATION` < `TIMESTAMP` < `CURRENT_TIMESTAMP`.
   - If YES, then continue to 3.
   - If NO, then reject the request

3. Look for record containing APPLICATION_KEY, KEY_EPH_PUB+NONCE and IDENTIFIER in `pa_unique_values` table
   - If record doesn't exist, then continue with step 4.
   - If record exist and `timestamp_expire` is older than `CURRENT_TIMESTAMP`, then continue with step 4
   - If record exist and `timestamp_expire` is not older than `CURRENT_TIMESTAMP`, then reject the request

4. Create a new record in `pa_unique_values` (see Mapping for ECIES) and accept the request.

#### Validations for ECIES with temp keys (V3.3)

Let's define `EXPIRATION` as 60 seconds and define `IDENTIFIER` as `TEMPORARY_KEY_ID`, then

1. Test wether ECIES payload contains TIMESTAMP
   - If YES, then continue with step 2
   - If NO, then reject the request (timestamp is required for V3.2+ protocols)

2. Test whether TIMESTAMP is within range  `CURRENT_TIMESTAMP - EXPIRATION` < `TIMESTAMP` < `CURRENT_TIMESTAMP`.
   - If YES, then continue to 3.
   - If NO, then reject the request

3. Look for record containing APPLICATION_KEY, KEY_EPH_PUB+NONCE and IDENTIFIER in `pa_unique_values` table
   - If record doesn't exist, then continue with step 4.
   - If record exist and `timestamp_expire` is older than `CURRENT_TIMESTAMP`, then continue with step 4
   - If record exist and `timestamp_expire` is not older than `CURRENT_TIMESTAMP`, then reject the request

4. Create a new record in `pa_unique_values` (see Mapping for ECIES with temporary keys) and accept the request.

#### Validations for AEAD_V4

Let's define `EXPIRATION` as 60 seconds and define `IDENTIFIER` as `TEMPORARY_KEY_ID`, then

1. Test wether ECIES payload contains TIMESTAMP
   - If YES, then continue with step 2
   - If NO, then reject the request (timestamp is required for V3.2+ protocols)

2. Test whether TIMESTAMP is within range  `CURRENT_TIMESTAMP - EXPIRATION` < `TIMESTAMP` < `CURRENT_TIMESTAMP`.
   - If YES, then continue to 3.
   - If NO, then reject the request

3. Look for record containing APPLICATION_KEY, NONCE and IDENTIFIER in `pa_unique_values` table
   - If record doesn't exist, then continue with step 4.
   - If record exist and `timestamp_expire` is older than `CURRENT_TIMESTAMP`, then continue with step 4
   - If record exist and `timestamp_expire` is not older than `CURRENT_TIMESTAMP`, then reject the request

4. Create a new record in `pa_unique_values` (see Mapping for AEAD_V4) and accept the request.

### QA specification

Tested by automated tests.

romanstrobl:
Update of replay attack protection for v4 and fixes of old functionality to match the specification.


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/util/ReplayAttackUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        return param;
+    }
+
+    /**
+     * Derive unique value parameters based on protocol version for activation scope.
+     *
+     * @param protocolVersion Protocol version.
+     * @param temporaryKeyId Temporary key ID.
+     *
```
jnpsk: Missing params in javadoc

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
```diff
@@ -52,40 +51,71 @@ class DefaultReplayVerificationService implements ReplayVerificationService {
...
+     *
+     * <p>Version 3.2+:
+     * <ul>
+     * <li>If TIMESTAMP < CURRENT_TIMESTAMP - TIMESTAMP_THRESHOLD, then reject request</li>
+     * <li>If TIMESTAMP > CURRENT_TIMESTAMP + TIMESTAMP_THRESHOLD, then reject request</li>
+     * </ul>
+     *
+     * @param protocolVersion Protocol version.
+     * @param requestTimestamp Request timestamp.
+     * @throws GenericServiceException Thrown incase request is rejected due to its timestamp.
```
jnpsk: typo
```suggestion
     * @throws GenericServiceException Thrown in case request is rejected due to its timestamp.
```

jnpsk (approved):
looks good

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/ReplayPersistenceService.java
```diff
@@ -66,11 +67,19 @@ public boolean uniqueValueExists(final String uniqueValue) {
...
+     * @param protocolVersion Protocol version.
      * @param uniqueValue Unique value.
      * @return Whether unique value was added successfully.
      */
-    public boolean persistUniqueValue(final UniqueValueType type, final String uniqueValue) {
-        final Instant expiration = Instant.now().plus(powerAuthServiceConfiguration.getRequestExpiration());
+    public boolean persistUniqueValue(final UniqueValueType type, final String protocolVersion, final String uniqueValue) {
+        final Duration requestExpiration;
+        if ("3.0".equals(protocolVersion) || "3.1".equals(protocolVersion)) {
+            // MAC TOKEN only has extended expiration, ECIES is checked only in protocol versions 3.2+
```
jnpsk, line 77: What is this comment about? There cannot be a combination of ECIES and version 3.0 or 3.1? Because it seems the extended period should be applied to TOKEN type only.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
```diff
@@ -52,40 +51,71 @@ class DefaultReplayVerificationService implements ReplayVerificationService {
...
         }
+        logger.debug("Persisted unique value, request type: {}, request timestamp: {}, unique value: {}", param.getUniqueValueType(), requestTimestamp, uniqueValue);
+    }
+
+    /**
+     * Check timestamp using the following rules:
+     *
+     * <p>Version 3.0 and 3.1:
+     * <ul>
+     * <li>If TIMESTAMP < CURRENT_TIMESTAMP - EXPIRATION, then reject request</li>
```
jnpsk, line 89: Isn't this applied to TOKEN type only?

romanstrobl: Yes, it is impossible to call this method in `3.0` or `3.1` for ECIES, because such requests do not have timestamps, so the call is always skipped.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/ReplayPersistenceService.java
romanstrobl: Only MAC TOKEN type is verified in versions `3.0` and `3.1`, ECIES does not have timestamps in these versions, so the replay attacks cannot be checked.

PR #2008, opened 2025-07-15
Fix #1910: Implement activation confirmation endpoint

### LINKED ISSUE #1910: Implement activation confirmation endpoint

romanstrobl:
Implementation of activation confirmation on PowerAuth server.


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/controller/api/v4/ActivationController.java
```diff
@@ -133,6 +135,24 @@ public ObjectResponse<UpdateActivationOtpResponse> updateActivationOtp(@Valid @R
...
+     * @throws Exception In case the service throws exception.
+     */
+    @PostMapping("/confirm")
+    public Response confirmActivation(@Valid @RequestBody ObjectRequest<ConfirmActivationRequest> request) throws Exception {
+        final ConfirmActivationRequest req = request.getRequestObject();
+        logger.info("action: confirmActivation, state: initiated, activationId: {}", req.getActivationId());
+        logger.debug("action: confirmActivation, state: initiated, request: {}", request);
+        activationServiceBehavior.confirmActivation(req);
+        logger.info("action: confirmActivation, state: succeeded");
+        logger.debug("action: confirmActivation, state: succeeded, response: emtpy");
```
jnpsk: typo
```suggestion
        logger.debug("action: confirmActivation, state: succeeded, response: empty");
```

jnpsk (approved):
looks ok

PR #2009, opened 2025-07-16
Fix #2007: Improve default configuration of replay verification service

### LINKED ISSUE #2007: Improve default configuration of replay verification service
### Description

Following issues were found:
- the default threshold of 10 seconds is too low even with time synchronization, we should increase it
- the old protocol versions `3.0` and `3.1` have too strict validation for `limitNewest`, in case of desynchronized client the request will be rejected for case when client time is more than `replayTimestampThreshold` in the future

The issues can be worked around by setting property `powerauth.service.replay.timestamp.threshold` to a high value, but we should provide reasonable defaults for all protocol versions.

### Technical Specification

- Default expiration value is set to 5 minutes for protocol 3.2+ instead of 60 seconds.
- Default treshold value is set to 2 minutes for all protocol versions instead of 10 seconds.
- The `limitNewest` will be checked as: ```If `TIMESTAMP` > `CURRENT_TIMESTAMP` + `EXPIRATION`, then reject request``` (instead of using the threshold for `limitNewest`)

romanstrobl:
See: https://github.com/wultra/powerauth-server/issues/2007


romanstrobl:
Closing pull request, it will be handeled separately after merge of `features/crypto4` branch.

PR #2010, opened 2025-07-16
Fix #2007: Improve default configuration of replay verification service (crypto4)

### LINKED ISSUE #2007: Improve default configuration of replay verification service

romanstrobl:
See: https://github.com/wultra/powerauth-server/issues/2007

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
@@ -104,13 +104,15 @@ class DefaultReplayVerificationService implements ReplayVerificationService {
         final Instant now = Instant.now();
         final Instant requestTime = requestTimestamp.toInstant();
         final Instant limitOldest;
+        final Instant limitNewest;
         if ("3.0".equals(protocolVersion) || "3.1".equals(protocolVersion)) {
             // MAC TOKEN only has extended expiration, ECIES is checked only in protocol versions 3.2+
             limitOldest = now.minus(powerAuthServiceConfiguration.getRequestExpirationExtended());
+            limitNewest = now.plus(powerAuthServiceConfiguration.getRequestExpirationExtended());
         } else {
             limitOldest = now.minus(powerAuthServiceConfiguration.getReplayTimestampThreshold());
+            limitNewest = now.plus(powerAuthServiceConfiguration.getReplayTimestampThreshold());
         }
-        final Instant limitNewest = now.plus(powerAuthServiceConfiguration.getReplayTimestampThreshold());
         if (requestTime.isBefore(limitOldest) || requestTime.isAfter(limitNewest)) {
             // Rollback is not required, error occurs before writing to database
             logger.warn("Rejected request due to invalid timestamp: {}, allowed range: {} - {}, protocol version: {}", requestTime, limitOldest, limitNewest, protocolVersion);
--- a/powerauth-java-server/src/main/resources/application.properties
+++ b/powerauth-java-server/src/main/resources/application.properties
@@ -52,7 +52,7 @@ powerauth.service.crypto.activationValidityInMilliseconds=300000
 powerauth.service.crypto.authenticationCodeMaxFailedAttempts=5
 powerauth.service.crypto.authenticationCodeValidationLookahead=20
 powerauth.service.crypto.offlineAuthenticationCodeComponentLength=8
-powerauth.service.crypto.requestExpirationInMilliseconds=60000
+powerauth.service.crypto.requestExpirationInMilliseconds=300000
 powerauth.service.crypto.requestExpirationInMillisecondsExtended=7200000
 powerauth.service.crypto.replayVerificationService=default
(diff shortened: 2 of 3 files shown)


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
```diff
@@ -104,13 +104,15 @@ private void checkTimestamp(String protocolVersion, Date requestTimestamp) throw
...
         final Instant requestTime = requestTimestamp.toInstant();
         final Instant limitOldest;
+        final Instant limitNewest;
         if ("3.0".equals(protocolVersion) || "3.1".equals(protocolVersion)) {
             // MAC TOKEN only has extended expiration, ECIES is checked only in protocol versions 3.2+
             limitOldest = now.minus(powerAuthServiceConfiguration.getRequestExpirationExtended());
+            limitNewest = now.plus(powerAuthServiceConfiguration.getRequestExpirationExtended());
         } else {
             limitOldest = now.minus(powerAuthServiceConfiguration.getReplayTimestampThreshold());
+            limitNewest = now.plus(powerAuthServiceConfiguration.getReplayTimestampThreshold());
```
zcgandcomp, line 114: co can we drop the config `powerAuthServiceConfiguration.getRequestExpiration()` in favor of `powerAuthServiceConfiguration.getReplayTimestampThreshold() `?

zcgandcomp (changes requested):
Please reconsider the number of required settings.

Also for tokens there are additional thresholds:
powerauth.service.token.timestamp.validity
powerauth.service.token.timestamp.forward.validity

shouldn't we remove the redundant control to have single configuration?

powerauth-java-server/src/main/resources/application.properties
```diff
@@ -74,7 +74,7 @@ powerauth.service.token.timestamp.validity=7200000
 powerauth.service.token.timestamp.forward.validity=1800000
 
 # Replay Attach Check Time Synchronization Threshold
-powerauth.service.replay.timestamp.threshold=10000
+powerauth.service.replay.timestamp.threshold=120000
```
zcgandcomp, line 77: let's use value close to 5 minutes

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
romanstrobl: I let the decision to @hvge.

powerauth-java-server/src/main/resources/application.properties
romanstrobl: I let the decision to @hvge.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/replay/DefaultReplayVerificationService.java
hvge: Let's not over-configure this. There are basically four configuration values:

1. extended "expiration" value for legacy tokens for protocol versions 3.0 and 3.1 (no time synchronization available)
2. standard "expiration" value for ECIES 3.2+, Tokens 3.2+
3. threshold value
4. temporary key expiration

Configurations 1 and 2 affects for how long we keep the record in unique table. The 3rd., threshold value, affects how big time deviation from the client we still accept for the further processing. The threshold should be applied only for protocols with time synchronization. And finally, temporary key expiration must be equal or shorter to standard "expiration".

So, if you want to reduce the number of configurations, then only 2. and 4. can be joined to one value.

romanstrobl: I suggest we resolve this change in a separate ticket. The goal of this fix is to configure more reasonable default values, not change the number of configuration values.

powerauth-java-server/src/main/resources/application.properties
romanstrobl: We should revisit this value in the specification and then fix it. For now I suggest we keep it as is until we decide about different value.

romanstrobl:
Closing pull request, future changes will be done in `develop` branch after merge of `features/crypto4` branch.

PR #2027, opened 2025-07-29
Fix #2024: Unify token validity checks with replay attack time limit checks

### LINKED ISSUE #2024: Unify token validity checks with replay attack time limit checks
### Task specification

Token validity is checked separately from replay attack timestamp checks.

Only one time check is required, avoid duplicate checking with multiple parameters.

romanstrobl:
Removed duplicate token timestamp checks. The request timestamps are now only checked in replay verification service.

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
@@ -218,20 +218,6 @@ public class PowerAuthServiceConfiguration {
     private Duration httpMaxIdleTime;
 
     /**
-     * Token timestamp validity, checked before validating the token.
-     */
-    @Value("${powerauth.service.token.timestamp.validity}")
-    @DurationMin(millis = 1)
-    private Duration tokenTimestampValidity;
-
-    /**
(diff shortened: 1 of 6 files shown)


zcgandcomp (dismissed):
The configuration and the validation of the token validity are removed.
There is an underlying check of unique value that checks the token against replay attacks AND validates the validity timeframe.

Anything passing this validation has to pass the next one - so this check is redundant.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -217,20 +217,6 @@ public class PowerAuthServiceConfiguration {
     @Value("${powerauth.service.http.connection.max-idle-time}")
     private Duration httpMaxIdleTime;
 
-    /**
-     * Token timestamp validity, checked before validating the token.
-     */
-    @Value("${powerauth.service.token.timestamp.validity}")
```
jnpsk, line 223: Those conf. properties are still present in the `Configuration-Properties.md`

romanstrobl: Removed now.

jnpsk (dismissed):

jnpsk (approved):

PR #2028, opened 2025-07-29
Fix #2025: Unify temporary key expiration with replay attack timestamp checks

### LINKED ISSUE #2025: Unify temporary key expiration with replay attack timestamp checks
### Task specification

Currently temporary key expiration can be set in a way which is problematic in following context:

```
Keep in mind that the expiration of the temporary key used in ECIES or AEAD encryption schemes must match the expiration of E2EE-related records in pa_unique_values. If the temporary key remains valid beyond the duration defined by our `EXPIRATION` constant, it may allow for replay attacks.
```

We should unify the temporary key expiration in database with replay attack timestamp checks to avoid misconfiguration.

In case it is not logical or practical to unify these configuration values, we should add a startup assertion instead.

romanstrobl:
Added state validation for invalid temporary key and replay verification configuration.

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
@@ -712,8 +712,17 @@ public class PowerAuthServiceConfiguration {
     }
 
     @PostConstruct
-    void validate() {
+    void validateProximityConfiguration() {
         Assert.state(proximityCheckOtpLength >= MINIMAL_PROXIMITY_CHECK_OTP_LENGTH,
                 "Proximity check OTP length %d is smaller then required minimal %d".formatted(proximityCheckOtpLength, MINIMAL_PROXIMITY_CHECK_OTP_LENGTH));
     }
+
+    @PostConstruct
+    public void validateReplayConfiguration() {
+        Assert.state(temporaryKeyValidity.toMillis() <= requestExpiration.toMillis(),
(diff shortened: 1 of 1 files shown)


zcgandcomp (dismissed):
The validity of temporary keys has to be shorter than the validity of a request.

We need to keep the request validity timeframe larger - to ensure the unique value is still stored.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -712,8 +712,17 @@ public void setProximityCheckOtpLength(int proximityCheckOtpLength) {
...
 
     @PostConstruct
-    void validate() {
+    void validateProximityConfiguration() {
         Assert.state(proximityCheckOtpLength >= MINIMAL_PROXIMITY_CHECK_OTP_LENGTH,
                 "Proximity check OTP length %d is smaller then required minimal %d".formatted(proximityCheckOtpLength, MINIMAL_PROXIMITY_CHECK_OTP_LENGTH));
     }
+
+    @PostConstruct
+    public void validateReplayConfiguration() {
```
jnpsk: Public access to the method seems unnecessary

jnpsk (dismissed):
Looks good

jnpsk (approved):

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
romanstrobl: Fixed.

PR #790, opened 2022-12-09
Fix #789: Clean up maven configuration

banterCZ:

PR #2038, opened 2025-08-05
Fix #790: Add support for HmacSHA384

### LINKED ISSUE #2035: Switch algorithm HS256 to HS384 for temporary keys
### Task specification

Let's use `HS384` instead of `HS256` for temporary key signing.

romanstrobl:
Tests switched to `HS384`.

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/util/jwt/MACProviderHS384.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/util/jwt/MACProviderHS384.java
@@ -1,6 +1,6 @@
 /*
  * PowerAuth Server and related software components
- * Copyright (C) 2024 Wultra s.r.o.
+ * Copyright (C) 2025 Wultra s.r.o.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as published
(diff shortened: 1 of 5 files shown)


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/util/jwt/MACProviderHS384.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.crypto.spec.SecretKeySpec;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * MAC provider that enforces algorithm HS384.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public abstract class MACProviderHS384 extends BaseJWSProvider {
```
jnpsk, line 36: Is defining custom provider and verifier necessary? The default ones seem to support `HS384` out of the box.

romanstrobl: We wanted to enforce HS384. The default validator allows HS256, too, which we do not want to allow.

jnpsk: I think that could also be checked in the jws header before calling the verifier, but I get the point now.

jnpsk (approved):

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/util/jwt/MACProviderHS384.java
romanstrobl: Yes, also a possibility.

PR #2045, opened 2025-08-11
Change packages to com.wultra

romanstrobl:


pom.xml
```diff
@@ -24,7 +24,7 @@
     <name>powerauth-server-parent</name>
     <description>PowerAuth Server</description>
 
-    <groupId>io.getlime.security</groupId>
+    <groupId>com.wultra.security</groupId>
```
banterCZ, line 27: @jnpsk Don't we have to reconfigure Sonatype?

banterCZ (commented):
Great, but I believe that it should be mentioned in migration guides.

pom.xml
jnpsk: As we have both namespaces registered, I do not think any changes are required. The CI/CD pipeline does not seem to have some rules based on the groupId either.

banterCZ (changes requested):
I overlooked that we are changing the package in 1.10.x, so I do not agree with that.

romanstrobl:
Closing pull request, we can release with old package names.

PR #2061, opened 2025-08-25
Fix #2060: Allow getting temporary key in removed state

### LINKED ISSUE #2060: Crypto4: Allow temporary key for REMOVED state

_No response_

### Environment

_No response_

Allow getting the temporary key for activation scope also when activation is in REMOVED state. In the current implementation, client is not able to get information that activation is removed.

### Log output

_No response_

hvge:


jnpsk (approved):
looks ok

romanstrobl (approved):
Looks OK.

PR #2064, opened 2025-08-28
Fix #2062: Update cryptography service initialization

### LINKED ISSUE #2062: Update cryptography service initialization
### Task specification

We use `cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P384)` even for cases when the algorithm may be different, e.g.:

https://github.com/wultra/powerauth-server/blob/cfccfd9fe9c47e66d1c0eb487302b7999bd7f40b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AuthenticationSharedServiceBehavior.java#L220C47-L220C115

We should use the shared secret algorithm from the activation instead.

romanstrobl:
Algorithm used by the service should be taken from activation.


jnpsk (approved):
looks ok

PR #2067, opened 2025-09-01
Fix #2066: Update column definition in audit table

### LINKED ISSUE #2066: Update column definition in audit table
### Task specification

Use CLOB for large data in case of H2 initialization from source.

romanstrobl:


banterCZ (approved):
Thanks!

PR #2069, opened 2025-09-02
Fix #2068: Remove master public key from application detail

### LINKED ISSUE #2068: Remove master public key from application detail
### Task specification

The application detail endpoint returns a master public key. The same key is returned in SDK configuration as serialized value, we don't need it anymore.

romanstrobl:
The application detail endpoint is now versioned.


jnpsk (approved):
looks good

PR #2084, opened 2025-09-23
Fix #2083: Update operation endpoints due to rename of signature type to authentication code type

### LINKED ISSUE #2083: Update operation endpoints due to rename of signature type to authentication code type
### Task specification

Update operation and operation template endpoints to reflect the terminology changes in v4. The signature type enumeration was renamed to to authentication code type.

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -21,7 +21,7 @@
 
 import com.wultra.security.powerauth.app.server.service.callbacks.CallbackUrlEventService;
 import com.wultra.security.powerauth.app.server.service.behavior.tasks.ActivationServiceBehavior;
-import com.wultra.security.powerauth.app.server.service.behavior.tasks.OperationServiceBehavior;
+import com.wultra.security.powerauth.app.server.service.behavior.tasks.v4.OperationServiceBehavior;
```
jnpsk, line 24: I know in the background both v3 and v4 operations are expired, still it may be a bit confusing why only v4 service is called.
It does not have to be solved now, just sharing my view.

jnpsk (approved):
looks good

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/task/CleaningTask.java
romanstrobl: This is OK, we are migrating to the new API. V3 will be removed eventually.

PR #2088, opened 2025-09-26
Fix #2087: Fix randomly failing tests

### LINKED ISSUE #2087: Fix randomly failing tests
### Task specification

Some of the converter tests are failing randomly. This is caused by non-authenticated AES, which has small probability of succeeding on different inputs. We should fix these tests.

romanstrobl:


powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/converter/ActivationSharedSecretConverterTest.java
```diff
@@ -76,24 +75,4 @@ void testFromDbValueEncryption() throws Exception {
         assertEquals(SHARED_SECRET_BASE64, result);
     }
 
-    @Test
-    void testEncryptionAndDecryptionDifferentUserIdFail() throws Exception {
-        final byte[] sharedSecretBytes = Base64.getDecoder().decode(SHARED_SECRET_BASE64);
-        final SharedSecret sharedSecretEncrypted = sharedSecretConverter.toDBValue(sharedSecretBytes, USER_ID, ACTIVATION_ID);
-        assertEquals(EncryptionMode.AES_HMAC, sharedSecretEncrypted.encryptionMode());
-        assertThrows(GenericServiceException.class, () ->
```
jnpsk, line 84: alternatively, we can avoid assertThrows and just check that the result is not correct

jnpsk (dismissed):

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/converter/ActivationSharedSecretConverterTest.java
romanstrobl: That's a better idea, I will do that.

romanstrobl:
I added handling of both cases in the tests - either an exception is thrown or the decrypted value is different. The case depends on the internals of AES.

jnpsk (approved):

PR #2092, opened 2025-09-29
Fix #2091: Allow v3 token verification in v4 endpoint

### LINKED ISSUE #2091: Allow v3 token verification in v4 endpoint
### Task specification

For better compatibility with PowerAuth cloud it would be useful to allow verifying both v3 and v4 token versions in the v4 token verification endpoint based on parameter `protocolVersion`.

romanstrobl:
I allow v3 token verification in v4 endpoint based on `protocolVersion`, this is for using `v3` tokens with PowerAuth cloud without using the deprecated v3 client.


romanstrobl:
We will handle this logic on client side.

PR #2093, opened 2025-09-30
Fix #2089: Document application configuration for activation code

### LINKED ISSUE #2089: Document application configuration for activation code
### Task specification

Document configuration for `activation_transfer` in the Enrollment Server https://github.com/wultra/enrollment-server/issues/1295

banterCZ:


docs/Activation-Transfer.md
```diff
@@ -0,0 +1,34 @@
+# Activation Transfer
+
+PowerAuth server allows the transfer of the existing activation.
```
romanstrobl: Should be either `transfer of an existing activation` or `transfer of existing activations` (`the` refers to a specific activation).

docs/Activation-Transfer.md
```diff
@@ -0,0 +1,34 @@
...
+
+To enable activation transfer, an entry with the key `activation_transfer` must exist in the table `pa_application_config`.
+This configuration is exposed to the Enrollment Server, and it will manifest itself in the response [Activation Spawn endpoint](https://developers.wultra.com/components/enrollment-server/develop/documentation/Mobile-Token-API#activation-spawn), issuing an activation code.
+
+Mind that this table supports encryption, see [Encrypting Records in Database](./Encrypting-Records-in-Database.md) for details.
+
+
+### Required Configuration
+
+- `allowedTargetApplicationIds` - List of application IDs to which the activation is allowed to transfer.
```
romanstrobl: Should be probably `to which the activation is allowed to be transferred`.

romanstrobl (dismissed):
I found two minor English syntax issues.

docs/Activation-Transfer.md
banterCZ: Grammarly is saying that the clarity of `to be transferred` is worse than `to transfer`.

romanstrobl: 1. “…allowed to transfer”

This sounds like the activation itself is doing the transferring. Grammatically, it treats “activation” as the agent of the action. That works if you imagine an activation has the ability to “move” itself, but conceptually it’s a bit off if it’s really a passive object (like a license or registration).

2. “…allowed to be transferred”

This makes it clear that something else (the system, the user, the software) transfers the activation, and the activation is the object of that action. Since an activation can’t transfer itself, the passive form matches the real-world semantics more closely.

banterCZ: Thanks for clarifying. I will fix that.

romanstrobl (approved):

PR #2104, opened 2025-10-07
Fix #2098: Store parent activation ID

### LINKED ISSUE #2098: Store parent activation ID
### Task specification

For activation transfer, we need to store the parent activation ID. It should be possible to remove the parent activation when the child is activated.

Affects database structure and REST API.

banterCZ:


powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/request/InitActivationRequest.java
```diff
@@ -20,14 +20,13 @@
...
 import com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation;
 import com.wultra.security.powerauth.client.model.enumeration.ActivationProtocol;
+import com.wultra.security.powerauth.client.model.enumeration.ActivationTransferType;
 import com.wultra.security.powerauth.client.model.enumeration.CommitPhase;
 import io.swagger.v3.oas.annotations.media.Schema;
-import jakarta.validation.constraints.Future;
-import jakarta.validation.constraints.NotBlank;
-import jakarta.validation.constraints.NotNull;
-import jakarta.validation.constraints.Positive;
+import jakarta.validation.constraints.*;
```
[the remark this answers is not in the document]
banterCZ: It is the default IDE settings.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/listener/ActivationEntityListener.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Processes activation status changes after an update.
+     * If the activation status changes to ACTIVE and the activation is a result of a transfer operation type of MOVE,
+     * this method handles the removal of the parent activation.
+     * Parent removal is not performed if the previous activation status was BLOCKED.
+     *
+     * @param activation The activation record entity that was updated
+     */
+    @PostUpdate
+    public void processActivationStatusChange(final ActivationRecordEntity activation) {
```
banterCZ: I have been thinking about placing this logic into the service layer, but right now, there are three places where the activation status is changed to ACTIVE.

- `ActivationServiceBehavior`
- `v3/ActivationProcessServiceBehavior`
- `v4/ActivationProcessServiceBehavior`

Even when extracted to a common logic, it could be fragile for future changes when somebody calls just `Activation#setStatus()`. Now all status changes must go through this event listener.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/response/v4/GetActivationStatusResponse.java
```diff
@@ -20,10 +20,7 @@
 package com.wultra.security.powerauth.client.model.response.v4;
 
 import com.fasterxml.jackson.annotation.JsonRawValue;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationProtocol;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationStatus;
-import com.wultra.security.powerauth.client.model.enumeration.CommitPhase;
+import com.wultra.security.powerauth.client.model.enumeration.*;
```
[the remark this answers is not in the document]
banterCZ: This is default IDE settings.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/response/v3/GetActivationStatusResponse.java
```diff
@@ -20,10 +20,7 @@
 package com.wultra.security.powerauth.client.model.response.v3;
 
 import com.fasterxml.jackson.annotation.JsonRawValue;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationOtpValidation;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationProtocol;
-import com.wultra.security.powerauth.client.model.enumeration.ActivationStatus;
-import com.wultra.security.powerauth.client.model.enumeration.CommitPhase;
+import com.wultra.security.powerauth.client.model.enumeration.*;
```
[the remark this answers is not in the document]
banterCZ: This is default IDE settings.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/listener/ActivationEntityListener.java
banterCZ: 😞 This is a dead end, entity listeners are not supposed to be for complex modification, we are getting `ConcurrentModificationException` and so on.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -429,8 +430,7 @@ public CommitActivationResponse commitActivation(CommitActivationRequest request
                 throw localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_INCORRECT_STATE);
             }
 
-            // Change activation state to ACTIVE
-            activation.setActivationStatus(ActivationStatus.ACTIVE);
+            changeActivationStatusToActiveAndDeleteParent(activation);
```
banterCZ, line 447: I realize that we have a pattern of changing the status, record history, and notify callback. Ideally, it would be great to encapsulate that logic (something like `ActivationStatusService`) and hide `#setActivation`. But right now, there are about 21 occurrences, and I do not want to bloat this PR. I may suggest an improvement in another issue, if agreed.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/enumeration/ActivationTransferType.java
```diff
@@ -0,0 +1,28 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.wultra.security.powerauth.client.model.enumeration;
+
+/**
+ * Activation transfer type.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+public enum ActivationTransferType {
```
romanstrobl, line 25: Can we add JavaDoc for each activation transfer type?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
```diff
@@ -317,6 +322,16 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
         }
     }
 
+    private static ActivationTransferType convert(final com.wultra.security.powerauth.app.server.database.model.enumeration.ActivationTransferType source) {
```
romanstrobl: Can we call the `convert` method consistently with `convertProtocol` below?

romanstrobl (dismissed):
Nice work.

powerauth-client-model/src/main/java/com/wultra/security/powerauth/client/model/enumeration/ActivationTransferType.java
banterCZ: Sure, added.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
banterCZ: I like overloading, but you are right, since there already is such a name convention. Changed.

romanstrobl (approved):
Thank you for the updates.

PR #2123, opened 2025-10-24
Fix #2122: Activation transfer cannot be configured using REST API

### LINKED ISSUE #2122: Activation transfer cannot be configured using REST API

develop

### Environment

localhost

The configuration key `activation_transfer` needs to be allowed to be configured via REST API.

### Log output

_No response_

romanstrobl:


jnpsk (approved):
looks good

PR #2127, opened 2025-10-27
Fix #2126: Update offline signatures

### LINKED ISSUE #2126: Update offline signatures
### Task specification

Updated specification:

For the special purposes, when the activation scoped data is signed and presented as QR code, we can use `KEY_MAC_PERSONALIZED_DATA` for compute MAC:

```java
SecretKey KDK_UTILITY    = KDF.derive(KEY_ACTIVATION_SECRET, "util");
SecretKey KEY_MAC_PERSONALIZED_DATA = KDF.derive(KDK_UTILITY, "util/mac/personalized-data");
```

The key can be then used for compute the signature with using KMAC-256 algorithm:

```java
byte[] LABEL = ByteUtils.encode("PA4MAC-QR");
byte[] MAC = Mac.kmac256(KEY_MAC_PERSONALIZED_DATA, data, LABEL, 32);
```

If resulting `MAC` is used for [Off-line signature QR code](https://developers.wultra.com/components/enrollment-server/develop/documentation/Offline-Signatures-QR-Code) then its new "signing key type" identifier is `"2"` - `KEY_MAC_PERSONALIZED_DATA`. See the [current list of keys](https://developers.wultra.com/components/enrollment-server/develop/documentation/Offline-Signatures-QR-Code#signature).

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OfflineAuthenticationServiceBehavior.java
```diff
@@ -176,11 +173,11 @@ public CreatePersonalizedOfflineAuthPayloadResponse createPersonalizedOfflineAut
             final SecretKey keyMacPersonalisedData = KeyFactory.deriveKeyMacPersonalizedData(activationSecretKey);
 
             // Construct KMAC-256 tag
-            final byte[] tagKmac = Kmac.kmac256(keyMacPersonalisedData, kmacData, KMAC_JOSE_SIGNATURE_CUSTOM_BYTES, 64);
-            final String signature = createJwsJson(Base64URL.encode(tagKmac));
+            final byte[] tagKmac = Kmac.kmac256(keyMacPersonalisedData, kmacData, KMAC_OFFLINE_SIGNATURE_CUSTOM_BYTES, 32);
```
hvge, line 176: There's missing key identifier "2" as a part of  `kmacData`.  The comment few lines earlier is wrong:
```java
            // Compute KMAC-256 of '{DATA}\n{NONCE}
            // {DATA} consist of data from request plus optional generated proximity TOTP value
```
It should be
```java
            // Compute KMAC-256 of '{DATA}\n{NONCE}\n2
            // {DATA} consist of data from request plus optional generated proximity TOTP value
```

hvge (changes requested):
The routine for constructing data for MAC is wrong.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OfflineAuthenticationServiceBehavior.java
romanstrobl: Ah ok, I will update it.

hvge (approved):
Looks OK now. Thank you.

PR #2129, opened 2025-10-29
Fix #2113: Add support for EC_P384_ML_L5 algorithm

### LINKED ISSUE #2113: Add support for EC_P384_ML_L5 algorithm
### Task specification

Add support for algorithm suite `EC_P384_ML_L5` and store `ML-DSA-87` keys.

romanstrobl:


hvge (commented):
I think you forgot to update `v4/AsymmetricSignatureServiceBehavior`.

hvge:
Also `CryptographyServiceFactory` class creates the same `CryptographyServiceHybrid` for L3 and L5. The implementation of `CryptographyServiceHybrid ` is fishy in many places, because it uses hardcoded `KeyType.MLDSA_65` also for L5. I don't know whether it's correct or not, but it seems wrong.

Currently I have problem with activation fingerprint that doesn't match with mobile SDK.

romanstrobl:
> Also `CryptographyServiceFactory` class creates the same `CryptographyServiceHybrid` for L3 and L5. The implementation of `CryptographyServiceHybrid ` is fishy in many places, because it uses hardcoded `KeyType.MLDSA_65` also for L5. I don't know whether it's correct or not, but it seems wrong.
> 
> Currently I have problem with activation fingerprint that doesn't match with mobile SDK.

I see now. I will split the class into two to avoid such issues, the generalization did not work well.

romanstrobl:
Both issues should be fixed now.

hvge (dismissed):
I didn't review the code. My approval is based on passed integration tests from mobile SDK (PR: https://github.com/wultra/powerauth-mobile-sdk/issues/779)

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ApplicationDetailServiceBehavior.java
```diff
@@ -96,7 +96,7 @@ private GetApplicationDetailResponse createApplicationDetailResponse(Application
 
         final List<ApplicationVersionEntity> versions = applicationVersionRepository.findByApplicationId(applicationId);
         for (ApplicationVersionEntity version : versions) {
-            final SdkConfiguration sdkConfig = new SdkConfiguration(version.getApplicationKey(), version.getApplicationSecret(), masterKeyPairEntity.getMasterKeyPublicBase64(), null, null);
+            final SdkConfiguration sdkConfig = new SdkConfiguration(version.getApplicationKey(), version.getApplicationSecret(), masterKeyPairEntity.getMasterKeyPublicBase64(), null, null, null);
```
jnpsk: Nitpick - With this number of parameters, some of which are null, it might be worth switching to builder.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ApplicationDetailServiceBehavior.java
```diff
@@ -125,14 +126,24 @@ private GetApplicationDetailResponse createApplicationDetailResponse(Application
...
+            publicKeyMlDsa87 = publicKeyRegistry.getPublicKey(KeyType.MLDSA_87)
+                    .map(publicKey -> {
+                        try {
+                            byte[] bytes = KEY_CONVERTOR_PQC_DSA.convertPublicKeyToBytes(publicKey);
+                            return Base64.getEncoder().encodeToString(bytes);
+                        } catch (GenericCryptoException e) {
+                            logger.warn("Public key conversion failed", e);
+                            return null;
+                        }
+                    }).orElse(null);
```
jnpsk: Nitpick - it might be beneficial for code readability to process this in a dedicated function, as it is repeated 3 times and bit overwhelming due to the nested try-catch blocks

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/BiometryServiceBehavior.java
```diff
@@ -141,12 +143,16 @@ public AddBiometryResponse addBiometry(AddBiometryRequest request) throws Generi
...
+                case EC_P384_ML_L3, EC_P384_ML_L5 -> {
                     try {
                         final SharedSecretRequestHybrid sharedSecretRequestHybrid = new SharedSecretRequestHybrid();
                         sharedSecretRequestHybrid.setEcClientPublicKey(sharedSecretRequest.getEcdhe());
                         sharedSecretRequestHybrid.setPqcEncapsulationKey(sharedSecretRequest.getMlkem());
-                        final ResponseCryptogram responseCryptogram = SHARED_SECRET_HYBRID.generateResponseCryptogram(sharedSecretRequestHybrid);
+                        final ResponseCryptogram responseCryptogram = switch (algorithm) {
+                            case EC_P384_ML_L3 -> SHARED_SECRET_HYBRID_ML_L3.generateResponseCryptogram(sharedSecretRequestHybrid);
+                            case EC_P384_ML_L5 -> SHARED_SECRET_HYBRID_ML_L5.generateResponseCryptogram(sharedSecretRequestHybrid);
+                            default -> null;
```
jnpsk: Nitpick - It seems the default branch cannot actually happen, still it may be more reasonable to throw an exception.

jnpsk (dismissed):
Looks good!
Just having few nitpicks regarding code readability to consider.

jnpsk (approved):
looks good!

PR #2133, opened 2025-11-05
Fix #1820: Update status endpoint for crypto4

### LINKED ISSUE #1820: Update status endpoint for crypto4
### Description

The activation status blob is enhanced with an additional fields in protocol 4 and protected with KMAC.

### Acceptance criteria

The activation status blog is update to version 4.

## Status blob

The activation status blob is enhanced with an additional fields in protocol 4 and protected with KMAC:

- `STATUS_DATA`:
  ```
  4B:0xDEC0DED4
  1B:${STATUS}
  1B:${CURRENT_VERSION}
  1B:${UPGRADE_VERSION}
  1B:${STATUS_FLAGS}
  4B:${RESERVED}
  1B:${CTR_BYTE}
  1B:${FAIL_COUNT}
  1B:${MAX_FAIL_COUNT}
  1B:${CTR_LOOK_AHEAD}
  32B:${CTR_DATA_HASH}
  ```

- `STATUS_MAC`:
  ```java
    byte[] STATUS_MAC = Mac.kmac256(KEY_MAC_STATUS, STATUS_DATA, "PA4MAC-STATUS");
  ```

- Final binary status blob returned from PowerAuth Server:
  ```java
  byte[] BINARY_STATUS_BLOB = ByteUtils.concat(STATUS_DATA, STATUS_MAC);
  ```

### Notable changes

- MAGIC value is now `0xDEC0DED4` instead of `0xDEC0DED1`
- `${STATUS_FLAGS}` byte contains the following flags:
  - bit 0: `STATUS_FLAG_ACTIVATION_CONFIRMATION`: indicate the pending activation confirmation.
  - bit 1: `STATUS_FLAG_UPGRADE_CONFIRMATION`: indicate the pending upgrade confirmation.
  - bit 2: `STATUS_FLAG_UNSUPPORTED_ALGORITHM`: indicate that activation is using no longer supported encryption algorithm. The upgrade process is recommended.
  - bit 7: `STATUS_FLAG_BIOMETRY_FACTOR_ON`: indicate that biometry factor is turned on on the server

- `${CTR_DATA_HASH}` is now 32 bytes long, instead of 16 and calculated as:
  ```java
  byte[] CTR_DATA_HASH = Mac.kmac256(CTR_DATA, KEY_MAC_CTR_DATA, "PA4MAC-CTR");
  ```

Note that `CTR_DATA_HASH` is always calculated with KMAC, even if the activation is not yet migrated to protocol V4. 

### QA specification

Tested by integration tests.

jnpsk:
Allow upgrade procedure to protocol V4

### DIFF
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
@@ -65,7 +65,7 @@ public class ActivationStatusServiceBehavior {
     /**
      * Current PowerAuth protocol major version. Activations created with lower version will be upgraded to this version.
      */
-    private static final byte POWERAUTH_PROTOCOL_VERSION = 0x3;
+    private static final byte POWERAUTH_PROTOCOL_VERSION = 0x4;
 
     private final ActivationRemoveServiceBehavior activationRemoveServiceBehavior;
     private final LocalizationProvider localizationProvider;
--- a/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
+++ b/powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
@@ -62,11 +62,6 @@ import java.util.*;
 @AllArgsConstructor
 public class ActivationStatusServiceBehavior {
 
-    /**
-     * Current PowerAuth protocol major version. Activations created with lower version will be upgraded to this version.
-     */
-    private static final byte POWERAUTH_PROTOCOL_VERSION = 0x4;
-
     private final ActivationRemoveServiceBehavior activationRemoveServiceBehavior;
     private final LocalizationProvider localizationProvider;
     private final PowerAuthServiceConfiguration powerAuthServiceConfiguration;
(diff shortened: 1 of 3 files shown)


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
```diff
@@ -65,7 +65,7 @@ public class ActivationStatusServiceBehavior {
     /**
      * Current PowerAuth protocol major version. Activations created with lower version will be upgraded to this version.
      */
-    private static final byte POWERAUTH_PROTOCOL_VERSION = 0x3;
+    private static final byte POWERAUTH_PROTOCOL_VERSION = 0x4;
```
jnpsk: Otherwise V3 activation status won't show there is a protocol upgrade available.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationStatusServiceBehavior.java
```diff
@@ -231,13 +231,7 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                             throw localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_INCORRECT_STATE);
                         }
                         // Assign the activation fingerprint
-                        if (activation.getVersion() == 4) {
-                            activationFingerPrint = cryptographyServiceFactory.getService(sharedSecretAlgorithm).generateActivationFingerprint(activation);
-                        } else {
-                            logger.error("Unsupported activation version: {}", activation.getVersion());
-                            // Rollback is not required, database is not used for writing
-                            throw localizationProvider.buildExceptionForCode(ServiceError.ACTIVATION_INCORRECT_STATE);
-                        }
```
jnpsk, line 240: During protocol upgrade, we may need to ask for activation status after start, but before confirm. Meaning the SDK asks for v4 status, but server fails on this condition, as the protocol version is switched after confirm on server.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
banterCZ: 🤔 At first glance, it confused me that in the v3 controller, we have `0x4`. Sure, the value makes sense, but it could be handy to refactor it to avoid a similar discrepancy in the future.

romanstrobl (dismissed):
Good catch!

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
jnpsk: good point, we may make it a global setting

romanstrobl (approved):
Looks OK.

banterCZ (approved):
Thanks for extracting the protocol version.

PR #2134, opened 2025-11-06
Fix #1817: Allow configuring supported algorithms for crypto4

### LINKED ISSUE #1817: Allow configuring supported algorithms for crypto4
### Description

Add configuration for specifying supported algorithms for crypto4. The configuration should be possible per application. This setting will impact which algorithms will be used for cryptography and affects performance.

Furthermore, validate that algorithms sent in requests from the client match server configuration and estabilished algorithm stored in database during activation.

romanstrobl:
Configuration of supported algorithms and dynamic generating / regenerating of master keypairs based on configuration, update of status and application detail endpoints.

Additionally, minimum supported version of cryptography protocol can be configured to disable the legacy v3 version.


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterKeyGenerationService.java
```diff
@@ -85,16 +87,18 @@ public void generateMasterKeyPairs(ApplicationEntity application) throws Generic
                 publicKeyRegistry = new PublicKeyRegistry();
                 writeRegistriesToEntity(keyPair, privateKeyRegistry, publicKeyRegistry, application);
```
jnpsk: Looks like empty v4 registries are already initialized during the `createInitialV3Entity` call.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmQueryService.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Get the list of supported shared secret algorithms for an application.
+     * @param application Application.
+     * @return List of supported shared secret algorithms.
+     */
+    public List<SharedSecretAlgorithm> getSupportedAlgorithms(ApplicationEntity application) {
+        return Stream.of(SharedSecretAlgorithm.values())
+                .filter(algorithm -> isAlgorithmSupported(application, algorithm))
+                .toList();
```
jnpsk, line 108: This includes one DB call per each existing `SharedSecretAlgorithm`. I suggest to process it in a single run instead.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmQueryService.java
```diff
@@ -0,0 +1,99 @@
...
+     * Get whether shared secret algorithm is allowed based on configuration of allowed algorithms.
+     * @param application Application.
+     * @param sharedSecretAlgorithm Shared secret algorithm.
+     * @return Whether shared secret algorithm is allowed.
+     */
+    public boolean isAlgorithmSupported(ApplicationEntity application, SharedSecretAlgorithm sharedSecretAlgorithm) {
+        GetApplicationConfigRequest configRequest = new GetApplicationConfigRequest();
+        configRequest.setApplicationId(application.getId());
+
+        GetApplicationConfigResponse configResponse = applicationConfigServiceBehavior.getApplicationConfig(configRequest);
```
jnpsk: I suggest to move this call after those static checks bellow to avoid unnecessary DB calls.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterKeyGenerationService.java
```diff
@@ -125,11 +125,24 @@ private MasterKeyPairEntity createInitialV3Entity(ApplicationEntity app) throws
         entity.setMasterPrivateKeys(masterPrivateKeys.privateKeysBase64());
         entity.setMasterPrivateKeysEncryption(masterPrivateKeys.encryptionMode());
         entity.setMasterPublicKeys(publicKeysConverter.toDBValue(publicKeyRegistry));
+        return entity;
+    }
 
-        entity.setMasterKeyPrivateBase64(Base64.getEncoder().encodeToString(KEY_CONVERTOR_EC.convertPrivateKeyToBytes(privateKey)));
-        entity.setMasterKeyPublicBase64(Base64.getEncoder().encodeToString(KEY_CONVERTOR_EC.convertPublicKeyToBytes(EcCurve.P256, publicKey)));
+    private void generateAndStoreV3KeyPair(MasterKeyPairEntity keyPairEntity, List<SharedSecretAlgorithm> supportedAlgorithms) throws CryptoProviderException {
```
jnpsk: I suggest to make the naming more explicit as the generation and storing is only conditional. Maybe also of the public method.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/UpgradeServiceBehavior.java
```diff
@@ -112,6 +114,9 @@ public StartUpgradeResponse startUpgrade(StartUpgradeRequest request) throws Gen
             final EncryptionContext context = new EncryptionContext(protocolVersion, applicationKey, null, EncryptorId.UPGRADE_START);
             final DecryptionResult decryptionResult = encryptionService.decryptRequest(encryptedRequest, context);
             final SharedSecretRequestPayload requestPayload = parseRequestPayload(decryptionResult.getDecryptedData(), activation.getActivationId());
+            final SharedSecretAlgorithm algorithm = SharedSecretAlgorithm.valueOf(requestPayload.getSharedSecretRequest().getAlgorithm());
+
+            algorithmValidationService.validateAlgorithmForActivation(activation, algorithm);
```
jnpsk: This validation does not seem to be correct, as the `algorithm` inside the `SharedSecretRequest` represents the target algorithm.
Same applies for the validation in the `SharedSecretServiceBehavior#deriveSharedSecret`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/SharedSecretServiceBehavior.java
```diff
@@ -100,6 +102,8 @@ public SharedSecretResponsePayload deriveSharedSecret(ActivationRecordEntity act
 
         final SharedSecretAlgorithm algorithm = SharedSecretAlgorithm.valueOf(sharedSecretRequest.getAlgorithm());
 
+        algorithmValidationService.validateAlgorithmForActivation(activation, algorithm);
```
jnpsk: There should be some bypass for upgrade process.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmValidationService.java
```diff
@@ -0,0 +1,87 @@
...
+        validateAlgorithmForApplication(activation.getApplication(), algorithm);
+
+        if (!algorithmMatchesActivation(activation, algorithm)) {
+            logger.warn("Cryptography algorithm does not match activation, activation ID: {}, algorithm: {}", activation.getActivationId(), algorithm);
+            throw localizationProvider.buildExceptionForCode(ServiceError.CRYPTOGRAPHY_ALGORITHM_NOT_SUPPORTED);
+        }
+    }
+
+    /**
+     * Get whether shared secret algorithm matches the algorithm stored during activation.
```
jnpsk:
```suggestion
     * Check whether shared secret algorithm matches the algorithm stored during activation.
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OfflineAuthenticationServiceBehavior.java
```diff
@@ -253,6 +253,11 @@ public CreateNonPersonalizedOfflineAuthPayloadResponse createNonPersonalizedOffl
             // Compute ECDSA signature of '{DATA}\n{NONCE}\n{KEY_MASTER_SERVER_PRIVATE_INDICATOR}'
             final byte[] signatureBase = (data + "\n" + nonce + "\n" + KEY_MASTER_SERVER_PRIVATE_INDICATOR).getBytes(StandardCharsets.UTF_8);
             final byte[] ecdsaSignatureBytes = cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P384).generateSignatureForApplication(KeyType.ECDSA_P384, signatureBase, application);
+            if (ecdsaSignatureBytes == null) {
+                logger.error("Missing EC_P384 master key pair for application ID: {}", applicationId);
+                // Rollback is not required, database is not used for writing
+                throw localizationProvider.buildExceptionForCode(ServiceError.NO_MASTER_SERVER_KEYPAIR);
+            }
```
jnpsk: I am not a huge fan of having such a specific and nonobvious meaning for a returned null value. At least, I would add it to the method's documentation.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
```diff
@@ -148,7 +148,7 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                     response.setEncryptedStatusBlob(Base64.getEncoder().encodeToString(randomStatusBlob));
                     response.setEncryptedStatusBlobNonce(randomStatusBlobNonce);
                     response.setActivationCode(activation.getActivationCode());
-                    response.setActivationSignature(Base64.getEncoder().encodeToString(activationSignature));
+                    response.setActivationSignature(activationSignature != null ? Base64.getEncoder().encodeToString(activationSignature) : null);
```
jnpsk: Isn't that an odd case, that v3 activations status is requested, but application key is not available?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationProcessServiceBehavior.java
```diff
@@ -79,7 +81,23 @@ public AeadEncryptedResponse processNewActivation(ActivationRecordEntity activat
...
+            }
+
+            // Check that cryptography algorithm is allowed
+            final SharedSecretAlgorithm algorithm = SharedSecretAlgorithm.valueOf(layer2Request.getSharedSecretRequest().getAlgorithm());
+
+            try {
+                algorithmValidationService.validateAlgorithmForApplication(activation.getApplication(), algorithm);
+            } catch (GenericServiceException e) {
+                // Activation failed due to unsupported algorithm, rollback transaction
+                throw localizationProvider.buildRollbackingExceptionForCode(ServiceError.CRYPTOGRAPHY_ALGORITHM_NOT_SUPPORTED);
```
jnpsk, line 100: This exact exception seems to be already thrown in the called `validateAlgorithmForApplication`  method.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationConfigServiceBehavior.java
```diff
@@ -152,12 +150,25 @@ public void removeApplicationConfig(final RemoveApplicationConfigRequest request
             }
             final List<ApplicationConfigEntity> configs = applicationConfigRepository.findByApplicationId(applicationId);
             configs.stream().filter(config -> config.getKey().equals(key)).forEach(applicationConfigRepository::delete);
+            processApplicationConfigChange(appOptional.get(), key);
```
jnpsk: Nitpick - It would be nice to call it iff a deletion actually happened.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmQueryService.java
```diff
@@ -0,0 +1,99 @@
...
+                .filter(config -> CONFIG_KEY_CRYPTO_SUPPORTED_ALGORITHMS.equals(config.getKey()))
+                .findFirst();
+
+        // All other algorithms are allowed by default
+        if (allowedAlgorithmsConfig.isEmpty() || allowedAlgorithmsConfig.get().getValues().isEmpty()) {
+            return true;
+        }
+
+        final ApplicationConfigurationItem configurationItem = allowedAlgorithmsConfig.get();
+        return configurationItem.getValues().contains(sharedSecretAlgorithm.name());
```
jnpsk: 🤔 So with no config , all algorithms (except nonhybrid PQC) are supported by default (effectively a black-list). But once a config is nonempty, it starts to work as white-list of algorithms?

jnpsk (commented):
Nice idea!
- It looks like each algorithm validation results in a database fetch, which is unavoidable when dynamic configuration of allowed algorithms is supported. IMO, we should be careful with such validations in that case. There are cases in the code, when the validation is called multiple times during a single request processing. This is more about optimization and code architecture. 
- More importantly, I noticed the target algorithm is validated against current algorithm of an activation in the upgrade flow, which have to be fixed.
- Less importantly I am wondering if a black-list wouldn't be more suitable for our use case.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationProcessServiceBehavior.java
romanstrobl: The difference is that this exeption is rollbacking, we have to do that for new activations.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmQueryService.java
romanstrobl: Yes.

romanstrobl:
I do not think a blacklist is a good idea, because we will be adding new algorithms in the future. Thus, added algorithms would be enabled automatically once the configuration is done. I think this behaviour is undesirable.

So either:
- there is no algorithm configuration for an app, all algorithms are supported (open mode)
- there is an algorithm configuration for an app, in this case only configured algorithms are enforced (strict mode)

I think blackisting would create a scenario which is too dynamic for the future.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterKeyGenerationService.java
romanstrobl: Thanks, I overlooked that, will remove it.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterKeyGenerationService.java
romanstrobl: What name do you suggest?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/UpgradeServiceBehavior.java
romanstrobl: This validation is incorrect, indeed. I will remove it.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/SharedSecretServiceBehavior.java
romanstrobl: I have removed the validation.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OfflineAuthenticationServiceBehavior.java
romanstrobl: I will move the error handling, so `null` value is avoided.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ActivationStatusServiceBehavior.java
romanstrobl: I will update the logic, an exception will be thrown instead.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationConfigServiceBehavior.java
romanstrobl: Good idea.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/AlgorithmQueryService.java
romanstrobl: Refactored, thanks for pointing it out.

romanstrobl:
I revised the configuration to be more logical:
- no configuration per app - default list of algorithms is supported: `EC_P256`, `EC_P384`, `EC_P384_ML_L3`, `EC_P384_ML_L5`
- empty configuration per app - no algorithm is supported
- non-empty configuration per app - configured list of algorithms is supported

I think this is the best way moving forward, once we add new algorithms there will be no surprise:
- default supported algorithm list is extended
- existing configurations are preserved and do not add algorithms automatically

jnpsk (approved):
looks ok

PR #2141, opened 2025-11-14
Fix #2140: Regenerate create_schema.sql scripts 

### LINKED ISSUE #2140: Add missing SQL statements to create_schema scripts
### Task specification

Some statements from the `migration_1.10.0_2.0.0.sql` are missing in the `create_schema.sql` - definitely for Postgres (check also scripts for other supported DBs).

pavsindelar:
Regenerated `create_schema.sql` scripts for all supported DBs using Liquibase.

### DIFF
--- a/docs/sql/mssql/create_schema.sql
+++ b/docs/sql/mssql/create_schema.sql
@@ -2,9 +2,9 @@
 -- Update Database Script
 -- *********************************************************************
 -- Change Log: ./docs/db/changelog/changesets/powerauth-java-server/db.changelog-module.xml
--- Ran at: 10.03.25 7:07
--- Against: null@offline:mssql
--- Liquibase version: 4.31.1
+-- Ran at: 11/14/25, 11:24 AM
+-- Against: null@offline:mssql?outputLiquibaseSql=none
+-- Liquibase version: 4.33.0
 -- *********************************************************************
 
 -- Changeset powerauth-java-server/1.4.x/20230322-audit.xml::1::Lubos Racansky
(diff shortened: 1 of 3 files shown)


docs/sql/mssql/create_schema.sql
```diff
@@ -114,7 +114,7 @@ GO
...
...ation_status int NOT NULL, activation_otp varchar(255), activation_otp_validation int CONSTRAINT DF_pa_activation_activation_otp_validation DEFAULT 0 NOT NULL, blocked_reason varchar(255), counter int NOT NULL, ctr_data varchar(255), device_public_key_base64 varchar(255), extras varchar(255), platform varchar(255), device_info varchar(255), flags varchar(255), failed_attempts int NOT NULL, max_failed_attempts int CONSTRAINT DF_pa_activation_max_failed_attempts DEFAULT 5 NOT NULL, server_private_key_base64 varchar(255) NOT NULL, server_private_key_encryption int CONSTRAINT DF_pa_activation_server_private_key_encryption DEFAULT 0 NOT NULL, server_public_key_base64 varchar(255) NOT NULL, timestamp_activation_expire datetime2(6) NOT NULL, timestamp_created datetime2(6) NOT NULL, timestamp_last_used datetime2(6) NOT NULL, timestamp_last_change datetime2(6), master_keypair_id int, version int CONSTRAINT DF_pa_activation_version DEFAULT 2, CONSTRAINT PK_PA_ACTIVATION PRIMARY KEY (activation_id), CONSTRAINT activation_keypair_fk FOREIGN KEY (master_keypair_id) REFERENCES pa_master_keypair(id), CONSTRAINT activation_application_fk FOREIGN KEY (application_id) REFERENCES pa_application(id));
```
pavsindelar, line 117: Not sure we want to modify even these "old" statements (there are 1-2 in all three files).

banterCZ: 🤔 And what is the difference?

banterCZ (approved):
Thanks for adding the missing changes.

docs/sql/mssql/create_schema.sql
pavsindelar: There is actually no difference, just reversed order of constraint definitions (in all cases), so I assume it is ok to merge as is (no functional impact). Maybe it was caused by a different Liquibase version.

banterCZ: Indeed, according to the metadata, it has been generated by 4.31.1
Let's observer the future changes.

PR #2143, opened 2025-11-20
Fix #2075: Update database encryption for crypto4

### LINKED ISSUE #2075: Update database encryption for crypto4
### Description

New algorithm should be added in line with crypto4 algorithms (AES-256, KMAC-256).

### Acceptance criteria

Data in database is encrypted using stronger encryption algorithms.

### Technical specification

Aead.seal(SecretKey encryptionKey, byte[] keyContext, byte[] nonce, byte[] associatedData, byte[] plaintext)

byte[] index = String.join("&", encryptionKeyProvider.get()).getBytes(StandardCharsets.UTF_8);
encryptionKey = KMAC-256(masterDbEncryptionKey, index, "PA4DBENC", 32);
keyContext = null
nonce = null
associatedData = "$table.$column".getBytes();

### QA specification

Tested by unit tests.

romanstrobl:
Implementation of AEAD_KMAC DB encryption algorithm, which is now enabled by default.


docs/PowerAuth-Server-2.0.0.md
```diff
@@ -1,5 +1,66 @@
...
+
+PowerAuth Server 2.0.0 introduces a new default database encryption algorithm used for encrypting sensitive values stored in the database, such as master private keys, activation private keys, temporary keys, callback authentication, and other sensitive data. This encryption is done by the PowerAuth server application and it should be used on top of database encryption at rest to avoid leakage of sensitive data.
+
+## New Default Encryption Algorithm: `AEAD_KMAC`
+
+The algorithm `AEAD_KMAC` is now the default database encryption algorithm. It uses:
+
+* a 256-bit master DB encryption key  
+* KMAC-256 for per-record key derivation  
+* AES in AEAD mode for authenticated encryption
```
jnpsk: "AES in AEAD mode" does not sound like a good wording. I would use something like authenticated encryption using AES or write the actual mode of the AES.

docs/PowerAuth-Server-2.0.0.md
```diff
@@ -1,5 +1,66 @@
...
+
+```
+powerauth.server.db.master.encryption.algorithm=NO_ENCRYPTION
+```
+
+When using `NO_ENCRYPTION`, no database encryption keys are required, and sensitive values will be stored in plaintext.
+
+### Securing the Encryption Key
+
+The master database encryption key is a highly sensitive secret and must be protected. Do not store this key directly in configuration files, source control, or container images. Use a secure secret management solution (such as a vault or encrypted configuration provider) to supply the key to the application at runtime and restrict access only to the PowerAuth Server process.
```
jnpsk: I would move this higher, above the `NO_ENCRYPTION` variant explanation.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -687,4 +702,34 @@ void validateReplayConfiguration() {
         );
     }
 
+    @PostConstruct
+    void validateDbEncryptionConfiguration() {
+        if (dbEncryptionAlgorithm != EncryptionAlgorithm.AEAD_KMAC) {
+            return;
+        }
```
jnpsk: Nitpick - we may also react to other choices with appropriate warning message (not for production use, or legacy algorithm).

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/DefaultEncryptionKeySupplier.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Default encryption key supplier for database per-record encryption.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class DefaultEncryptionKeySupplier implements EncryptionKeySupplier{
+
+    private final Supplier<List<String>> keyDerivationData;
+    private final Supplier<List<String>> associatedData;
```
jnpsk: What is the benefit of keeping that as a Supplier?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/DatabaseEncryptionFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    private final AesHmacEncryptionService aesHmacService;
+    private final AeadKmacEncryptionService aeadKmacService;
+
+    private Map<EncryptionAlgorithm, DatabaseEncryptor> encryptorMap() {
+        return Map.of(
+                EncryptionAlgorithm.NO_ENCRYPTION, noEncryptionService,
+                EncryptionAlgorithm.AES_HMAC, aesHmacService,
+                EncryptionAlgorithm.AEAD_KMAC, aeadKmacService
+        );
+    }
```
jnpsk: Nitpick - what about to build the Map once on the construction and store the services there?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/AeadKmacEncryptionService.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    @Override
+    public byte[] decrypt(byte[] encryptedData, EncryptionKeySupplier encryptionKeySupplier) throws GenericServiceException {
+        try {
+            final SecretKey masterKey = loadMasterDbEncryptionKey();
+            final SecretKey encryptionKey = deriveKmacKey(masterKey, encryptionKeySupplier.keyDerivationData());
+            final byte[] ad = deriveAssociatedData(encryptionKeySupplier.associatedData());
+            return Aead.open(encryptionKey, null, ad, encryptedData);
```
jnpsk, line 79: It is not critical in this case, but I am wondering if adding a domain separation string instead of null wouldn't be more solid.

jnpsk (dismissed):
Looks good!

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/AeadKmacEncryptionService.java
romanstrobl: We already use:
- derived key from related data (binding for the key)
- additional data for mac verification (binding for the mac)

We cannot separate by protocol version, because the encryption needs to work accross multiple protocol versions. What kind of domain separation do you have in mind?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/DatabaseEncryptionFactory.java
romanstrobl: We can do that, however the services are autowired, so it cannot be static.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/DefaultEncryptionKeySupplier.java
romanstrobl: We can keep as a List, it was just how I refactored it.

docs/PowerAuth-Server-2.0.0.md
romanstrobl: Ok I will update it.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/encryption/AeadKmacEncryptionService.java
romanstrobl: Note that the Kmac is already used for domain separation for the key, so I am not sure we need any additional domain separation here.

jnpsk: ok, understand

jnpsk (approved):
Nice!

PR #2148, opened 2025-11-20
Fix #2142: Update pom.xml to skip publishing and add central publishing plugin

### LINKED ISSUE #2142: Automatic publish in Sonatype
### Task specification

_No response_

filipbursik-wultra:


pom.xml
```diff
@@ -226,6 +227,11 @@
     <build>
         <pluginManagement>
             <plugins>
+                <plugin>
```
banterCZ, line 230: 🤔 I see the only `pluginManagement` section. Is the parent pom published?

filipbursik-wultra: It's fixed.

banterCZ (approved):

PR #2151, opened 2025-11-21
Fix #2146: Improve crypto agility for signatures in init activation response

### LINKED ISSUE #2146: Improve crypto agility for signatures in init activation response
### Task specification

Currently, the new signatures are returned as fields. They should be returned as a map to allow easier extendability for the future.

Same change should be applied to activation status. Furthermore, supported algorithms should be checked when generating signatures.

romanstrobl:


jnpsk (approved):
Looks OK

Issue #2152, opened 2025-11-21
Add Docker Image Build


filipbursik-wultra:
### Task specification

Set up Docker image building in the project as part of the CI process.

PR #2165, opened 2025-11-26
Make dev profile working

banterCZ:
- Configure localhost datasource
- Add AEAD_KMAC for dev runner configuration

A follow-up to #2152

### DIFF
--- a/powerauth-java-server/src/main/resources/application-dev.properties
+++ b/powerauth-java-server/src/main/resources/application-dev.properties
@@ -6,7 +6,5 @@ spring.liquibase.change-log=classpath:db/changelog/db.changelog-module.xml
 
 powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction=60000
 
-powerauth.server.db.master.encryption.algorithm=NO_ENCRYPTION
-
 logging.level.com.wultra=DEBUG
 logging.level.com.wultra.core.audit=INFO
--- a/powerauth-java-server/src/main/resources/application-dev.properties
+++ b/powerauth-java-server/src/main/resources/application-dev.properties
@@ -1,7 +1,12 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/powerauth
+
 # Liquibase
 spring.liquibase.enabled=true
 spring.liquibase.change-log=classpath:db/changelog/db.changelog-module.xml
 
 powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction=60000
 
+powerauth.server.db.master.encryption.algorithm=NO_ENCRYPTION
+
 logging.level.com.wultra=DEBUG
+logging.level.com.wultra.core.audit=INFO
--- a/.run/PowerAuthServerApplication.run.xml
+++ b/.run/PowerAuthServerApplication.run.xml
@@ -2,6 +2,9 @@
   <configuration default="false" name="PowerAuthServerApplication" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
     <option name="ACTIVE_PROFILES" value="dev" />
     <option name="DEBUG_MODE" value="true" />
+    <envs>
+      <env name="POWERAUTH_SERVER_DB_MASTER_ENCRYPTION_KEY_AEAD_KMAC" value="NlcNpEU4l79W2qc95FQfuxR10lm7uPV171vuFcblrNA=" />
+    </envs>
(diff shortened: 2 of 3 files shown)


powerauth-java-server/src/main/resources/application-dev.properties
```diff
@@ -1,7 +1,12 @@
...
 # Liquibase
 spring.liquibase.enabled=true
 spring.liquibase.change-log=classpath:db/changelog/db.changelog-module.xml
 
 powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction=60000
 
+powerauth.server.db.master.encryption.algorithm=NO_ENCRYPTION
+
 logging.level.com.wultra=DEBUG
+logging.level.com.wultra.core.audit=INFO
```
michal-rozehnal-w, line 10: Out of curiosity, without this line, is it disabled by default or is different level used?

powerauth-java-server/src/main/resources/application-dev.properties
```diff
@@ -1,7 +1,12 @@
+spring.datasource.url=jdbc:postgresql://localhost:5432/powerauth
+
 # Liquibase
 spring.liquibase.enabled=true
 spring.liquibase.change-log=classpath:db/changelog/db.changelog-module.xml
 
 powerauth.service.scheduled.job.fido2AuthenticatorCacheEviction=60000
 
+powerauth.server.db.master.encryption.algorithm=NO_ENCRYPTION
```
banterCZ: 🤔 @romanstrobl suggests adding the key rather than turning the encryption off. I just wanted to avoid possible misuse of the dev key in production.

michal-rozehnal-w (dismissed):

powerauth-java-server/src/main/resources/application-dev.properties
banterCZ: In that case, _audit_ is logged as DEBUG, because of the setting for `com.wultra` above, and it is quite noisy.

powerauth-java-server/src/main/resources/application-dev.properties
michal-rozehnal-w: Can we use the same value as in the [documentation](https://developers.wultra.com/components/powerauth-server/develop/documentation/Encrypting-Records-in-Database#enabling-record-encryption)? The `dev` profile should be used only for development on a local machine with dummy data—nothing sensitive—and it shouldn’t affect production properties.

romanstrobl: We want to enforce usages of encryption for PowerAuth 2.0. We can omit this for development but then we will miss a chance to directly test this feature. I leave the decision to you, but personally I prefer to turn encryption on to be close to production functionality.

romanstrobl: Note that the documentation intentially forces clients to configure an encryption key because security model of crypto4 has changed and we store many additional keys and shared secrets that we could derive previously but due to nature of KEMs we cannot do that anymore.

banterCZ: Thank you for noticing. I will reconfigure that.

romanstrobl: I understand the concern of checking in the key into source core but it can also be supplied using a property externally so that is another option to use without adding the key here.

romanstrobl: For development I think you can generate an all-zero 256 bit key and use it, just supply externally when starting the app? Not sure what works for you. If its complicated for you, we can use NO_ENCRYPTION. For me its not a problem because I develop on Tomcat to stay close to production which has proven many times as useful because cli startup behaves slightly differently.

banterCZ: 🤔 The best way is probably setting the env in `./run/PowerAuthServerApplication.run.xml`. It will work for developers, but will not be tempting if not seen in the properties, which are part of the final war.

romanstrobl: I will update all documentation soon.

romanstrobl: For now, see: https://github.com/wultra/powerauth-server/blob/develop/docs/PowerAuth-Server-2.0.0.md#update-database-encryption-configuration

michal-rozehnal-w (approved):

PR #2210, opened 2025-12-29
Fix #2209: Premature version increase for PowerAuth crypto

### LINKED ISSUE #2209: Premature version increase for PowerAuth crypto

develop

### Environment

localhost

The `powerauth-java-crypto` version should be `2.0.0` for the next release, not `2.0.1-SNAPSHOT`.

romanstrobl:
Reverted premature version increase for PA crypto.


romanstrobl:
The pull request is invalid, the version was correct, just not published publicly.

PR #2212, opened 2026-01-06
Fix #2204: Optimize key storage in database

### LINKED ISSUE #2204: Optimize key storage in database
### Task specification

The goal is to optimize the database structure for key-related data. The new DB structure should consist of three tables with the following schema:
- id (PK, number)
- key (text)

With a reference column in pa_activation

jnpsk:


docs/db/changelog/changesets/powerauth-java-server/2.0.x/20250314-crypto4-pqc.xml
```diff
@@ -18,22 +18,12 @@
     <changeSet id="2" logicalFilePath="powerauth-java-server/2.0.x/20250314-crypto4-pqc.xml" author="Roman Strobl">
         <preConditions onFail="MARK_RAN">
             <not>
-                <columnExists tableName="pa_activation" columnName="device_public_keys" />
```
romanstrobl, line 21: Note that we are changing existing liquibase change sets. They have been already applied in some testing environments, and thus the database migration needs to be fixed there manually.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20250314-crypto4-pqc.xml
```diff
@@ -18,22 +18,12 @@
...
-                <columnExists tableName="pa_activation" columnName="server_private_keys" />
-                <columnExists tableName="pa_activation" columnName="server_private_keys_encryption" />
-                <columnExists tableName="pa_activation" columnName="server_public_keys" />
                 <columnExists tableName="pa_activation" columnName="shared_secret" />
                 <columnExists tableName="pa_activation" columnName="shared_secret_encryption" />
             </not>
         </preConditions>
         <comment>Add new columns for crypto4 keys to pa_activation table</comment>
         <addColumn tableName="pa_activation">
-            <column name="device_public_keys" type="clob"/>
```
banterCZ, line 31: The checksum has changed. What shall we do with environments where the changeset was already applied?

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20251222-crypto4-keys-tables.xml
```diff
@@ -0,0 +1,106 @@
...
+
+    <changeSet id="3" logicalFilePath="powerauth-java-server/2.0.x/20251222-crypto4-keys-tables.xml" author="Jan Pesek">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="pa_server_public_key"/>
+            </not>
+        </preConditions>
+        <comment>Create a new table pa_server_public_key</comment>
+        <createTable tableName="pa_server_public_key">
+            <column name="id" type="bigint">
```
banterCZ: Recently, I have discovered the parameter `defaultValueSequenceNext`. But in that case, the sequence has to be declared beforehand.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/ServerPrivateKeyEntity.java
```diff
@@ -0,0 +1,75 @@
...
+    private Long id;
+
+    /**
+     * Server private keys for V4 cryptography algorithms serialized into JSON.
+     */
+    @Column(name = "key_data", columnDefinition = "CLOB")
+    private String keyData;
+
+    /**
+     * Application level encryption algorithm {@code (0 = NO_ENCRYPTION, 1 = AES_HMAC)}.
```
banterCZ, line 52: 🤔 In general, I prefer `EnumType#STRING`. Have we chosen `ORDINAL` for compatibility reasons?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AsymmetricSignatureServiceBehavior.java
```diff
@@ -66,6 +66,7 @@ public class AsymmetricSignatureServiceBehavior {
      * @return ECDSA signature for provided data payload.
      * @throws GenericServiceException In case signature verification fails.
      */
+    @Transactional
```
banterCZ: Could it be read-only, or is there a side-effect?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/JwtSignatureServiceBehavior.java
```diff
@@ -67,6 +67,7 @@ public class JwtSignatureServiceBehavior {
      * @return Sign JWT response.
      * @throws GenericServiceException In case of invalid request or signature calculation failure.
      */
+    @Transactional
```
banterCZ: ditto

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/KeyProvider.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Retrieve server private keys associated with the given activation.
+     *
+     * @param activation The activation which server private keys to retrieve.
+     * @return An {@link Optional} containing the server private key registry, or empty if none exists.
+     * @throws GenericServiceException If the conversion from database representation fails.
+     */
+    public Optional<PrivateKeyRegistry> getServerPrivateKeys(final ActivationRecordEntity activation) throws GenericServiceException {
+        final ServerPrivateKeyEntity entity = activation.getServerPrivateKey();
```
banterCZ, line 68: 😬 I hate checked exceptions. Without that, we could nicely refactor to

```java
return Optional.ofNullable(activation.getServerPrivateKey())
        .map(entity -> new PrivateKeysRecord(entity.getEncryptionAlgorithm(), entity.getKeyData()))
        .map(record -> serverPrivateKeysConverter.fromDBValue(record, activation.getUserId(), activation.getActivationId()));
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/KeyProvider.java
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * Stores a device public key for the given activation.
+     *
+     * @param activation The activation to which the device public key should be attached.
+     * @param keyType The type of the public key to store.
+     * @param publicKey The public key to store.
+     * @throws GenericServiceException If the conversion to database representation fails.
+     */
+    public void storeDevicePublicKey(final ActivationRecordEntity activation, final KeyType keyType, final PublicKey publicKey) throws GenericServiceException {
+        DevicePublicKeyEntity entity = activation.getDevicePublicKey();
```
banterCZ: In general, I would rather see `final` and avoid reassigning. I realize that this is short and simple, but we never know how the code evolves.

banterCZ (dismissed):
Thanks, looks good. My biggest concern is about the changed liquibase checksum and what to do with our environments.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AsymmetricSignatureServiceBehavior.java
romanstrobl: Good catch!

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/JwtSignatureServiceBehavior.java
romanstrobl: Good catch!

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20250314-crypto4-pqc.xml
jnpsk: When I run it locally I updated the `databasechangelog` table with the new checksum.
Although inconvenient, I preferred merging the changes instead of having opposing changesets, as it is not released version.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/ServerPrivateKeyEntity.java
jnpsk: Yes, as the `pa_activation` columns uses integer representation (for both existing and new columns), I kept that for consistency

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20251222-crypto4-keys-tables.xml
jnpsk: Is there any benefit as it is handled by Hibernate?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/KeyProvider.java
jnpsk: Good point!
Agree that your code is more convenient. Still this feels more like conceptual approach of the project which I would avoid solving in this issue 😄

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/KeyProvider.java
jnpsk: Sure, I can do it that way

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AsymmetricSignatureServiceBehavior.java
jnpsk: Looks like it can be set as read-only

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/JwtSignatureServiceBehavior.java
jnpsk: Looks like it can be set as read-only

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20251222-crypto4-keys-tables.xml
banterCZ: Just for manual insertion or easier test scripts.

romanstrobl (dismissed):
Looks good. Let's run backend end-to-end tests to confirm that the change has no side effects.

banterCZ (approved):

PR #2215, opened 2026-01-08
Fix #2204: Optimize key storage in database

### LINKED ISSUE #2204: Optimize key storage in database

jnpsk:
Fix persistence order.
On parallel requests sometimes the activation got an update before the associated key was persisted leading to FK constraint violation.


romanstrobl (approved):
Looks OK.

PR #2216, opened 2026-01-08
Fix #2206: Store the fingerprint in the pa_activation table

### LINKED ISSUE #2206: Store the fingerprint in the pa_activation table
### Task specification

To optimize data load, store the calculated device fingerprint and avoid future loading of key and counting on fly for each get status.

jnpsk:


romanstrobl (approved):
Great!

PR #2222, opened 2026-01-14
Fix #2221: Unify application detail endpoints

### LINKED ISSUE #2221: Unify application detail endpoints
### Task specification

The `v3` application detail endpoint should be able to return `v4` keys, so that clients who retrieve SDK configuration via `v3` REST API obtain all new keys for configuration of the SDK. The logic should be the same for both endpoints versions.

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ApplicationDetailServiceBehavior.java
```diff
@@ -100,6 +114,40 @@ private GetApplicationDetailResponse createApplicationDetailResponse(Application
             masterPublicKeyBase64 = null;
         }
 
+        final List<SharedSecretAlgorithm> supportedAlgorithms = algorithmQueryService.getSupportedAlgorithms(application);
```
jnpsk, line 110: I suggest to use this `supportedAlgorithms` list also for the `EC_P256` support decision above, to save a database query.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ApplicationDetailServiceBehavior.java
```diff
@@ -110,7 +158,10 @@ private GetApplicationDetailResponse createApplicationDetailResponse(Application
             final SdkConfiguration sdkConfig = SdkConfiguration.builder()
                     .appKey(version.getApplicationKey())
                     .appSecret(version.getApplicationSecret())
-                    .masterPublicKeyP256(masterPublicKeyBase64)
+                    .masterPublicKeyP256(masterKeyPairEntity.getMasterKeyPublicBase64())
```
jnpsk: This results in the `EC_P256` key being added regardless of whether it is supported, which differs from the behavior of the other keys and from the top level content of the `response.masterPublicKey`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ApplicationDetailServiceBehavior.java
romanstrobl: Good idea.

jnpsk (approved):
Looks good

PR #2224, opened 2026-01-15
Fix #2223: Activation fingerprint handling during server upgrade

### LINKED ISSUE #2223: Activation fingerprint handling during server upgrade
### Task specification

For v4 the activation fingerprint is stored precalculated, however for v3 activations it is not. When the server gets upgrade, the v4 endpoint returns empty fingerprint for v3 activations before they are expired, although a non-null fingerprint should be returned.

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationStatusServiceBehavior.java
```diff
@@ -213,6 +215,15 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                         statusBlob = null;
                     }
 
+                    final String activationFingerprint;
+                    if (activation.getActivationFingerprint() != null) {
+                        // Activation fingerprint is pre-calculated on PA server 2.x.x
+                        activationFingerprint = activation.getActivationFingerprint();
+                    } else {
+                        // Activation was created on a previous PA server version which does not store the fingerprint
+                        activationFingerprint = cryptographyServiceFactory.getService(activation.getCryptoAlgorithm()).generateActivationFingerprint(activation);
```
jnpsk: I don’t think this change fixes the issue. The `crypto_algorithm` column was introduced in 2.0 and is not populated for V3 activations (and it appears it’s not set even for new V3 activations). Meaning the factory will throw an exception for such cases every time.
I think we have to explicitly handle V3 activations without stored fingerprint, unhappily by hardcoding the legacy algorithm for the service factory.

romanstrobl: Ah yes, you're right. I will hardcode it then for v3.

jnpsk (dismissed):
Looks good

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ActivationStatusServiceBehavior.java
```diff
@@ -213,6 +215,15 @@ public GetActivationStatusResponse getActivationStatus(GetActivationStatusReques
                         statusBlob = null;
                     }
 
+                    final String activationFingerprint;
+                    if (activation.getActivationFingerprint() != null) {
+                        // Activation fingerprint is pre-calculated on PA server 2.x.x
+                        activationFingerprint = activation.getActivationFingerprint();
+                    } else {
+                        // Activation was created on a previous PA server version (v3) which does not store the fingerprint
+                        activationFingerprint = cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P256).generateActivationFingerprint(activation);
```
zcgandcomp: So to have it completely foolproof should not we have something like:

```
if (activation.getCryptoAlgorithm() == null ) {
                        activationFingerprint = cryptographyServiceFactory.getService(SharedSecretAlgorithm.EC_P256).generateActivationFingerprint(activation);
} else {
                        activationFingerprint = cryptographyServiceFactory.getService(activation.getCryptoAlgorithm()).generateActivationFingerprint(activation);
}
```

jnpsk: I believe the current code assumes either that a fingerprint is present in the database or that this is a V3 activation, so this change should not affect the current behavior.

Potential advantage of this additional change is future-proofing and using `EC_P256` algorithm as a fallback feels slightly more generic than treating the V3 fingerprint as a fallback for a missing fingerprint.

romanstrobl: Ok, I will add it, althrough I don't think this case is currently possible, but it's good future-proofing.

romanstrobl:
We should merge this issue after release because of https://github.com/wultra/powerauth-server/issues/2225

jnpsk (approved):

PR #2226, opened 2026-01-19
Fix #2225: Add information about missing fingerprint for v4 REST API into migration guide

### LINKED ISSUE #2225: Add information about missing fingerprint for v4 REST API into migration guide
### Task specification

In case that an activation is created using the `v3` REST API before upgrade to the `2.0.x` release, the activation fingerprint is not stored in the database. In this case the activation status in `v4` REST API returns an empty fingerprint.

We are handling this corner case in issue https://github.com/wultra/powerauth-server/issues/2223, however until this is resolve we should document this problem in the migration guide for the `2.0.0` release.

romanstrobl:


docs/PowerAuth-Server-2.0.0.md
```diff
@@ -179,6 +179,14 @@ We have unified validations in PowerAuth server REST API. The error code returne
 
 The validation of requests is now stricter and more complete to ensure data integrity. In case you get the `ERR0024` error in your integration with PowerAuth server, please make sure the requests contain all parameters, as seen in REST API documentation available at `http[s]://[hostname]:[port]/powerauth-java-server/swagger-ui/index.html`.
 
+### Missing Activation Fingerprint for v3 Activations Queried via v4 Status Endpoint
+
+In case an activation was created using the `v3` REST API before upgrading to the `2.0.x` release, the activation fingerprint is not stored in the database.
```
jnpsk: This seems like too specific information to start with.

docs/PowerAuth-Server-2.0.0.md
```diff
@@ -179,6 +179,14 @@ We have unified validations in PowerAuth server REST API. The error code returne
...
+
+As a result, when the activation status is requested using `v4` REST API, the response contains a null fingerprint value. For `v3` activations use the `v3` REST API. The issue is only temporary during upgrade until all `v3` activations are committed or expired.
```
jnpsk: I am not sure I understand the last sentence.
🤔  Isn't the status filtering applied by the PAC only?

docs/PowerAuth-Server-2.0.0.md
zcgandcomp: If an activation was created using the `v3` REST API before upgrading to `2.0.x`, the activation fingerprint is unavailable and returned as `null`. If you need the fingerprint, use the `v3` REST API. Internal components (PowerAuth Restul Integration and Enrollment Server) automatically select the correct API version.

romanstrobl: Do you want me to replace my text with your text?

docs/PowerAuth-Server-2.0.0.md
romanstrobl: This is not about status filtering. After upgrade, existing activations will expire. For new activations the fingerprint is stored correctly even for `v3`. Thus the issue is temporary, e.g. for 5 minutes max, depending on the expiration period.

romanstrobl: Anyway I will remove it to avoid confusion.

jnpsk (dismissed):

docs/PowerAuth-Server-2.0.0.md
```diff
@@ -179,6 +179,14 @@ We have unified validations in PowerAuth server REST API. The error code returne
 
 The validation of requests is now stricter and more complete to ensure data integrity. In case you get the `ERR0024` error in your integration with PowerAuth server, please make sure the requests contain all parameters, as seen in REST API documentation available at `http[s]://[hostname]:[port]/powerauth-java-server/swagger-ui/index.html`.
 
+### Missing Activation Fingerprint for v3 Activations Queried via v4 Status Endpoint
+
+If an activation was created using the `v3` REST API before upgrading to the `2.0.x` release, the activation fingerprint is unavailable (not stored in the database) and will be returned as `null` when queried via the `v4` REST API.
+
+If you need the fingerprint for such activations, use the `v3` REST API. The choice of REST API version is done correctly by applications which use the PowerAuth Restful Integration.
```
zcgandcomp: Enrollment Server uses Restful Integration, but this is not in the documentation. It would be best to reference it explicitly.

romanstrobl: Ok. I will add this information as well.

jnpsk (approved):

zcgandcomp (approved):
Ok

PR #2239, opened 2026-02-05
Fix #2238: Null value in parameters in operation create causes an exception

### LINKED ISSUE #2238: Null value in parameters in operation create causes an exception

1.8.0 and higher

### Environment

localhost

The escaping of characters for parameters with `null` values causes an exception:

```
2026-02-04 14:02:09,179 INFO --- [http-nio-127.0.0.1-8970-exec-6] i.g.s.p.a.s.c.api.OperationsController : action: createOperation, state: initiated, userId: 1524, applications: [net.getik.mxvista], templateName: ibOperation2Params
2026-02-04 14:02:09,179 ERROR --- [http-nio-127.0.0.1-8970-exec-6] i.g.s.p.a.s.s.b.t.OperationServiceBehavior : Runtime exception or error occurred, transaction will be rolled back

java.lang.NullPointerException: null

2026-02-04 14:02:09,180 ERROR --- [http-nio-127.0.0.1-8970-exec-6] o.s.b.w.servlet.support.ErrorPageFilter : Forwarding to error page from request [/rest/v3/operation/create] due to exception [null]

java.lang.NullPointerException: null
```

### Log output

_No response_

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
jnpsk: Does it really fix the issue? It seems `Collectors#toMap` also requires non nulls.

romanstrobl: Hmm I see. We will need to work around that, too.

romanstrobl: Let's check now.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OperationServiceBehaviorTest.java
```diff
@@ -784,6 +784,30 @@ void testOperationRejectFailedDifferentUserId() throws GenericServiceException {
...
+        OperationDetailResponse response = assertDoesNotThrow(() ->
+                operationService.createOperation(request)
+        );
+        assertNotNull(response);
+        assertNotNull(response.getParameters());
+
+        assertEquals("testValue1", response.getParameters().get("testKey1"));
+        assertTrue(response.getParameters().containsKey("testKey2"));
+        assertNull(response.getParameters().get("testKey2"));
+    }
```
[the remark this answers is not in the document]
romanstrobl: I will add the test case.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehaviorTest.java
```diff
@@ -784,6 +784,30 @@ void testOperationRejectFailedDifferentUserId() throws GenericServiceException {
...
+        OperationDetailResponse response = assertDoesNotThrow(() ->
+                operationService.createOperation(request)
+        );
+        assertNotNull(response);
+        assertNotNull(response.getParameters());
+
+        assertEquals("testValue1", response.getParameters().get("testKey1"));
+        assertTrue(response.getParameters().containsKey("testKey2"));
+        assertNull(response.getParameters().get("testKey2"));
+    }
```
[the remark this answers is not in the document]
romanstrobl: I will add the test case.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
banterCZ: @jnpsk What was the issue? It seems that `Collectors#toMap` internally uses `HashMap`.

romanstrobl: It does not work, I tried:

```
java.lang.NullPointerException: null
	at java.base/java.util.Objects.requireNonNull(Objects.java:208) ~[na:na]
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:180) ~[na:na]
	at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[na:na]
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133) ~[na:na]
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[na:na]
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682) ~[na:na]
	at com.wultra.security.powerauth.app.server.service.behavior.tasks.v3.OperationServiceBehavior.escapeParameters(OperationServiceBehavior.java:253) ~[classes/:na]
	at com.wultra.security.powerauth.app.server.service.behavior.tasks.v3.OperationServiceBehavior.createOperation(OperationServiceBehavior.java:187) ~[classes/:na]
```

jnpsk: Still it throws NPE when value is null. There is a t[hread on Stack Overflow](https://stackoverflow.com/questions/24630963/nullpointerexception-in-collectors-tomap-with-null-entry-values) regarding that.
And the [implementation](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/Collectors.java#L1442-L1449) seems to actually require non null.

banterCZ: Ok, just nitpicking. I see

```
BiConsumer<Map<K, V>, T> uniqKeysMapAccumulator(Function<? super T, ? extends K> keyMapper,
                                                    Function<? super T, ? extends V> valueMapper) {
        return (map, element) -> {
            K k = keyMapper.apply(element);
            V v = Objects.requireNonNull(valueMapper.apply(element));
```

So `.filter(e -> e.getValue() != null)` would fix it as well. But you may keep it. I was just curious.

romanstrobl: No we don't want to filter values, because they would be removed.

romanstrobl: I believe we want to keep the non-null keys with null values, right?

jnpsk (approved):

banterCZ (approved):
Thanks for the fix.

PR #2259, opened 2026-03-02
Fix #2258: Coverity: Incorrect definition of Serializable class

### LINKED ISSUE #2258: Coverity: Incorrect definition of Serializable class
### Task specification

501696 [Se: Incorrect definition of Serializable class](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_FB)
This Serializable class defines a non-primitive instance field which is neither transient, Serializable, or java.lang.Object, and does not appear to implement the Externalizable interface or the readObject() and writeObject() methods. ([From SpotBugs description](https://scan4.scan.coverity.com/doc/en/spotbugs/sb_checker_ref.html#FB.SE_BAD_FIELD))

banterCZ:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/ServerPrivateKeyEntity.java
```diff
@@ -34,7 +36,10 @@
 @Entity
 @Table(name = "pa_server_private_key")
 @Getter @Setter
-public class ServerPrivateKeyEntity {
+public class ServerPrivateKeyEntity implements Serializable {
```
banterCZ, line 39: I would like to get rid of `Serializable` completely, but if this class is referenced from the serializable class, we have to do this.

jnpsk (approved):

PR #2283, opened 2026-03-13
Resolves #2173: Add mobileSdkConfig to CreateApplicationResponse

### LINKED ISSUE #2173: Created application response should return mobile SDK configuration
### Task specification

vita-kotacka:
## Summary

Adds `mobileSdkConfig` (via a `versions` list) to the `CreateApplicationResponse`, so callers receive SDK configuration immediately after creating an application — without a separate `getApplicationDetail` call.

### `powerauth-client-model`
- Added `List<ApplicationVersion> versions` field to `CreateApplicationResponse`

### `powerauth-java-server`
- **`MasterKeyGenerationService`**: changed `generateMasterKeyPairs` return type from `void` to `MasterKeyPairEntity` (returns the saved entity, no extra DB round-trip)
- **`ApplicationServiceBehavior`**: refactored `createApplication` to compute public keys (P256, P384, ML-DSA-65, ML-DSA-87) from the returned master key pair and build `mobileSdkConfig` — following the same pattern as `ApplicationDetailServiceBehavior`

### Tests added
- `ApplicationControllerTest` (v3 and v4)
- `ApplicationServiceBehaviorTest` (11 tests for `createApplication`)
- `ApplicationDetailServiceBehaviorTest` (v3: 12 tests, v4: 12 tests for `getApplicationDetail`)
- `MasterKeyGenerationServiceTest` (9 tests for `generateMasterKeyPairs`)


powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/controller/api/v3/ApplicationControllerTest.java
```diff
@@ -0,0 +1,118 @@
...
+    void createApplication_shouldReturnWrappedResponse_whenServiceSucceeds() throws Exception {
+        final CreateApplicationRequest req = new CreateApplicationRequest();
+        req.setApplicationId("app-1");
+        final ObjectRequest<CreateApplicationRequest> request = new ObjectRequest<>();
+        request.setRequestObject(req);
+
+        final CreateApplicationResponse serviceResponse = new CreateApplicationResponse();
+        when(applicationService.createApplication(req)).thenReturn(serviceResponse);
+
+        final ObjectResponse<CreateApplicationResponse> response = applicationController.createApplication(request);
```
banterCZ, line 59: 🤔 I am just surprised by testing the controller via Java API; it is quite unusual. I suggest testing the layer below or using `MockMvc` instead.

banterCZ (commented):
I let the review of keys to Roman.
Just a short note, it's a great idea to have tests for controllers, but in that case, I suggest using `MockMvc` to test JSON mapping, exception handlers, etc.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
```diff
@@ -140,24 +161,66 @@ public CreateApplicationResponse createApplication(CreateApplicationRequest requ
             application = applicationRepository.save(application);
 
             // Generate master server key pairs
-            masterKeyGenerationService.generateMasterKeyPairs(application);
+            final MasterKeyPairEntity masterKeyPair = masterKeyGenerationService.generateMasterKeyPairs(application);
 
             // Use cryptography methods before writing to database to avoid rollbacks
+            final List<SharedSecretAlgorithm> supportedAlgorithms = algorithmQueryService.getSupportedAlgorithms(application);
```
romanstrobl, line 158: Lines 167 - 189 are duplicated in `ApplicationServiceBehaviour` and `ApplicationDetailServiceBehaviour`. Is there any reason why we cannot reuse the same logic? Either using a shared service method, or a utility class? It seems that too much logic is the same accross these two classes.

romanstrobl (changes requested):
I would like to review a large duplicate code fragment, maybe we can improve it?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
banterCZ: I agree with Roman that we should reuse logic, not using inheritance, but delegation to a shared component.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/controller/api/v3/ApplicationControllerTest.java
vita-kotacka: Is _MockMvc_ something which is broadly used in our components? I don't see any such usage in the PowerAuth compoonent. 🤔

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
romanstrobl: On the first sight, the logic in `ApplicationDetailServiceBehaviour` is quite easy to reuse with minimal refactoring of method `createApplicationDetailResponse`, it can be public and reused, I think.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/controller/api/v3/ApplicationControllerTest.java
banterCZ: You are right, we do not use it much. Historically, there is a technical debt of lower test coverage. The end-to-end tests are placed in a separate repository https://github.com/wultra/powerauth-tests
🤔 I see that elsewhere, it is also used `WebTestClient`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterPublicKeyService.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Holds the four possible master public keys extracted from a {@link MasterKeyPairEntity}.
+     *
+     * @param p256    Base64-encoded EC P-256 master public key, or {@code null} when not supported.
+     * @param p384    Base64-encoded EC P-384 master public key, or {@code null} when not supported.
+     * @param mlDsa65 Base64-encoded ML-DSA-65 master public key, or {@code null} when not supported.
+     * @param mlDsa87 Base64-encoded ML-DSA-87 master public key, or {@code null} when not supported.
+     */
+    public record MasterPublicKeys(String p256, String p384, String mlDsa65, String mlDsa87) {}
```
romanstrobl: It is not a good practice to define records inside a service. Please define the record in a separate file, so we don't have to use e.g. `MasterPublicKeyService.MasterPublicKeys`. It would be OK to define the record within the service if it was only used there, but it's used in other classes, too.

romanstrobl (changes requested):
The changes are almost perfect now. I found one more issue I think we should resolve around best practices of defining records.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterPublicKeyService.java
banterCZ: To be honest, I did it several times (defining a record in the service). It is up to you here, but we may open this topic at tech grooming.

romanstrobl: Here are my reasons:
- the type definition is more verbose (e.g. `MasterPublicKeyService.MasterPublicKeys`)
- it creates an unnecessary coupling to the service
- it makes the type feel like an implementation detail of the service
- potential issues with serialization

I think that private records should stay within the service class, public records should be defined separately for reasons above.

romanstrobl: I also find it confusing that we have `MasterPublicKeys` defined within a service but `PrivateKeysRecord`, `ServerPrivateKeyRecord`, and `SharedSecretRecord` are defined in separate files.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/controller/api/v3/ApplicationControllerTest.java
vita-kotacka: I would keep the test as it is for now. And once we know the way we want to approach it, there can be a refactoring ticket.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ApplicationServiceBehavior.java
vita-kotacka: Refactored. ✅

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/crypto/MasterPublicKeyService.java
vita-kotacka: Refactored. ✅

romanstrobl (approved):
Good job!

PR #2286, opened 2026-03-16
Resolves #2004 Optimize activation code generation

### LINKED ISSUE #2004: Optimize Activation Code generation
### Description

In class  `ActivationInitServiceBehavior` method `InitActivation`, there is a uniqueness control based on like select from pa_activation.
The control relies on select statement with like operation. This is not optimal from a performance and safety point of view. 

### Acceptance criteria

The uniqueness is ensured with reasonable limits.
No side effects.
The performance characteristics are better than the current solution.

### Technical specification

A possibility to add a unique index and flush the record. This ensures complete uniqueness control and acts better than relying only on optimistic concurrency (due to a flush). 
But the callbacks are now using transactions, so no side effects should be introduced, even relying solely on a unique index check.

### QA specification

_No response_

vita-kotacka:
## Summary

Resolves #2004

Optimizes activation ID and code generation in `initActivation` by eliminating application-level SELECT COUNT uniqueness checks in favor of DB-level constraints.

### Bug fix
- Added missing null check for `MasterKeyPairEntity` in `initActivation` — all other callers already checked this and threw `NO_MASTER_SERVER_KEYPAIR`

### Performance optimization
- **Activation ID**: Removed SELECT COUNT loop entirely; UUID collision probability (~10⁻¹⁸) is negligible and the primary key constraint handles the rare case
- **Activation code**: Replaced SELECT COUNT loop with a DB unique constraint `pa_activation_code_application_uk` on `(application_id, activation_code)`; saves one DB roundtrip per `initActivation` call
- `DataIntegrityViolationException` on insert is caught and mapped to `UNABLE_TO_GENERATE_ACTIVATION_CODE`

### Query fix
- Rewrote `getActivationCountByActivationCode` from a LIKE prefix query (`LIKE :activationIdShort%`) to an exact match (`= :activationCode`), fixing a correctness bug where codes sharing a prefix were incorrectly counted as collisions

### Liquibase
- Drop non-unique index `pa_activation_code`
- Add unique constraint `pa_activation_code_application_uk` on `(application_id, activation_code)`

### Tests
- `ActivationInitServiceBehaviorTest` — 15 unit tests for `initActivation`
- `ActivationRepositoryTest` — 4 repository tests for `getActivationCountByActivationCode`, including a regression test verifying the LIKE→= fix


docs/db/changelog/changesets/powerauth-java-server/2.0.x/20260316-activation-code-unique.xml
```diff
@@ -0,0 +1,44 @@
...
+  ~
+  ~ You should have received a copy of the GNU Affero General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
+
+    <changeSet id="1" logicalFilePath="powerauth-java-server/2.0.x/20260316-activation-code-unique.xml" author="Vit Kotacka">
```
banterCZ: Aren't we targeting 2.1, or is it supposed to be a backport?

docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml
```diff
@@ -0,0 +1,44 @@
...
+    </changeSet>
+
+    <changeSet id="2" logicalFilePath="powerauth-java-server/2.0.x/20260316-activation-code-unique.xml" author="Vit Kotacka">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <uniqueConstraintExists tableName="pa_activation" constraintName="pa_activation_code_application_uk"/>
+            </not>
+        </preConditions>
+        <comment>Add unique constraint on (application_id, activation_code) to enforce activation code uniqueness at DB level, replacing the application-level SELECT COUNT uniqueness check</comment>
+        <addUniqueConstraint tableName="pa_activation"
```
banterCZ, line 39: @romanstrobl @zcgandcomp If we want to deploy without downtime, ideally, how would this affect the existing customers? Shouldn't we prefer some optimization, such as _parallel index build_?

docs/db/changelog/changesets/powerauth-java-server/2.0.x/db.changelog-version.xml
```diff
@@ -11,9 +11,10 @@
     <include file="20250925-crypto4-upgrade.xml" relativeToChangelogFile="true" />
     <include file="20251005-activation-parent-id.xml" relativeToChangelogFile="true" />
     <include file="20251106-allow-null-legacy-keypairs.xml" relativeToChangelogFile="true" />
-    <include file="20251231-allow-null-legacy-server-keypair.xml" relativeToChangelogFile="true" />
```
banterCZ, line 14: 🤔 Why do we reorder changesets?

banterCZ (commented):
Since Roman asked me also for a review, I added a few comments.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20260316-activation-code-unique.xml
romanstrobl: I agree, we should taget `2.1.x`.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/db.changelog-version.xml
```diff
@@ -11,9 +11,10 @@
...
+    <include file="20251215-add-tag-2.0.0.xml" relativeToChangelogFile="true" />
     <include file="20251222-crypto4-keys-tables.xml" relativeToChangelogFile="true" />
+    <include file="20251231-allow-null-legacy-server-keypair.xml" relativeToChangelogFile="true" />
     <include file="20260107-crypto4-store-fingerprint.xml" relativeToChangelogFile="true" />
-    <include file="20251215-add-tag-2.0.0.xml" relativeToChangelogFile="true" />
+    <include file="20260316-activation-code-unique.xml" relativeToChangelogFile="true" />
```
romanstrobl: The changeset does not belong to the `2.0.x` folder, it should be in the `2.1.x` folder.

docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml
romanstrobl: For 1 million rows this index should take ~5 seconds on modern hardware. I am not sure we need to optimize this unless the client has tens of millions of activations?

romanstrobl (changes requested):
Good job. I added several comments.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/db.changelog-version.xml
banterCZ: Do we want to mention that in the migration guide? And shouldn't it be reflected in `Database-Structure.md`?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
```diff
@@ -136,40 +137,20 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
             // Generate hash from activation OTP
             final String activationOtpHash = StringUtils.hasText(activationOtp) ? PasswordHash.hash(activationOtp.getBytes(StandardCharsets.UTF_8)) : null;
 
-            // Generate new activation data, generate a unique activation ID
-            String activationId = null;
-            for (int i = 0; i < powerAuthServiceConfiguration.getActivationGenerateActivationIdIterations(); i++) {
```
jnpsk, line 141: It seems the configuration property `activationGenerateActivationIdIterations` has not any usage now. Same applies for `activationGenerateActivationCodeIterations`.

jnpsk (commented):
I agree, the proactive database lookup is unnecessary.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/20260316-activation-code-unique.xml
vita-kotacka: Moved to `2.1.x`.

docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml
vita-kotacka: @romanstrobl The optimization is not on level of DB, but Java - the unnecessary `SELECT COUNT` was removed and very unlikely collision is handled by the unique constraint.

docs/db/changelog/changesets/powerauth-java-server/2.0.x/db.changelog-version.xml
vita-kotacka: They weren't ordered by date. Obsolete now after the changeset was moved to `2.1.x` directory/version.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
vita-kotacka: The properties have been removed. ✅

docs/db/changelog/changesets/powerauth-java-server/2.0.x/db.changelog-version.xml
vita-kotacka: Moved. ✅

vita-kotacka: @banterCZ I don't see any specific index description in the file. `activation_code` is already marked as an `index`.

vita-kotacka: Migration guide `2.1.0` has been updated with DB changes.

docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml
```diff
@@ -0,0 +1,44 @@
...
+    </changeSet>
+
+    <changeSet id="2" logicalFilePath="powerauth-java-server/2.1.x/20260316-activation-code-unique.xml" author="Vit Kotacka">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <uniqueConstraintExists tableName="pa_activation" constraintName="pa_activation_code_application_uk"/>
+            </not>
+        </preConditions>
+        <comment>Add unique constraint on (application_id, activation_code) to enforce activation code uniqueness at DB level, replacing the application-level SELECT COUNT uniqueness check</comment>
+        <addUniqueConstraint tableName="pa_activation"
```
banterCZ, line 39: Sorry, not mentioned earlier. Unfortunatelly we manage `/docs/sql/*` manually. We run liquibase scripts and edit the files. In the future, it should be done by some pipelines.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
jnpsk: I think that not only getters may be removed, but also the fields together with the properties file records.

vita-kotacka: Fixed. ✅

docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml
vita-kotacka: 😁

vita-kotacka: Fixed. ✅

docs/sql/postgresql/migration_2.0.0_2.1.0.sql
```diff
@@ -0,0 +1,17 @@
...
+-- Liquibase version: 4.33.0
+-- *********************************************************************
+
+-- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::1::Vit Kotacka
+-- Drop non-unique index on pa_activation(activation_code) before replacing with unique constraint
+DROP INDEX pa_activation_code;
+
+-- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::2::Vit Kotacka
+-- Add unique constraint on (application_id, activation_code) to enforce activation code uniqueness at DB level, replacing the application-level SELECT COUNT uniqueness check
+ALTER TABLE pa_activation ADD CONSTRAINT pa_activation_code_application_uk UNIQUE (application_id, activation_code);
```
banterCZ, line 16: I know, it is too much manual work, but I believe that it should also be included in `create_schema.sql`.

banterCZ (dismissed):

docs/sql/postgresql/migration_2.0.0_2.1.0.sql
vita-kotacka: Fixed. ✅

docs/PowerAuth-Server-2.1.0.md
```diff
@@ -9,3 +9,12 @@ The `$.requestObject.applicationId` attribute used to verify the challenge can b
 ### POST /fido2/assertions/challenge
 
 The response contains a new `$.responseObject.operationId` attribute representing the operation ID the assertion is associated with.
+
+## Database Changes
```
romanstrobl, line 13: I am wondering, do we want to emphasize that building the index can take some times on large tables? It may be useful to run the liquibase manually to avoid an outage in this case?

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.jdbc.Sql;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+/**
+ * Test for {@link ActivationRepository}.
```
romanstrobl, line 32: Missing author.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehaviorTest.java
```diff
@@ -0,0 +1,338 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
```
romanstrobl, line 63: Missing JavaDoc and author.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
```diff
@@ -237,6 +217,13 @@ public InitActivationResponse initActivation(InitActivationRequest request) thro
         } catch (GenericServiceException ex) {
             // already logged
             throw ex;
+        } catch (DataIntegrityViolationException ex) {
+            // The DB enforces uniqueness of (application_id, activation_code) via pa_activation_code_application_uk
+            // and uniqueness of activation_id via the primary key. UUID activation ID collision probability is
+            // negligible (~10^-18), so this is almost certainly an activation code collision.
+            logger.error("Unable to generate unique activation code, activation code collision occurred", ex);
```
romanstrobl: Shouldn't we log the UUID?

romanstrobl (commented):
I added several additional comments.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehaviorTest.java
vita-kotacka: ✅

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
vita-kotacka: ✅

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
vita-kotacka: @romanstrobl No strong opinion on this. What is the common practice in Wultra?

vita-kotacka: Logging the raw UUID on collision leaks confirmation that a record with that ID exists — a minor but unnecessary information disclosure. Better to log a sanitized indicator or keep it at `DEBUG` level rather than exposing the value in higher-severity logs.

romanstrobl: Yes, it can be sanitized, but I think we should log these events on `WARN` level, because they indicate that there is a problem with random generator.

docs/PowerAuth-Server-2.1.0.md
vita-kotacka: Added to the guide. ✅

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationInitServiceBehavior.java
vita-kotacka: Now it's logging the UUID in the form `12345678-abcd`. The log level is `WARN`.

romanstrobl (approved):
Looks good now.

PR #2288, opened 2026-03-18
Resolves #2173 Created application response should return mobile SDK configuration

### LINKED ISSUE #2173: Created application response should return mobile SDK configuration

vita-kotacka:
## Summary

Resolves #2173

Added missing JavaDoc and `@author` tags to new Java files.


vita-kotacka:
Created by mistake.

PR #2289, opened 2026-03-19
Resolves #2173 Add missing JavaDoc and @author to new files

### LINKED ISSUE #2173: Created application response should return mobile SDK configuration

vita-kotacka:
## Summary

Resolves #2173

Added missing JavaDoc and `@author` tags to new Java files introduced in this issue.


romanstrobl (changes requested):
We include our e-mail address with author, e.g.:

`@author Petr Dvorak, petr@wultra.com`

romanstrobl (approved):
Looks OK.

PR #2297, opened 2026-03-25
Fix #2296: Convert credential ID to Base64 standard if received in Base64 URL by API

### LINKED ISSUE #2296: Count with Base64 URL encoded credential ID in FIDO2 APIs
### Task specification

As of now, we are relying on the external party's normalization of the registration and assertion flows.
If the normalization is not performed, the system fails to create an assertion challenge. 
Internal format is Base64 and this is risky to change due to compatibility constraints.
To avoid issues in the integration of the product we will compensate on API level.
Both registration and assertion flows has to be addressed.

pavsindelar:
The ultimate solution is to normalize the Base64 encoding on input (discussed with @zcgandcomp).

### DIFF
--- a/powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/converter/serialization/Base64UrlToBase64Deserializer.java
+++ b/powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/converter/serialization/Base64UrlToBase64Deserializer.java
@@ -1,0 +1,81 @@
+/*
+ * PowerAuth Server and related software components
+ * Copyright (C) 2026 Wultra s.r.o.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.wultra.security.powerauth.fido2.model.converter.serialization;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+
+import java.io.IOException;
+import java.io.Serial;
+
+/**
+ * Jackson deserializer that transparently normalizes Base64URL-encoded strings
+ * to standard Base64 encoding (RFC 4648 §4). The deserializer accepts both encodings
+ * and always produces a standard Base64 string.
+ *
+ * @author Pavel Sindelar, pavel.sindelar@wultra.com
+ */
+public class Base64UrlToBase64Deserializer extends StdDeserializer<String> {
+
+    @Serial
+    private static final long serialVersionUID = 4871203406847302591L;
+
+    public Base64UrlToBase64Deserializer() {
+        this(null);
+    }
+
+    public Base64UrlToBase64Deserializer(final Class<?> valueClass) {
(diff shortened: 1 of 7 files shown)


powerauth-fido2-model/src/main/java/com/wultra/security/powerauth/fido2/model/converter/serialization/Base64UrlToBase64Deserializer.java
```diff
@@ -0,0 +1,71 @@
...
+
+        final int remainder = normalized.length() % 4;
+        if (remainder == 0) {
+            // no padding needed
+            return normalized;
+        }
+
+        // add padding
+        return normalized + "=".repeat(4 - remainder);
+    }
```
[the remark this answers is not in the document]
pavsindelar: Fixed.

zcgandcomp (dismissed):
The logic of the mechanism looks good to to.

romanstrobl (changes requested):
The deserializer will likely accept malformed base64 strings, such as `abcd$%` or with malformed padding. Perhaps we should validate the Base64 by trying to decode it?

We may also want to trim the whitespaces?

pavsindelar:
> The deserializer will likely accept malformed base64 strings, such as `abcd$%` or with malformed padding. Perhaps we should validate the Base64 by trying to decode it?
> 
> We may also want to trim the whitespaces?

@romanstrobl, good point, we shall validate that 👍

powerauth-fido2/src/test/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/Base64UrlToBase64DeserializerTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    void testDeserialize_dashOnly_convertedToPlus() throws Exception {
+        final String result = objectMapper.readValue("\"AA-A\"", String.class);
+        assertEquals("AA+A", result);
+    }
+
+    @Test
+    void testDeserialize_underscoreOnly_convertedToSlash() throws Exception {
```
banterCZ: Nitpick, you may use `@ParameterizedTest`

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/controller/RESTControllerAdvice.java
```diff
@@ -140,4 +140,14 @@ public class RESTControllerAdvice {
         return new ObjectResponse<>("ERROR", error);
     }
 
+    @ExceptionHandler(HttpMessageNotReadableException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    public @ResponseBody ObjectResponse<PowerAuthError> handleHttpMessageNotReadableException(final HttpMessageNotReadableException e) {
+        logger.error("Error occurred while processing the request: {}", e.getMessage());
+        logger.debug("Exception details:", e);
```
banterCZ, line 147: 🤔 I am thinking whether we shouldn't always log the stacktrace.

powerauth-fido2/src/test/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/Base64UrlToBase64DeserializerTest.java
pavsindelar: Good point, changed to parameterized tests.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/controller/RESTControllerAdvice.java
pavsindelar: The pattern in this exception handler is consistently to log the stack trace on a debug level. Also, in case of this exception, the message is very descriptive so I don't think it is necessary really.

pavsindelar:
> > The deserializer will likely accept malformed base64 strings, such as `abcd$%` or with malformed padding. Perhaps we should validate the Base64 by trying to decode it?
> > We may also want to trim the whitespaces?
> 
> @romanstrobl, good point, we shall validate that 👍

@romanstrobl, I have added a validation by decoding before returning the normalized string. As for stripping the whitespaces, I wouldn't do that for a JSON field. If the client calls the endpoint with extra whitespaces, he gets 400 with descriptive error message, allowing him to fix the issue. We validate just for a valid Base64 string.

romanstrobl (approved):
Looks good now.

PR #2298, opened 2026-03-27
Fix #2177: Update offline authentication documentation

### LINKED ISSUE #2177: Update offline authentication documentation
### Task specification

The documentation refers to WebFlow for QR code structure. We should move all relevant documentation for offline authentication to PowerAuth server to avoid such references.

jnpsk:
- Updating the Offline Authentication doc to match v4
- Moving some information from documentation in Enrollment Server


romanstrobl (approved):
Thanks!

PR #2305, opened 2026-04-01
Fix #2304: Improve log of operation list controller

### LINKED ISSUE #2304: Improve log of operation list controller
### Task specification

Right now, we are not able to distinguish whether the response is empty.

banterCZ:


romanstrobl (approved):
Nice improvement!

PR #2313, opened 2026-04-07
Fix #2312: Bad performance of create operation

### LINKED ISSUE #2312: Bad performance of create operation

develop

### Environment

local

Due to backtracking in the Java regex engine, the complexity of this regex pattern is O(n²), which leads to poor performance on large inputs:

```
private static final Pattern PATTERN_FORBIDDEN_ASCII = Pattern.compile(".*[\\x00-\\x09\\x0B-\\x1F].*");
```

The leading and trailing `.*` are greedy and cause the regex engine to backtrack when no match is found. In such cases, the engine tries matching at all positions in the input, rescanning the string many times.

For large inputs without forbidden characters it leads to O(n²) behavior. Removing the leading and trailing `.*` makes the complexity O(n).

### Log output

_No response_

romanstrobl:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
```diff
@@ -89,7 +89,7 @@ public class OperationServiceBehavior {
     /**
      * All characters with ASCII code < 32 (except 10 line feed) are forbidden (e.g. \t should not be in the string).
      */
-    private static final Pattern PATTERN_FORBIDDEN_ASCII = Pattern.compile(".*[\\x00-\\x09\\x0B-\\x1F].*");
+    private static final Pattern PATTERN_FORBIDDEN_ASCII = Pattern.compile("[\\x00-\\x09\\x0B-\\x1F]");
```
banterCZ, line 95: Couldn't we comment on the reasoning? This works, but somebody may break it by accident in the future. Or even better, use your benchmark as a unit test and fail it when the threshold is reached.
May we also try `matches()` instead of `find()`?

banterCZ (dismissed):
Thanks for the fix. I am thinking about how to make it robust to avoid regression in the future.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
romanstrobl: I will add the comment.

romanstrobl: As for `matches()` vs `find()`, I suggest with stick with `find()`.

For me the intention is clear with `find()`, we search for any forbidden character. Matches requires the `.*` addition and makes it more complicated in my opinion.

The .find() version does:
- for each char:
  - check char range
    
The .matches() version does:
- consume entire string (.*)
- backtrack
- try match
- backtrack more

romanstrobl: If you want the test, feel free to add it as the author of this code.

banterCZ (commented):
I have added a benchmark test, we may drop it if not agreed.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OperationServiceBehavior.java
```diff
@@ -301,13 +304,17 @@ private void validate(final OperationCreateRequest request) throws GenericServic
...
             }
-            if (PATTERN_FORBIDDEN_ASCII.matcher(parameter.getValue()).find()) {
+            if (isForbiddenAscii(parameter.getValue())) {
                 logger.warn("TEXT parameter value: '{}' contains invalid characters.", parameter.getValue());
                 throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
             }
         }
     }
 
+    protected static boolean isForbiddenAscii(final String value) {
```
banterCZ: We may extract it to a separate class, now or later.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ForbiddenAsciiRegexPerformanceTest.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Test class for {@link OperationServiceBehavior#isForbiddenAscii(String)}.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ */
+class ForbiddenAsciiRegexPerformanceTest {
+
+    private static final double THRESHOLD_NS = 1000.0; // 1 microsecond
+
+    @Test
```
banterCZ, line 39: It lasts 23 seconds, which makes the build longer. We may introduce a different test suite or move it to the integration test phase.

romanstrobl: Can we speed it up? We can do a quick test within 1 second to confirm that the performance problem is not present, we don't need to run benchmarks during build.

romanstrobl: The benchmark seems incorrect. The problem only appears on very large strings (e.g. 24 kB), not on strings which have < 100 chars.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
jnpsk: This solution looks good for me.
However, as an alternative, we may avoid the regex completely as it just checks if there are printable chars or new line only, which should not be a complicated condition for a loop over a string.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ForbiddenAsciiRegexPerformanceTest.java
banterCZ: A good point. Let's try a different approach, single run with a longer String.

romanstrobl:
I think the simple test is better, even though it's not a benchmark, it does not slow down the build and will catch early performance problems with the regexp on large strings. Thanks.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
romanstrobl: Yes, I was also thinking about this solution, however it seems quite low level and the performance without `.*` is fine.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/OperationServiceBehavior.java
romanstrobl: Let's raise an issue.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ForbiddenAsciiRegexPerformanceTest.java
romanstrobl: Perfect, now we don't slow down the build and test the performance problem as it was analyzed.

banterCZ (dismissed):

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/ForbiddenAsciiRegexPerformanceTest.java
```diff
@@ -0,0 +1,55 @@
...
+        final String input = generateString();
+
+        // Warmup to avoid measuring initialization cost
+        assertFalse(OperationServiceBehavior.isForbiddenAscii(input));
+
+        // Assert for regexp performance
+        assertTimeout(Duration.ofMillis(10), () -> {
+            OperationServiceBehavior.isForbiddenAscii(input);
+        }, "Regex performance is too slow for input");
+    }
```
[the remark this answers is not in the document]
romanstrobl: This is fine, we use a warmup to avoid exactly this issue.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/ForbiddenAsciiRegexPerformanceTest.java
```diff
@@ -0,0 +1,55 @@
...
+        final String input = generateString();
+
+        // Warmup to avoid measuring initialization cost
+        assertFalse(OperationServiceBehavior.isForbiddenAscii(input));
+
+        // Assert for regexp performance
+        assertTimeout(Duration.ofMillis(10), () -> {
+            OperationServiceBehavior.isForbiddenAscii(input);
+        }, "Regex performance is too slow for input");
+    }
```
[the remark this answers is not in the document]
romanstrobl: This is fine, we use a warmup to avoid exactly this issue.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/OperationServiceBehavior.java
```diff
@@ -300,13 +303,17 @@ private void validate(final OperationCreateRequest request) throws GenericServic
...
             }
-            if (PATTERN_FORBIDDEN_ASCII.matcher(parameter.getValue()).find()) {
+            if (isForbiddenAscii(parameter.getValue())) {
                 logger.warn("TEXT parameter value: '{}' contains invalid characters.", parameter.getValue());
                 throw localizationProvider.buildExceptionForCode(ServiceError.INVALID_REQUEST);
             }
         }
     }
 
+    protected static boolean isForbiddenAscii(final String value) {
```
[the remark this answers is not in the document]
banterCZ: Ok, let's make it _package private_.

banterCZ (approved):

PR #2334, opened 2026-04-13
Fix #2306: In DSA verification endpoint for v4 allow P-256 curve

### LINKED ISSUE #2306: In DSA verification endpoint for v4 allow P-256 curve
### Task specification

Currently P-256 curve is not supported in `/rest/v4/dsa/verify`.

romanstrobl:


jnpsk (approved):
looks good

PR #2338, opened 2026-04-15
Fix #2307: Remove unused authentication code types

### LINKED ISSUE #2307: Remove unused authentication code types
### Task specification

Some of the authentication code types are not used in practice:
- `BIOMETRY`
- `KNOWLEDGE`
- `POSSESSION_KNOWLEDGE_BIOMETRY`

We should remove them.

jnpsk:


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/converter/AuthenticationCodeTypeConverter.java
```diff
@@ -19,37 +19,34 @@
...
-            case KNOWLEDGE -> PowerAuthCodeType.KNOWLEDGE;
-            case BIOMETRY -> PowerAuthCodeType.BIOMETRY;
             case POSSESSION_KNOWLEDGE -> PowerAuthCodeType.POSSESSION_KNOWLEDGE;
             case POSSESSION_BIOMETRY -> PowerAuthCodeType.POSSESSION_BIOMETRY;
-            default -> PowerAuthCodeType.POSSESSION_KNOWLEDGE_BIOMETRY;
         };
     }
 
-    public AuthenticationCodeType convertFrom(String authCodeType) {
+    public static AuthenticationCodeType convert(final String authCodeType) {
```
romanstrobl, line 41: What happens in case this method is called with one of the removed signature types? E.g. for tokens it could theoretically be taken from DB for a non-existent signature type (`TokenEntity.signature_type`).

romanstrobl (approved):
Thanks for the cleanup!

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/converter/AuthenticationCodeTypeConverter.java
jnpsk: This method calls `AuthenticationCodeType#enumFromString`, which returns null for unknown enum value. The only usage is in the token validation procedure to build a response (i.e. with which factor the token was issued).
In your scenario, the token should still be validated, just its response would contain `null` in `ValidateTokenResponse.authenticationCodeType`. Although this sounds inconvenient, I assume the token validation endpoint would not be called on token created with `3FA` or `KNOWLEDGE` or `BIOMETRY` as it is not supported by SDK for some time.

PR #2311, opened 2026-04-07
Resolves #2004: Optimize activation code generation — DB changeset rework

vita-kotacka:
## Summary

Rework of the DB migration for issue #2004 (optimize activation code generation) based on offline review feedback.

### Liquibase Changesets
- **Added `20260316-activation-code-fix-nulls.xml`** — NULL retrofit changeset for legacy databases
  - Uses `preConditions onFail="MARK_RAN"` — completely skipped (zero cost) on clean databases
  - Sets `activation_code = 'LEGACY-<uuid>'` for any NULL rows using DB-native UUID functions
- **Reordered migration steps** for production safety: NULL retrofit → ADD NOT NULL → ADD UNIQUE → DROP old index
  - NOT NULL fails fast (cheap full scan) before the expensive index build
- **Column order in unique constraint**: `(activation_code, application_id)` — `activation_code` is more selective (random 23-char string) and appears as the leading column

### Migration SQL
Regenerated all three `migration_2.0.0_2.1.0.sql` files (PostgreSQL, Oracle, MSSQL) to reflect the updated changesets.

### Performance Test Scripts (`docs-private/db/`)
Scripts to benchmark the migration on 10M rows across all supported databases.

#### Results summary (ADD UNIQUE on 10M rows)
| DB | Duration | Notes |
|----|----------|-------|
| Oracle 23ai Free | 6.9 s | Native ARM64 |
| PostgreSQL 18.2 | 18.9 s | Native ARM64 |
| Azure SQL Edge 2.0 | 22.0 s | Native ARM64 |
| MSSQL 2022 | 88.8 s | QEMU/amd64 |
| Oracle 19c Enterprise | 251.9 s | QEMU/amd64 |

See `docs-private/db/README.md` for full results and analysis.

### Oracle Script Fixes
- Replaced `INSERT /*+ APPEND */` (caused silent data generation failure in PL/SQL) with batched `FORALL` loop
- Added `ALTER SESSION SET CONTAINER = FREEPDB1` for correct PDB context
- Removed invalid `GRANT CREATE INDEX` privilege
- Added exception handler for visibility on failure

## Testing
- ✅ PostgreSQL 18.2 — perf test run successfully
- ✅ Oracle 19c Enterprise — perf test run successfully
- ✅ Oracle 23ai Free — perf test run successfully
- ✅ MSSQL / Azure SQL Edge — perf test run successfully

PR #2358, opened 2026-04-28
Fix #2004: Drop pa_activation_code index before ALTER COLUMN on MSSQL

### LINKED ISSUE #2004: Optimize Activation Code generation

vita-kotacka:
## Summary

Fix for MSSQL migration failure introduced by PR #2311.

## Problem

MSSQL error 5074: *"The index 'pa_activation_code' is dependent on column 'activation_code'."*

MSSQL blocks `ALTER COLUMN` on any column that has a dependent index — even when only adding `NOT NULL`. PostgreSQL and Oracle do not have this restriction.

## Fix

Added changeset `id=0` to drop the old non-unique index `pa_activation_code` **before** the `ALTER COLUMN` (changeset `id=1`).

Existing changesets `id=1..3` are **unchanged** to avoid re-running on environments where the migration already succeeded on PostgreSQL or Oracle (Liquibase tracks by `id + logicalFilePath + author`).

### New execution order
| Changeset | Operation |
|-----------|-----------|
| fix-nulls::1 | Backfill NULL activation_code (MARK_RAN if none) |
| unique::0 | **DROP INDEX** pa_activation_code (new — fixes MSSQL) |
| unique::1 | ADD NOT NULL on activation_code |
| unique::2 | ADD UNIQUE constraint (activation_code, application_id) |
| unique::3 | DROP INDEX pa_activation_code (safety net — MARK_RAN on fresh installs) |

### Behaviour on all scenarios
| Scenario | id=0 | id=1..3 |
|----------|------|---------|
| Fresh MSSQL | drops index ✅ | NOT NULL succeeds, UNIQUE added, id=3 MARK_RAN |
| Fresh PostgreSQL/Oracle | drops index ✅ | same as above |
| Already ran id=1..3 (PostgreSQL/Oracle) | MARK_RAN ✅ | skipped (already in DATABASECHANGELOG) |

## Files changed
- `docs/db/changelog/changesets/powerauth-java-server/2.1.x/20260316-activation-code-unique.xml`
- `docs/sql/postgresql/migration_2.0.0_2.1.0.sql`
- `docs/sql/oracle/migration_2.0.0_2.1.0.sql`
- `docs/sql/mssql/migration_2.0.0_2.1.0.sql`

### DIFF
--- a/docs/PowerAuth-Server-2.1.0.md
+++ b/docs/PowerAuth-Server-2.1.0.md
@@ -34,15 +34,23 @@ The migration applies the following changes in order:
 
 1. **Backfill NULL `activation_code` values** _(legacy databases only)_ — if any rows have a `NULL` activation code, they are updated with a unique `LEGACY-<uuid>` placeholder value. This step is automatically **skipped** (marked as ran) if no NULL values exist, which is the expected case for all databases managed by Liquibase since 1.4.x.
 
-2. **Add `NOT NULL` constraint** on `activation_code` — aligns the DB schema with the JPA entity definition. This is a fast operation (metadata-only on MSSQL when no NULLs exist; full table scan on PostgreSQL and Oracle).
+2. _(MSSQL only)_ **Drop the old non-unique index** `pa_activation_code` — MSSQL blocks `ALTER COLUMN` when a dependent index exists. This step is automatically **skipped** on PostgreSQL and Oracle.
 
-3. **Add unique constraint** `pa_activation_code_application_uk` on `(activation_code, application_id)` — enforces activation code uniqueness at the database level per application, replacing the previous application-level check. **This is an index build and can take significant time on large tables** (see performance data below).
+3. **Add `NOT NULL` constraint** on `activation_code` — aligns the DB schema with the JPA entity definition. This is a fast operation (metadata-only on MSSQL when no NULLs exist; full table scan on PostgreSQL and Oracle).
 
-4. **Drop the old non-unique index** `pa_activation_code` on `activation_code` — replaced by the unique constraint above (near-instant operation).
+4. **Add unique constraint** `pa_activation_code_application_uk` on `(activation_code, application_id)` — enforces activation code uniqueness at the database level per application, replacing the previous application-level check. **This is an index build and can take significant time on large tables** (see performance data below).
+
+5. _(PostgreSQL, Oracle)_ **Drop the old non-unique index** `pa_activation_code` — replaced by the unique constraint above (near-instant operation). Automatically skipped on MSSQL (already dropped in step 2).
 
 > ⚠️ **Maintenance window recommended.**
-> Step 3 (index build) locks the table and can take several minutes on large deployments.
+> Step 4 (index build) locks the table and can take several minutes on large deployments.
 > Consider running the Liquibase migration manually during a maintenance window before upgrading the application.
+
+> ⚠️ **MSSQL: brief index gap during migration.**
+> On MSSQL, step 2 drops the old `pa_activation_code` index before the new unique constraint is created in step 4.
+> During this window, queries filtering on `activation_code` alone will perform a full table scan.
+> The gap is bounded by steps 3 and 4 only, which are fast operations on MSSQL (metadata-only NOT NULL + index build).
+> Plan the migration during a maintenance window to avoid query performance impact.
 
 > ⚠️ **Legacy databases with NULL `activation_code` values.**
 > Step 1 (backfill) will UPDATE any rows with NULL `activation_code`.
--- a/docs/sql/oracle/migration_2.0.0_2.1.0.sql
+++ b/docs/sql/oracle/migration_2.0.0_2.1.0.sql
@@ -16,10 +16,6 @@ UPDATE pa_activation
             SET activation_code = 'LEGACY-' || LOWER(RAWTOHEX(SYS_GUID()))
             WHERE activation_code IS NULL;
 
--- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::0::Vit Kotacka
--- MSSQL only: drop non-unique index pa_activation_code before ALTER COLUMN. MSSQL blocks ALTER COLUMN when a dependent index exists (error 5074); PostgreSQL and Oracle do not have this restriction.
-DROP INDEX pa_activation_code;
-
 -- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::1::Vit Kotacka
 -- Add NOT NULL constraint on pa_activation(activation_code) to align the DB schema with the JPA entity definition. Applied before the unique constraint to fail fast if NULL values exist, avoiding an expensive index build on dirty data.
 ALTER TABLE pa_activation MODIFY activation_code NOT NULL;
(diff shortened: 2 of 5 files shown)


romanstrobl (commented):
Shouldn't we do this only on MSSQL? On other databases it will just slow down the processing?

vita-kotacka:
Fixed in d7694134 — changeset id=0 now uses `<dbms type="mssql"/>` precondition, so it is skipped (MARK_RAN) on PostgreSQL and Oracle.

zcgandcomp (changes requested):
Like this is fine, dropping index before introducing the not null.

The not null is in my opinion nice to have - so for future references we should avoid nice to have changes.

But this will lead to a situation where the update cannot be done on the MSSQL without an outage. AS you drop the index, and the new one is not yet created.

If this cannot be prevented (like it can by not introducing the not null), we need to state this in the migration guidel

vita-kotacka:
Addressed in ba6774ff — added a note about the MSSQL index gap (step 2 drops the old index, gap exists until step 4 creates the unique constraint). Also clarified that step 2 is MSSQL-only and step 5 (PostgreSQL/Oracle drop) is automatically skipped on MSSQL. Maintenance window warning updated to reference the correct step number.

Agree on the general principle of avoiding nice-to-have changes in fixes — noted for future reference.

docs/sql/oracle/migration_2.0.0_2.1.0.sql
```diff
@@ -16,6 +16,10 @@ UPDATE pa_activation
             SET activation_code = 'LEGACY-' || LOWER(RAWTOHEX(SYS_GUID()))
             WHERE activation_code IS NULL;
 
+-- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::0::Vit Kotacka
+-- MSSQL only: drop non-unique index pa_activation_code before ALTER COLUMN. MSSQL blocks ALTER COLUMN when a dependent index exists (error 5074); PostgreSQL and Oracle do not have this restriction.
+DROP INDEX pa_activation_code;
```
romanstrobl: The comment says MSSQL only but the script is for Oracle.

docs/sql/postgresql/migration_2.0.0_2.1.0.sql
```diff
@@ -16,6 +16,10 @@ UPDATE pa_activation
             SET activation_code = 'LEGACY-' || gen_random_uuid()::text
             WHERE activation_code IS NULL;
 
+-- Changeset powerauth-java-server/2.1.x/20260316-activation-code-unique.xml::0::Vit Kotacka
+-- MSSQL only: drop non-unique index pa_activation_code before ALTER COLUMN. MSSQL blocks ALTER COLUMN when a dependent index exists (error 5074); PostgreSQL and Oracle do not have this restriction.
+DROP INDEX pa_activation_code;
```
romanstrobl: The comment says MSSQL only but the script is for PostgreSQL.

romanstrobl (changes requested):
We need to update the SQL scripts.

docs/sql/oracle/migration_2.0.0_2.1.0.sql
zcgandcomp: Well, this is problem with the xml, as in the xml only condition is marked for MSSQL, but the drop is for all DB

                <dbms type="mssql"/>

docs/PowerAuth-Server-2.1.0.md
```diff
@@ -34,16 +34,24 @@ The migration applies the following changes in order:
...
 
-4. **Drop the old non-unique index** `pa_activation_code` on `activation_code` — replaced by the unique constraint above (near-instant operation).
+4. **Add unique constraint** `pa_activation_code_application_uk` on `(activation_code, application_id)` — enforces activation code uniqueness at the database level per application, replacing the previous application-level check. **This is an index build and can take significant time on large tables** (see performance data below).
+
+5. _(PostgreSQL, Oracle)_ **Drop the old non-unique index** `pa_activation_code` — replaced by the unique constraint above (near-instant operation). Automatically skipped on MSSQL (already dropped in step 2).
 
 > ⚠️ **Maintenance window recommended.**
-> Step 3 (index build) locks the table and can take several minutes on large deployments.
+> Step 4 (index build) locks the table and can take several minutes on large deployments.
 > Consider running the Liquibase migration manually during a maintenance window before upgrading the application.
```
zcgandcomp, line 47: Liquibase should not be run manually, and be mindful, that when PowerAuth server is used as a component of PowerAuth Cloud the manual liquibase is not even possible,

vita-kotacka:
Fixed in fb795e80 — removed changeset id=0 from the PostgreSQL and Oracle SQL scripts. Liquibase offline mode doesn't evaluate `<dbms>` preconditions, so it appeared in all 3 files even though at runtime it is MSSQL-only. The MSSQL script correctly retains it.

romanstrobl (approved):
Looks OK.

docs/sql/postgresql/migration_2.0.0_2.1.0.sql
vita-kotacka: See https://github.com/wultra/powerauth-server/pull/2358#issuecomment-4342581984

docs/sql/oracle/migration_2.0.0_2.1.0.sql
vita-kotacka: See https://github.com/wultra/powerauth-server/pull/2358#issuecomment-4342581984

docs/PowerAuth-Server-2.1.0.md
```diff
@@ -34,16 +34,24 @@ The migration applies the following changes in order:
...
+3. **Add `NOT NULL` constraint** on `activation_code` — aligns the DB schema with the JPA entity definition. This is a fast operation (metadata-only on MSSQL when no NULLs exist; full table scan on PostgreSQL and Oracle).
 
-4. **Drop the old non-unique index** `pa_activation_code` on `activation_code` — replaced by the unique constraint above (near-instant operation).
+4. **Add unique constraint** `pa_activation_code_application_uk` on `(activation_code, application_id)` — enforces activation code uniqueness at the database level per application, replacing the previous application-level check. **This is an index build and can take significant time on large tables** (see performance data below).
+
+5. _(PostgreSQL, Oracle)_ **Drop the old non-unique index** `pa_activation_code` — replaced by the unique constraint above (near-instant operation). Automatically skipped on MSSQL (already dropped in step 2).
 
 > ⚠️ **Maintenance window recommended.**
```
zcgandcomp, line 45: Just a comment: The DB changes should be what was changed, not the full description of the steps. It could be, but this is not intented.

If a special procedure is required, it has to be specified.

zcgandcomp (changes requested):
The drop index is for all DB.
And it was changed in the mean time.

zcgandcomp (approved):
Ok,
the script is ok.
I would rephrase the migration guide, but it's not crtiical

PR #2375, opened 2026-05-07
Fix #2374: Operation claim should fail when initiated with an app that is unable to confirm

### LINKED ISSUE #2374: Operation claim should fail when initiated with an app that is unable to confirm

2.1.x

### Environment

_No response_

1. Create a new anonymous operation with a registration flag X
2. Try to claim the operation with a registered device that does not have flag X

=> The claim operation succeeds

Expected behavior: The claim fails because the device cannot approve the operation anyway.

### Log output

_No response_

romanstrobl:


zcgandcomp (changes requested):
We need to introduce a new claim API with a claim for activationId (and transitively for userId)

Reason: The current fix will prevent claiming the operation when a user doesn't have ANY application for claim. BUT this will not prevent the user from trying to claim operation with activation without a flag, when the user has another one. 
The subsequent approval will fail, so this is more of a convenience issue than a security issue, but even so, this fix doesn't correct the reported behavior.

romanstrobl:
Ok, this will have a bigger impact when we start to use `activationId` during claim.

romanstrobl:
I am closing the pull request, because the change requires a new claim REST API method with `activationId` parameter to be implemented. We will implement it later properly.

PR #2381, opened 2026-05-12
Fix #2147: Implement auditing of asymmetric signatures

### LINKED ISSUE #2147: Implement auditing of asymmetric signatures
### Task specification

Currently only authentication codes are audited in the `pa_signature_audit` table. We should extend the auditing to asymmetric signatures, too. Hence, the name of the table will remain as is for crypto4.

romanstrobl:
This PR adds audit logging for asymmetric signature verification (ECDSA and ML-DSA) to the `pa_signature_audit` table, bringing it in line with the existing symmetric signature auditing.


docs/db/changelog/changesets/powerauth-java-server/db.changelog-module.xml
```diff
@@ -21,5 +21,6 @@
     <include file="1.10.x/db.changelog-version.xml" relativeToChangelogFile="true" />
     <include file="2.0.x/db.changelog-version.xml" relativeToChangelogFile="true" />
     <include file="2.1.x/db.changelog-version.xml" relativeToChangelogFile="true" />
+    <include file="2.2.x/db.changelog-version.xml" relativeToChangelogFile="true" />
```
jnpsk, line 24: The `docs/sql/*/create_schema.sql` should also be updated

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AuditingServiceBehavior.java
```diff
@@ -217,6 +231,67 @@ public void logAuthenticationAuditRecord(ActivationRecordDto activation, Authent
...
+        signatureAuditRecord.setActivation(activationRepository.getReferenceById(activation.getActivationId()));
+        signatureAuditRecord.setActivationStatus(activation.getActivationStatus());
+        signatureAuditRecord.setActivationCounter(activation.getCounter());
+        signatureAuditRecord.setActivationCtrDataBase64(activation.getCtrDataBase64());
+        signatureAuditRecord.setDataBase64(dataBase64);
+        signatureAuditRecord.setSignature(signatureBase64);
+        signatureAuditRecord.setSignatureType(ASYMMETRIC_SIGNATURE_TYPE);
+        signatureAuditRecord.setSignatureAlgorithm(signatureAlgorithm);
+        signatureAuditRecord.setSignatureFormat(signatureFormat);
+        signatureAuditRecord.setSignatureVersion(activation.getVersion().toString());
```
jnpsk, line 263: Doesn't the `activation.getVersion()` hold the major version only? Currently the `pa_signature_audit` holds full version.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/AuditingServiceBehavior.java
```diff
@@ -160,8 +169,9 @@ public SignatureAuditResponse getSignatureAuditLog(SignatureAuditRequest request
      * @param version          Signature version.
      * @param note             Record additional info (for example, reason for signature validation failure).
      * @param currentTimestamp Record timestamp.
+     * @throws GenericServiceException If the signature version is not supported.
      */
-    public void logSignatureAuditRecord(ActivationRecordDto activation, SignatureData signatureData, SignatureType signatureType, boolean valid, Integer version, String note, Date currentTimestamp) {
+    public void logSignatureAuditRecord(ActivationRecordDto activation, SignatureData signatureData, SignatureType signatureType, boolean valid, Integer version, String note, Date currentTimestamp) throws GenericServiceException {
```
jnpsk, line 174: 🤔 I did not look at that deeper, however if this exception is thrown, doesn't that mean that logging may cause 400 of the signature verification regardless of the signature validity?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AuditingServiceBehavior.java
```diff
@@ -217,6 +231,67 @@ public void logAuthenticationAuditRecord(ActivationRecordDto activation, Authent
...
+     * @param valid             Flag indicating whether the signature was valid.
+     * @param note              Additional information (for example, reason for verification failure).
+     * @param currentTimestamp  Verification timestamp.
+     */
+    public void logAsymmetricSignatureAuditRecord(ActivationRecordEntity activation, String dataBase64, String signatureBase64,
+                                                  String signatureAlgorithm, String signatureFormat,
+                                                  boolean valid, String note, Date currentTimestamp) {
+        // Audit the asymmetric signature into the database
+        final SignatureEntity signatureAuditRecord = new SignatureEntity();
+        signatureAuditRecord.setActivation(activationRepository.getReferenceById(activation.getActivationId()));
```
jnpsk: The passed `ActivationRecordEntity` is not in same transaction? The database roundtrip seem unnecessary

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/SignatureEntity.java
```diff
@@ -122,6 +122,18 @@ public class SignatureEntity implements Serializable {
     @Column(name = "signature_data_body", updatable = false, columnDefinition = "CLOB")
     private String signatureDataBody;
 
+    /**
+     * Signature algorithm (e.g. PowerAuth-V3, PowerAuth-V4, ECDSA, ML-DSA-67).
```
jnpsk, line 126: What is PowerAuth signature algorithm in this context? Are there situations in which we cannot log specific algorithm?

jnpsk (commented):
Adding some comments after first read

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v4/AuditingServiceBehavior.java
romanstrobl: We do not have detailed information for minor version when calling the asymmetric signature endpoints.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/model/entity/SignatureEntity.java
romanstrobl: I refer to https://developers.wultra.com/components/powerauth-crypto/2.1.x/documentation/Computing-and-Validating-Authentication-Codes 

I do not think there is any other name other than `PowerAuth` for this algorithm.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/v3/AuditingServiceBehavior.java
romanstrobl: The exception can only occur in case the cryptography protocol is not supported. I do not think we need to audit such requests. The log contains an `error` for this case.

jnpsk (dismissed):
Looks good!

jnpsk (dismissed):

docs/db/changelog/changesets/powerauth-java-server/2.2.x/20260428-asymmetric-signature-audit.xml
```diff
@@ -47,8 +47,8 @@
     </changeSet>
 
     <changeSet id="3" logicalFilePath="powerauth-java-server/2.2.x/20260428-asymmetric-signature-audit.xml" author="Roman Strobl">
-        <comment>Change data type of signature column in pa_signature_audit table from varchar(255) to text to support longer asymmetric signatures</comment>
-        <modifyDataType tableName="pa_signature_audit" columnName="signature" newDataType="text"/>
+        <comment>Change data type of signature column in pa_signature_audit table from varchar(255) to varchar(8000) to support longer asymmetric signatures</comment>
+        <modifyDataType tableName="pa_signature_audit" columnName="signature" newDataType="varchar(8000)"/>
```
jnpsk, line 51: Do not we usually use text type for long strings? Is there significant benefit for using this approach?
In this case, the new create script should be generated, ideally with refreshed `arch_db_structure.png`.

romanstrobl: The problem is that `text` will require problematic migration on Oracle and MSSQL (on Postgres there is no performance problem) because existing data with tens of millions records need to be migrated.

jnpsk (approved):

PR #2382, opened 2026-05-13
Fix #2379: Support Spring Boot 4.0.x

### LINKED ISSUE #2379: Support Spring Boot 4.0.x
### Task specification

Add support for Spring Boot 4.0.x which is incompatible with previous major release.

---

Spring Boot 4 migration epic: wultra/tasklist#949

Official migration guide: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-4.0-Migration-Guide

romanstrobl:


powerauth-client-model/pom.xml
```diff
@@ -44,13 +44,9 @@
             <groupId>org.hibernate.validator</groupId>
             <artifactId>hibernate-validator</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-annotations</artifactId>
```
banterCZ, line 49: Shouldn't we have as few dependencies as possible for the client model and keep `jackson-annotations` only?

powerauth-fido2/src/main/java/com/wultra/powerauth/fido2/rest/model/converter/serialization/AttestationStatementDeserializer.java
```diff
@@ -65,56 +62,51 @@ public AttestationStatementDeserializer(Class<AttestationStatement> vc) {
...
+        } else {
+            result.setAlgorithm(SignatureAlgorithm.UNKNOWN);
+        }
+        final byte[] signature = (byte[]) map.get("sig");
+        result.setSignature(signature);
+        if (signature == null) {
+            result.setAttestationType(AttestationType.NONE);
+            return result;
+        }
+        Object x5cObj = map.get("x5c");
```
banterCZ: Not directly related to Spring Boot 4, but may we make it _final_ when we are already touching the line?

```suggestion
        final Object x5cObj = map.get("x5c");
```

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -277,18 +276,15 @@ public class PowerAuthServiceConfiguration {
...
     @Bean
-    public ObjectMapper objectMapper() {
-        final ObjectMapper objectMapper = new ObjectMapper();
-        objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
-        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-        objectMapper.enable(SerializationFeature.INDENT_OUTPUT);
-        objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
-        objectMapper.registerModule(new JavaTimeModule());
-        return objectMapper;
+    public JsonMapperBuilderCustomizer jsonMapperBuilderCustomizer() {
```
banterCZ, line 282: 👀 This is an interesting concept, returning a _customizer_ instead of the _mapper_.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
```diff
@@ -17,26 +17,26 @@
...
  *
  * @author Vit Kotacka, vit.kotacka@wultra.com
  */
-@DataJpaTest
+@SpringBootTest
 @ActiveProfiles("test")
-@Import(ObjectMapper.class)
 @Sql
+@Transactional
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_CLASS)
```
banterCZ: I would rather avoid `DirtiesContext`, if possible. It slows down the tests.

powerauth-java-server/src/test/resources/com/wultra/security/powerauth/app/server/service/PowerAuthServiceTest.sql
```diff
@@ -6,8 +6,8 @@ INSERT INTO pa_master_keypair (id, application_id, master_key_private_base64, ma
...
-    ('ce0b5480-468a-42c2-a920-d36a0d10c594', 221, 'user2', 'POWERAUTH', null, 'N67RO-EUY2L-UQU4N-6EYQQ', 1, null, 0, null, 0, null, null, null, null, null, '[ ]', 0, 5, 'AI4zPmmfmWBqMIU3F1E9xJddDaXIisrTZaR53B7W99A2', 0, 'BERFmVrT9CYwPnMbpNsXKwebbyzOrDfvSslj2lFCGvnClRsnkmmruEyVlSsHXqYfe26mSQ4CQM2xafsXSSxRBHI=', 0, '2022-06-07 09:48:34.306000', '2022-06-07 09:43:34.306000', '2022-06-07 09:43:34.306000', '2022-06-07 09:43:34.309000', 221, false, false, false, 3);
+    ('0d34cfc4-af98-4eb8-aba5-58f766bd2967', 221, 'user1', 'powerauth', 'oldest', 'DO3MN-7NUEG-5ETT4-CJZEA', 1, null, 0, null, 0, null, null, null, null, null, '[ ]', 0, 5, 'APD6fH4t8WRiLr60mQQBMusCfgHjl5/oxe3fAE8NM6IB', 0, 'BDE7yrNfAL/RRSTvoKOp8C24EPr5/OZNljr8UkwcjQIgZKypfhFDMd+JCDJOgw3/a4J5z52y+QcOhqWEIGDiHuI=', 0, '2022-06-07 09:48:34.164000', '2022-04-04 09:43:34.164000', '2022-06-07 09:43:34.164000', '2022-06-07 09:43:34.167000', 221, false, false, false, 3),
```
banterCZ, line 11: 👍 yes, the convertor using lower case

banterCZ (commented):
Some tests are failing (🐤 probably because of not merged wultra-core, right?); otherwise, great job. Could we please start writing a changelog? I would rather not see `DirtiesContext` in tests, if possible, to run them as fast as possible.

vita-kotacka (dismissed):
LGTM, migration looks thorough 👍

One thing caught my eye — same issue as in java-core #452: `WRITE_DATES_AS_TIMESTAMPS` is missing from the new builder in `PowerAuthServiceConfiguration.jsonMapperBuilderCustomizer()`.

The old bean had:
```java
objectMapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```

But the new chain only disables `FAIL_ON_EMPTY_BEANS` and `FAIL_ON_UNKNOWN_PROPERTIES`. Since Jackson 3 still defaults to timestamps, date/time fields will come out as numbers instead of ISO-8601 strings — which would break API clients.

Quick fix: `.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)` in the builder chain.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
romanstrobl: Without it the tests will fail. You can try to find another solution?

romanstrobl:
> Some tests are failing (🐤 probably because of not merged wultra-core, right?); otherwise, great job. Could we please start writing a changelog? I would rather not see `DirtiesContext` in tests, if possible, to run them as fast as possible.

The tests will not be failing after we merge wultra-core.

I will add the changelog.

The tests will not pass without `DirtiesContext` anymore, because they conflict on data.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
banterCZ: Ok, let's upgrade Spring Boot. I may fine-tune that later on.

romanstrobl:
> But the new chain only disables `FAIL_ON_EMPTY_BEANS` and `FAIL_ON_UNKNOWN_PROPERTIES`. Since Jackson 3 still defaults to timestamps, date/time fields will come out as numbers instead of ISO-8601 strings — which would break API clients.

See: https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md#changes-serializationfeature

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
romanstrobl: Yes, and I had to do it because there is another `@Primary` `@Bean` published for `ObjectMapper`, so these two would clash.

powerauth-client-model/pom.xml
romanstrobl: Yes.

romanstrobl: The problem is it will not build without `databind`:

error: package tools.jackson.databind.annotation does not exist

This is caused by `@Jacksonized` annotations. For now I will keep it as is, unless you have another idea.

banterCZ (dismissed):
Thanks for the changelog.

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepositoryTest.java
banterCZ: I guess that the root cause is that `@DataJpaTest` does perform rollback out of the box.

vita-kotacka (approved):
LGTM

PR #2392, opened 2026-05-28
Fix #2390: Temporary block of activation

### LINKED ISSUE #2390: Temporary block of activation
### Description

As of now, the activation becomes blocked after reaching the configured max attempts.

We want to remove the block after a configured time period.
The feature should apply for cryptography protocol v4 (and higher).

- The block period is configurable, default 5 minutes.
- To prevent brute forcing, in case of consecutive blocking, the block period is multiplied by a configurable multiplier (default 2). i.e., defaulst 1st block period is 5 minutes, second 10 minutes, third 20 minutes, and so on.
- The timestamp when the block would be lifted is at disposal in SDK (activation status response).
- After the block period ends, the activation status is set to `ACTIVE`, unless the activation was removed.
- The effect of unblocking should be immediate - at least in case a user tries to use the activation.
- The effect of unlocking should NOT reset the fail attempt counter., e.g., 6th consecutive incorrect attempt blocks the device for a prolonged period, but the activation should be usable after unblocking.
- The successful attempt currently resets the fail attempt counter, this attempt should also reset the multiplier for the block period.
- The operation status is independent from the activation status i.e., operations fail when their max attempts is reached.

### Acceptance criteria

The activation blocks after reaching the max attempts, but is eventually unblocked.

### Technical specification

The basic activation status enum should not be expanded - a temporarily blocked activation should behave as a blocked activation.

The temporary block feature is configurable.
The block period is configurable.
The multiplier for extending the block period is configurable.

### QA specification

Tested manually because of timeouts.

romanstrobl:
See https://github.com/wultra/powerauth-server/issues/2390


powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/TemporaryBlockService.java
```diff
@@ -0,0 +1,218 @@
...
+            activation.setTimestampBlockExpire(null);
+            logger.info("Activation blocked permanently, activation ID: {}", activation.getActivationId());
+            return;
+        }
+        final long nextBlockCount = activation.getTemporaryBlockCount() + 1L;
+        activation.setTemporaryBlockCount(nextBlockCount);
+        final long periodMs = computeBlockPeriodMillis(nextBlockCount);
+        final long currentMs = currentTimestamp.getTime();
+        if (periodMs > Long.MAX_VALUE - currentMs) {
+            activation.setTimestampBlockExpire(new Date(Long.MAX_VALUE));
```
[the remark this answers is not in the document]
romanstrobl: Fixed using `LocalDate.MAX`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationServiceBehavior.java
```diff
@@ -240,6 +245,9 @@ public LookupActivationsResponse lookupActivations(LookupActivationsRequest requ
             }
 
             for (ActivationRecordEntity activation : filteredActivationList) {
+                // Expire temporary block before reading status
+                temporaryBlockService.expireTemporaryBlockIfRequired(activation, new Date());
```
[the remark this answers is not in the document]
romanstrobl: Lookup should work avoid expiring so that the functionality works the same way as for deactivating pending activations.

docs/PowerAuth-Server-2.2.0.md
```diff
@@ -11,6 +11,26 @@ The database table `pa_signature_audit` has been extended with new columns used
...
+
+The database table `pa_activation` has been extended with new columns to support the temporary activation block feature:
+- column `timestamp_block_expire` with type `DATETIME` (nullable) - timestamp after which a temporary activation block is expired
+- column `temporary_block_count` with type `BIGINT` (NOT NULL, DEFAULT 0) - counter of consecutive temporary blocks, used as exponent for the block-period multiplier
+
+Existing rows are migrated with `temporary_block_count = 0` and `timestamp_block_expire = NULL`. Existing permanently blocked activations remain blocked until they are unblocked manually; only newly triggered `MAX_FAILED_ATTEMPTS` blocks are temporary.
+
+## Temporary Activation Block Feature
+
+A new feature automatically expires the activation block after a configurable period when the block was caused by reaching the maximum number of failed authentication attempts. The feature applies to activations using cryptography protocol v4. Activations using older protocol versions continue to be blocked permanently.
```
jnpsk: Nitpick - I would write automatically unblocks the activation instead, to be more explicit (instead expires the activation block)

docs/WebServices-Methods-V3.md
```diff
@@ -695,36 +695,37 @@ REST endpoint: `POST /rest/v3/activation/status`
...
+| `String`                  | `applicationId`              | An identifier of an application                                                                                        |
+| `String[]`                | `applicationRoles`           | Application roles                                                                                                      |
+| `DateTime`                | `timestampCreated`           | A timestamp when the activation was created                                                                            |
+| `DateTime`                | `timestampLastUsed`          | A timestamp when the activation was last used                                                                          |
+| `DateTime`                | `timestampLastChange`        | A timestamp of last activation status change                                                                           |
+| `DateTime`                | `timestampBlockExpire`       | Timestamp after which a temporary activation block is automatically lifted (null when no temporary block is in effect) |
```
jnpsk: Isn't the response field `timestampBlockExpire` always null for v3?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepository.java
```diff
@@ -139,6 +139,19 @@ public interface ActivationRepository extends JpaRepository<ActivationRecordEnti
...
+     * Fetch all activations that are temporarily blocked because of reaching the maximum number of failed
+     * authentication attempts and whose temporary block period has already expired according to the provided
+     * current timestamp and they can be unblocked.
+     *
+     * @param blockedStatus    Activation status indicating a blocked activation ({@link ActivationStatus#BLOCKED}).
+     * @param blockedReason    Blocked reason that is applicable for a temporary block ({@code MAX_FAILED_ATTEMPTS}).
+     * @param currentTimestamp Current timestamp.
+     * @return Stream of activations whose temporary block has expired.
+     */
+    @Query("SELECT a FROM ActivationRecordEntity a WHERE a.activationStatus = :blockedStatus AND a.blockedReason = :blockedReason AND a.timestampBlockExpire IS NOT NULL AND a.timestampBlockExpire <= :currentTimestamp")
```
jnpsk, line 152: I guess this query would benefit from an index on the `activation_status`, maybe also on the `timestamp_block_expire`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,220 @@
...
+@RequiredArgsConstructor
+public class TemporaryBlockService {
+
+    private static final long MAX_DATE_MILLISECONDS = 253_370_764_800_000L;
+
+    /**
+     * Minimal cryptography protocol version for which the temporary activation block feature applies.
+     * Activations using an older protocol version are blocked permanently (legacy behavior).
+     */
+    private static final long MIN_PROTOCOL_VERSION_FOR_TEMPORARY_BLOCK = 4L;
```
jnpsk, line 53: But it seems the functionality is actually implemented for V3, just blocked by this condition?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param blockCount Current block count.
+     * @return Temporary block period in milliseconds.
+     * @throws GenericServiceException Thrown in case of server configuration error.
+     */
+    private long computeBlockPeriodMillis(final long blockCount) throws GenericServiceException {
+        final long blockPeriodBase = powerAuthServiceConfiguration.getTemporaryBlockPeriodInMilliseconds();
+        final int multiplier = powerAuthServiceConfiguration.getTemporaryBlockMultiplier();
+        if (blockPeriodBase <= 0 || multiplier <= 0 || blockCount <= 0) {
+            logger.error("Incorrect server configuration, temporary block period: {}, multiplier: {}, block count: {}", blockPeriodBase, multiplier, blockCount);
+            throw localizationProvider.buildExceptionForCode(ServiceError.CONFIGURATION_ERROR);
```
jnpsk: The new `CONFIGURATION_ERROR` code is used here only, but it seems it cannot be reached. The configuration should already be validated on startup and `blockCount` cannot be negative without a database intervention. Anyway, it would be translated to BAD REQUEST, which does not seem to be the case.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
```diff
@@ -132,6 +132,33 @@ public class PowerAuthServiceConfiguration {
...
+    private long temporaryBlockPeriodInMilliseconds;
+
+    /**
+     * Multiplier used to extend the temporary block period for consecutive blocks. With the default value
+     * of 2 the n-th consecutive temporary block lasts {@code periodInMilliseconds * 2^(n-1)} milliseconds.
+     * Must be at least 1.
+     */
+    @Value("${powerauth.service.crypto.temporaryActivationBlock.multiplier:2}")
+    @Min(1)
+    private int temporaryBlockMultiplier;
```
jnpsk, line 160: I am thinking if double wouldn't be more versatile, but for an activation blocking I guess int is sufficient.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,220 @@
...
+        }
+        // Compute blockPeriodBase * multiplier^(blockCount-1)
+        long period = blockPeriodBase;
+        for (long i = 1; i < blockCount; i++) {
+            if (period > Long.MAX_VALUE / multiplier) {
+                // Overflow protection, use maximum long value
+                period = Long.MAX_VALUE;
```
jnpsk: Loop may end when reaching this block. It feels a bit like overkill for small counts, but it is a good way to keep it generic.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,220 @@
...
+            activation.setTimestampBlockExpire(null);
+            logger.info("Activation blocked permanently, activation ID: {}", activation.getActivationId());
+            return;
+        }
+        final long nextBlockCount = activation.getTemporaryBlockCount() + 1L;
+        activation.setTemporaryBlockCount(nextBlockCount);
+        final long periodMs = computeBlockPeriodMillis(nextBlockCount);
+        final long currentMs = currentTimestamp.getTime();
+        if (periodMs > MAX_DATE_MILLISECONDS - currentMs) {
+            activation.setTimestampBlockExpire(new Date(MAX_DATE_MILLISECONDS));
```
jnpsk, line 94: Seems the `computeBlockPeriodMillis` method can have ceiling to  `MAX_DATE_MILLISECONDS` instead of the `Long.MAX_VALUE`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,220 @@
...
+                .findActivationForUpdate(activationToCheck.getActivationId())
+                .orElse(null);
+        if (activation == null) {
+            logger.warn("Activation was removed while expiring temporary block, activation ID: {}", activationToCheck.getActivationId());
+            return;
+        }
+        // Re-check the locked activation to avoid race conditions
+        if (!isTemporaryBlockExpired(activation, currentTimestamp)) {
+            return;
+        }
```
jnpsk: This recheck is in the same transaction, does it actually recheck?

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/task/CleaningTask.java
```diff
@@ -81,6 +81,16 @@ public void expireActivations() {
         activationServiceBehavior.expireActivations();
     }
 
+    @Scheduled(fixedRateString = "${powerauth.service.scheduled.job.expireTemporaryActivationBlocks:5000}")
+    @SchedulerLock(
+            name = "expireTemporaryActivationBlocksTask",
+            lockAtLeastFor = "#{T(java.lang.Math).round(${powerauth.service.scheduled.job.expireTemporaryActivationBlocks:5000} * 0.8)}")
+    public void expireTemporaryActivationBlocks() {
```
jnpsk, line 88: By default the feature is disabled, so this job is effectively no-op. However, it does not seem there is a simple way to make the scheduled job conditional and the overhead is small, so we can leave it as is.

jnpsk (commented):
Good job! 👏 
Sharing some minor notes.

docs/WebServices-Methods-V3.md
romanstrobl: I will update the documentation.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
romanstrobl: Yes, the logic is shared, just blocked for V3. I will update the service name to reflect this.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/configuration/PowerAuthServiceConfiguration.java
romanstrobl: I think int is sufficient.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
romanstrobl: Good point, we can change this error to something more generic.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
romanstrobl: Good point, I will exit the loop in case of overflow.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
romanstrobl: Yes, we can limit it like this.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/database/repository/ActivationRepository.java
romanstrobl: The index on `timestamp_block_expire` should be the most efficient, because most values are `NULL`.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
romanstrobl: Good point. I guess we should refresh the entity context for this kind of time-sensitive queries. I will research the best option to achieve this.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,223 @@
...
+        for (long i = 1; i < blockCount; i++) {
+            if (period > MAX_DATE_MILLISECONDS) {
+                // Do not allow longer block period than the value MAX_DATE_MILLISECONDS
+                period = MAX_DATE_MILLISECONDS;
+                break;
+            } else {
+                period *= multiplier;
+            }
+        }
+        return period;
```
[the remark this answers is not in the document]
romanstrobl: I will handle the division by multiplier. We don't want to handle the misconfiguration cases.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,223 @@
...
+        // Re-check the locked and refreshed activation to avoid race conditions
+        if (!isTemporaryBlockExpired(activation, currentTimestamp)) {
+            return;
+        }
+        activation.setActivationStatus(ActivationStatus.ACTIVE);
+        activation.setBlockedReason(null);
+        activation.setTimestampBlockExpire(null);
+        // Leave one last authentication attempt available, so the user can recover by successful verification.
+        // The temporary block counter is preserved so that the next failure is blocked for a longer period.
+        activation.setFailedAttempts(Math.max(activation.getFailedAttempts() - 1L, 0L));
```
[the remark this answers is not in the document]
romanstrobl: This change is incorrect, at least one attempt must be available to allow successful authentication.

powerauth-java-server/src/main/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/ActivationBlockService.java
```diff
@@ -0,0 +1,223 @@
...
+     */
+    public void blockActivation(final ActivationRecordEntity activation, final Date currentTimestamp) throws GenericServiceException {
+        // If the activation is already temporarily blocked and the block period has not elapsed yet, keep
+        // the existing block in place.
+        if (isTemporaryBlockActive(activation, currentTimestamp)) {
+            logger.debug("Activation is already temporarily blocked, keeping the existing block, activation ID: {}, expires at: {}",
+                    activation.getActivationId(), activation.getTimestampBlockExpire());
+            return;
+        }
+        activation.setActivationStatus(ActivationStatus.BLOCKED);
```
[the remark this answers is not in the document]
romanstrobl: I will add tests.

docs/WebServices-Methods-V3.md
jnpsk: The response objects are shared for v3 and v4? Or what is the reason of adding that field?

romanstrobl: Yes, `Activation` model class is shared.

jnpsk (approved):
Looks good

PR #2394, opened 2026-05-29
Fix #2393: Define signature column length for tests

### LINKED ISSUE #2393: Define signature column length for tests
### Task specification

The H2 DB used in tests needs a length definition for the `signature` column in `signature_audit` for asymmetric signatures.

romanstrobl:
The `signature` column length should be set for tests which use H2 database.


banterCZ (approved):
👏 Thanks for the fix!

PR #2437, opened 2026-08-04
Fix #2436: Callbacks send date as string instead of long (timestamp)

### LINKED ISSUE #2436: Callbacks send date as string instead of long (timestamp)

2.2

### Environment

_No response_

field: timestampExpires

### Log output

_No response_

banterCZ:


michal-rozehnal-w (dismissed):
Looks good

zcgandcomp (changes requested):
why are we changing API?

banterCZ:
We are not changing the API; we are fixing the API that was changed by accident because of the Jackson upgrade.

romanstrobl (changes requested):
I will only approve this change after a proper analysis of the problem.

romanstrobl:
Incorrect fix, we will fix this via https://github.com/wultra/powerauth-server/pull/2438

romanstrobl:
Main part of fix: https://github.com/wultra/java-core/pull/476/changes

PR #2438, opened 2026-08-05
Fix #2436: Callbacks send date as string instead of long (timestamp)

### LINKED ISSUE #2436: Callbacks send date as string instead of long (timestamp)

romanstrobl:
Reverted serialization change for Dates in callbacks.


powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehaviorTest.java
```diff
@@ -249,6 +251,41 @@ void testRemoveCallbackUrl() throws Exception {
...
+
+        @SuppressWarnings("unchecked")
+        final ArgumentCaptor<Map<String, Object>> captor = ArgumentCaptor.forClass(Map.class);
+        verify(callbackUrlEventService).createAndSaveEventForProcessing(eq(enabledCallback), captor.capture());
+
+        final Map<String, Object> callbackData = captor.getValue();
+
+        // The timestamps must be converted to numeric epoch milliseconds so that both callback dispatch
+        // sub-paths (instant dispatch and reload from the database) serialize them as JSON numbers.
+        assertInstanceOf(Long.class, callbackData.get("timestampCreated"));
```
banterCZ, line 281: Just thinking out loud. Existing callback data is stored as text, is it? But since there is a fix for the REST client, it should be fine https://github.com/wultra/java-core/pull/476
I was about to suggest backporting the test to 2.1.x, but it would fail, right?
I expect that this fix is not needed, but yes, it may reduce confusion.

banterCZ (dismissed):

powerauth-java-server/src/test/java/com/wultra/security/powerauth/app/server/service/behavior/tasks/CallbackUrlBehaviorTest.java
romanstrobl: Yes, we do this to avoid confusion.

banterCZ (approved):

PR #2445, opened 2026-08-12
Fix #2444: Authenticator data with extension deserialization may fail

### LINKED ISSUE #2444: Authenticator data with extension deserialization may fail

2.2.x

### Environment

_No response_

`AuthenticatorDataDeserializer` fails to parse authenticator data whose key blob is followed by trailing bytes (e.g. WebAuthn extension data).
This is caused by the change of default behaviour of Jackson 3, where `FAIL_ON_TRAILING_TOKENS` is now enabled by default.

### Log output

```
com.wultra.powerauth.fido2.rest.model.converter.serialization.Fido2DeserializationException: Unable to deserialize credentialPublicKey.
...
Caused by: tools.jackson.databind.exc.MismatchedInputException: Trailing token (`JsonToken.START_OBJECT`) found after value (bound as `java.util.Map`): not allowed as per `DeserializationFeature.FAIL_ON_TRAILING_TOKENS`
 at [Source: (byte[])[80 bytes]; byte offset: #77]
```

jnpsk:
Fixes #2444
FIDO2 authenticator data that includes CBOR extensions failed to deserialize after the Jackson 2 → 3 migration.

**Root cause**

Jackson 3 enables  `DeserializationFeature.FAIL_ON_TRAILING_TOKENS`  by default (it was off in Jackson 2). `AuthenticatorDataDeserializer`  passes all bytes after the credential ID — the public-key CBOR plus any appended extension CBOR — to the CBOR mapper. The public-key map parses fine (CBOR is self-delimiting), but the trailing extension bytes now trigger a failure.

**Proposed Fix**

Disable  `FAIL_ON_TRAILING_TOKENS`  on the CBOR mapper used for public-key parsing, restoring the previous lenient behavior and ignoring trailing extension bytes.


petrdvorak (approved):
Looks OK, thank you for the fix.

romanstrobl (approved):
Should we add this change the changelog? For the rest I approve.

romanstrobl (approved):
Good catch!

